PR #2, opened 2015-01-11
Fix memory leak by incorrect assignment of Intrusive_Auto_Ptr.

likema:


jwillemsen:
Please extend or add a new test under ACE/tests that reproduces this issue automatically which we can we use to validate the problem and also that makes sure the issue doesn't get introduced back in the future.

likema:
OK, done.

likema:
Before fixing it, the new test will be crashed. Now, it passed.

jwillemsen:
Thanks, as last request please add some documentation to the class and test to explain the behavior and what the test expects

likema:
Since the class already had straightforward docs, I just added docs for the unit test.

jwillemsen:
Did a local test and reviewed the changes, will merge this, thanks for making the fix and extending the test.

PR #6, opened 2015-02-10
ACE_Process std file handles on Windows

mitza-oci:
Allow redirection of child's stdout/stderr when there is no
stdin handle available (for example when running in a service).
TAO_IDL drv_preproc: check for errors from ACE_Process

### DIFF
--- a/TAO/TAO_IDL/driver/drv_preproc.cpp
+++ b/TAO/TAO_IDL/driver/drv_preproc.cpp
@@ -1354,7 +1354,14 @@ DRV_pre_proc (const char *myfile)
           throw Bailout ();
         }
 
-      cpp_options.set_handles (ACE_INVALID_HANDLE, fd);
+      if (cpp_options.set_handles (ACE_INVALID_HANDLE, fd) == -1)
+        {
+          ACE_ERROR ((LM_ERROR,
+                      "%s: cannot set stdout for child process: %p\n",
+                      ACE_TEXT_CHAR_TO_TCHAR (idl_global->prog_name ())));
+
+          throw Bailout ();
+        }
     }
 
   if (idl_global->compile_flags () & IDL_CF_INFORMATIVE)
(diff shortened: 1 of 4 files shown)


TAO/TAO_IDL/driver/drv_preproc.cpp
```diff
@@ -1354,7 +1354,14 @@ DRV_pre_proc (const char *myfile)
           throw Bailout ();
         }
 
-      cpp_options.set_handles (ACE_INVALID_HANDLE, fd);
+      if (cpp_options.set_handles (ACE_INVALID_HANDLE, fd) == -1)
+        {
+          ACE_ERROR ((LM_ERROR,
+                      "%s: cannot set stdout for child process: %p\n",
+                      ACE_TEXT_CHAR_TO_TCHAR (idl_global->prog_name ())));
```
jwillemsen: Given that prog_name() returns a char* why not log using %C, that way ACE_TEXT_CHAR_TO_TCHAR is not needed at all

mitza-oci: It was copied from the other logging in that file when the patch started out in TAO 1.6a.  I'll update it.

jwillemsen: Thanks, could you at the same moment check the other occurences of prog_name() logging? Then all are consistent.

mitza-oci: The other uses in drv_preproc.cpp are OK.  The remaining inconsistent ones are in drv_args.cpp and tao_id.cpp.

jwillemsen:
Did a quick google search and according to Microsoft documentation handle 0 is equal to keyboard input (http://www.microsoft.com/resources/documentation/windows/xp/all/proddocs/en-us/redirection.mspx?mfr=true)

mitza-oci:
That's true for cmd shell but it doesn't seem to apply at the OS API.  https://msdn.microsoft.com/en-us/library/windows/desktop/ms683231%28v=vs.85%29.aspx "If an application does not have associated standard handles, such as a service running on an interactive desktop, and has not redirected them, the return value is NULL."

jwillemsen:
Ok, thanks for the link, at the moment the travis ci build also has no problems feel free to merge the pull request

PR #14, opened 2015-02-26
Net app vx works64

shuston:
This is a set of changes I am assisting a customer with. I do not have a build environment to test with, but the customer does and they've approved the changes for VxWorks.

### DIFF
--- a/ACE/ace/OS_NS_Thread.cpp
+++ b/ACE/ace/OS_NS_Thread.cpp
@@ -5132,7 +5132,9 @@ ACE_END_VERSIONED_NAMESPACE_DECL
 int
 spa (FUNCPTR entry, ...)
 {
-  static _Vx_usr_arg_t const ACE_MAX_ARGS = 10;
+  // The called entrypoint can get the function name plus the normal 10
+  // optional arguments.
+  static _Vx_usr_arg_t const ACE_MAX_ARGS = 1 + 10;
   static char *argv[ACE_MAX_ARGS] = { 0 };
   va_list pvar;
   _Vx_usr_arg_t argc;
--- a/ACE/ace/OS_NS_Thread.cpp
+++ b/ACE/ace/OS_NS_Thread.cpp
@@ -4438,7 +4438,7 @@ ACE_OS::thr_create (ACE_THR_FUNC func,
                          (int) flags,
                          stacksize,
                          thread_args->entry_point (),
-                         (_Vx_usr_arg_t) thread_args,
+                         (ACE_VX_USR_ARG_T) thread_args,
                          0, 0, 0, 0, 0, 0, 0, 0, 0);
 #   if 0 /* Don't support setting of stack, because it doesn't seem to work. */
     }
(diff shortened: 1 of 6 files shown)


jwillemsen:
Steve, I don't have a VxWorks environment but when I quickly look at the changes I think your pull request will break support for any VxWorks version < 6.9, for example _Vx_usr_arg_t doesn't seem to be supported in older releases, see ace/Stack_Trace.cpp where there are some ACE macros to workaround the unavailability of that type in older releases. Probably this is the case for any new type/define you use, but hard to say without someone setting up a full build with VxWorks 6.9 and one with an older release, maybe your customer can do that?

shuston:
Good points, Johnny. I will work on the needed changes. Thanks for the quick review!

jwillemsen:
The VxWorks 5.5 taskLib documentation describes the API of the older systems, see the following link for that old API description http://www.vxdev.com/docs/vx55man/vxworks/ref/taskLib.html

ACE/ace/OS_NS_Thread.cpp
```diff
@@ -4472,7 +4472,7 @@ ACE_OS::thr_create (ACE_THR_FUNC func,
     }
 #   endif /* 0 */
 
-  if (tid == ERROR)
+  if (tid == TASK_ID_ERROR)
```
jwillemsen: According to VxWorks 5.5 the return value in older VxWorks versions is really ERROR, not TASK_ID_ERROR, see http://www.vxdev.com/docs/vx55man/vxworks/ref/taskLib.html#taskSpawn

ACE/ace/OS_NS_Thread.cpp
```diff
@@ -5292,19 +5298,19 @@ spae (FUNCPTR entry, ...)
...
   // larger stack size (instead of ::sp ()'s 20000).
-  int const ret = ::taskSpawn (argv[0],    // task name
-                               100,        // task priority
-                               VX_FP_TASK, // task options
-                               ACE_NEEDS_HUGE_THREAD_STACKSIZE, // stack size
-                               entry,      // entry point
-                               argc,       // first argument to main ()
-                               (int) argv, // second argument to main ()
-                               0, 0, 0, 0, 0, 0, 0, 0);
+  TASK_ID const ret = ::taskSpawn (argv[0],    // task name
```
jwillemsen: Probably TASK_ID is also a type that is VxWorks 6.9 specific

ACE/ace/OS_NS_signal.inl
```diff
@@ -15,6 +15,14 @@ kill (pid_t pid, int signum)
   ACE_UNUSED_ARG (pid);
   ACE_UNUSED_ARG (signum);
   ACE_NOTSUP_RETURN (-1);
+#elif defined (ACE_VXWORKS)
+  /*
```
jwillemsen: What when RTP is used instead of kernel mode, at that moment VxWorks does have support for a process

shuston: That's fine - the only difference is the signature for kill() on VxWorks, and it should work regardless. This change can stay unless/until someone with a VxWorks build environment can test it.

ACE/ace/OS_NS_Thread.cpp
shuston: Probably right... will change this.

ACE/ace/OS_NS_Thread.cpp
shuston: Ok, will change this.

PR #21, opened 2015-03-02
Add .gitignore files based on svn:ignore properties

mitza-oci:
Added .gitignores based on both svn:ignore and listing files are created by the build process.
The ones created by the build process are not an exhaustive set since I didn't build all tests/examples/etc., but I did build both TAO_ACE.mwc and CIAO_TAO_DAnCE_OpenDDS.mwc.


jwillemsen:
Adam, can you make sure that the make_release.py release script doesn't package the .gitignore files in the release packages?

mitza-oci:
Sure, if you don't mind me editing it without being able to test it.  Looks like we add to the 3rd argument to create_file_lists (which is "exclude").

jwillemsen:
On a local clone of your own fork you can easily test, maybe just comment out the push as additional safety during testing

mitza-oci:
Looks like I can run with --kit and it won't attempt anything that would change the repo.

mitza-oci:
Testing make_release.py worked fine.

PR #23, opened 2015-03-03
Fix to resume and register event handlers after returning -1 from handle_*

mitza-oci:
Dev_Poll_Reactor on Linux uses epoll with one-shot and
suspends handlers when an event is received.  New logic attempts
to resume handler when -1 is returned.  Added test.
From OCITAO 2.2a Fri Nov 14 16:47:56 UTC 2014

### DIFF
--- a/ACE/tests/Dev_Poll_Reactor_Echo_Test.cpp
+++ b/ACE/tests/Dev_Poll_Reactor_Echo_Test.cpp
@@ -1,0 +1,641 @@
+//=============================================================================
+/**
+ *  @file    Dev_Poll_Reactor_Echo_Test.cpp
+ *
+ *  $Id$
+ *
+ *  This test implements a simple echo server using the
+ *  Dev_Poll_Reactor.  This forces the reactor to behave like a
+ *  reactor would in a typical client/server application, i.e.,
+ *  receive a message then send a messages.
+ *  @author Justin Wilson <wilsonj@ociweb.com>
+ */
+//=============================================================================
+
+#include "test_config.h"
+
+#if defined (ACE_HAS_DEV_POLL) || defined (ACE_HAS_EVENT_POLL)
+
+#include "ace/OS_NS_signal.h"
+#include "ace/Reactor.h"
+#include "ace/Dev_Poll_Reactor.h"
+
+#include "ace/Acceptor.h"
+#include "ace/Connector.h"
+
+#include "ace/SOCK_Acceptor.h"
+#include "ace/SOCK_Connector.h"
+#include "ace/SOCK_Stream.h"
+
+#include "ace/OS_NS_unistd.h"
+#include "ace/OS_NS_netdb.h"
+
+#include <queue>
+
+typedef ACE_Svc_Handler<ACE_SOCK_STREAM, ACE_NULL_SYNCH> SVC_HANDLER;
+
+// ----------------------------------------------------
+
+class Client : public SVC_HANDLER
+{
+public:
+
+  Client (void);
+
+  //FUZZ: disable check_for_lack_ACE_OS
+  virtual int open (void * = 0);
+  //FUZZ: enable check_for_lack_ACE_OS
+
+  virtual int handle_output (ACE_HANDLE handle);
+
+  virtual int handle_input (ACE_HANDLE handle);
+
+  virtual int handle_timeout (const ACE_Time_Value &current_time,
+                              const void *act);
+
+  virtual int handle_close (ACE_HANDLE handle,
+                            ACE_Reactor_Mask mask);
+
+  std::string sent;
+  std::string received;
+
+private:
+  unsigned int call_count_;
+};
+
+
+class Server : public SVC_HANDLER
+{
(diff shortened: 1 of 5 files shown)


ACE/tests/Dev_Poll_Reactor_Echo_Test.cpp
```diff
@@ -0,0 +1,641 @@
+//=============================================================================
+/**
+ *  @file    Dev_Poll_Reactor_Echo_Test.cpp
+ *
+ *  $Id$
```
jwillemsen: $Id$ is not used anymore, can be removed

ACE/tests/Dev_Poll_Reactor_Echo_Test.cpp
```diff
@@ -0,0 +1,641 @@
...
+  ACE_UNUSED_ARG (sigmin);
+  ACE_UNUSED_ARG (sigmax);
+#endif /* ACE_LACKS_UNIX_SIGNALS */
+
+  return 0;
+}
+
+// ----------------------------------------------------
+
+// ACE_THR_FUNC_RETURN
```
jwillemsen: Any reason to have all this code commented out? Could it be removed?

ACE/tests/Dev_Poll_Reactor_Echo_Test.cpp
```diff
@@ -0,0 +1,641 @@
...
+                      -1);
+
+  if (reactor.run_reactor_event_loop () != 0)
+    ACE_ERROR_RETURN ((LM_ERROR,
+                       ACE_TEXT ("(%t) %p\n"),
+                       ACE_TEXT ("Error when running client ")
+                       ACE_TEXT ("reactor event loop")),
+                      -1);
+
+  ACE_DEBUG((LM_DEBUG, "sent: %s\n", client_handler->sent.c_str ()));
```
jwillemsen: Send and received are std string, logging should use %C, %s is for tchar

ACE/tests/Dev_Poll_Reactor_Echo_Test.cpp
mitza-oci: I don't know, I didn't write this.  I don't mind removing it though.

jwillemsen: If it is not used let us remove it now

mitza-oci:
The run_test.lst file has "Reactor_Fairness_Test: !FIXED_BUGS_ONLY" so it's not expected to pass the test even without any library changes?  I suppose that's due to TP_Reactor.
I tried the test and it seems like it passed the Dev_Poll part.
Results (5 entries):
   handle 15: 740367
   handle 16: 736513
   handle 17: 738764
   handle 18: 739612
   handle 19: 738600

As for the "Why" question, I'll have to let Justin answer that one.

shuston:
(Bringing this comment up from the code line - sorry for my github newbie mistake).
Why do you want to do this? Does your new test case fail without the change?

shuston:
Re Reactor_Fairness_Test: !FIXED_BUGS_ONLY - yes, Select_Reactor and TP_Reactor are unfair. But Dev_Poll_Reactor should be fair - your test run shows it is, so now I'm just interested in why the change in the first place?

jrw972:
Suppose that an EventHandler is registered for both read and write.  The reactor suspends the handler and dispatches to handle_output.  If handle_output returns -1, the event handler will remain suspended even though is should be eligible to receive handle_input.  The test checks for this case by making the server register for input and output at the same time and return -1 from handle_output.

shuston:
Ok,that makes sense. Can you tell me that you've run your new test program against ACE without your change and it fails? If yes, then I wholeheartedly agree with this change but would urge you to open a bugzilla for the bug you've uncovered so that it's on record with your fix.

mitza-oci:
I did run the new test against an existing libACE yesterday, and it failed.
I will open the bugzilla bug since Justin doesn't have an account there.

mitza-oci:
http://bugzilla.dre.vanderbilt.edu/show_bug.cgi?id=4200

shuston:
Terrific - I'm in favor of merging the pull. Go ahead, or I will if you'd rather.

mitza-oci:
Got it merged.  Thanks for reviewing.

jwillemsen:
Adam, can you check coverity 1287148, a new issue reported on the new unit test that is contributed as part of this pull request

mitza-oci:
I'll fix it in another pull request.

jwillemsen:
Adam, can you check coverity 1288509, a new one after your last fix for coverity in this new test

mitza-oci:
On 3/15/2015 3:57 AM, Johnny Willemsen wrote:
>
> Adam, can you check coverity 1288509 
> <https://github.com/DOCGroup/ATCD/commit/128850918c92301442e7d037865f2fdc4fb36b1e>, 
> a new one after your last fix for coverity in this new test
>
>

OK, I missed something last time around.

PR #32, opened 2015-03-13
Add ability for ACE_INET_Addr to hold all the addresses assigned to a name

shuston:
Fixed problems detected after previous pull for this branch.

### DIFF
--- a/ACE/ace/INET_Addr.cpp
+++ b/ACE/ace/INET_Addr.cpp
@@ -182,7 +182,8 @@ ACE_INET_Addr::ACE_INET_Addr (void)
 ACE_INET_Addr &
 ACE_INET_Addr::operator= (const ACE_INET_Addr& rhs)
 {
-  this->set (rhs);
+  if (this != &rhs)
+    this->set (rhs);
   return *this;
 }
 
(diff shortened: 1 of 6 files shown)


ACE/ace/INET_Addr.cpp
```diff
@@ -152,11 +152,38 @@ ACE_INET_Addr::hash (void) const
...
 ACE_INET_Addr::ACE_INET_Addr (void)
   : ACE_Addr (determine_type (), sizeof (inet_addr_))
 {
   // ACE_TRACE ("ACE_INET_Addr::ACE_INET_Addr");
-  this->reset ();
+  this->reset_i ();
+}
+
+ACE_INET_Addr &
+ACE_INET_Addr::operator= (const ACE_INET_Addr& rhs)
```
jwillemsen: Shouldn't this check for this != rhs? Also please add a test case for self assignment

jwillemsen:
Steve, looks the ACE_INet_Addr now fails at runtime and we have a new Coverity error, can you check this

jwillemsen:
It is CID1288508 with description "Failed to override method The method will not be called when virtual dispatch is used, this may cause the class to not perform as expected. In ACE_INET_Addr::​set_addr(void const*, int): Virtual function does not override parent method because of type signature mismatch, usually due to const (CWE-710)"

PR #33, opened 2015-03-14
Fix errno checking logic in TLI

likema:
When building in AIX 5.1, GCC reported

```
ACE.cpp: In function 'ssize_t ACE::t_snd_n_i(ACE_HANDLE, const void*, size_t, int, const ACE_Time_Value*, size_t*)':
ACE.cpp:1532:39: warning: suggest parentheses around '&&' within '||'
```

After reading ACE.cpp, I find it is not only a warning but also a bug.


jwillemsen:
Looks ok, will merge on a day when there are more changes going to master

likema:
OK

PR #39, opened 2015-03-16
ACE_Addr::set_addr(void*, int) was changed to ACE_Addr::set_addr(const v...

shuston:
...oid*, int). All descendant classes have the same change. Fixed multiple-address handling in ACE_INET_Addr and check for alias in INET_Addr_Test if a hostname reverse lookup doesn't match the forward lookup.

### DIFF
--- a/ACE/ace/INET_Addr.cpp
+++ b/ACE/ace/INET_Addr.cpp
@@ -1050,9 +1050,6 @@ int ACE_INET_Addr::set_address (const char *ip_addr,
         }
 #endif /* ACE_HAS_IPV6 */
 
-      this->inet_addrs_.clear ();
-      this->inet_addrs_iter_ = this->inet_addrs_.begin ();
-
       return 0;
     }   /* end if (len == 4) */
 #if defined (ACE_HAS_IPV6)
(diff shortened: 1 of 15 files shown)


ACE/ace/INET_Addr.cpp
```diff
@@ -1050,9 +1050,6 @@ int ACE_INET_Addr::set_address (const char *ip_addr,
         }
 #endif /* ACE_HAS_IPV6 */
```
jwillemsen: Is it intended to remove these clear() invocations?

jwillemsen:
Is this a change that is really required? Can imagine that this could lead to very strange problems when the user upgrades, at least without the C++11 override keyword the compiler will not detect this when the user has their own ACE_Addr derived classes.

shuston:
The set_addr() method never should have been non-const. It's a decent time to make an API change and it's more correct. The change and its ramifications are noted in the NEWS file.

jwillemsen:
Ok

mitza-oci:
A few TAO tests started failing after this was merged, for example TAO/tests/IORTable_Locator/run_test.pl.  It looks like the port number from -ORBListenEndpoints isn't being parsed into the INET_Addr.

shuston:
I'm trying to figure out what's wrong without learning TAO internals... could you please let me know what ACE_INET_Addr usage(s) is likely to be failing? I added some address:port string tests to INET_Addr_Test but they appear ok. Is IPv6 involved? Multihomed?

pmesnier:
I've extended the INET_Addr_Test to emulate TAO's use case highlighting the issue at hand. See Test_TAO_Use, you'll see the same addr instance is set() more than once which now seems to be adding a new inet addr to a list rather than overwriting.

shuston:
Great - thanks! I was hoping that was the problem. I just figured out that Multihomed_INET_Addr_Test_IPV6 failed for the same reason and am working on a branch to merge in the fix. Details forthcoming.

shuston:
Pull request 42 has the fix for this. Thank you for the help narrowing down the broken use case, Phil!
https://github.com/DOCGroup/ATCD/pull/42

PR #41, opened 2015-03-17
Cast cond_init's type to select correct variant of the method; see Bugzi...

shuston:
...lla #4201


jwillemsen:
I haven't checked all code but what about changing the argument of the calling argument from int To short, looks on first sight we have some inconsistenties in how the type is passed.

shuston:
We do have inconsistencies. Most uses of the 'type' are int. Not sure why this one is short, but I'd hazard a guess it is to get around the very problem I ran into here. I don't have time to chase down all the paths and change all the types at this point, so will merge it in.

PR #44, opened 2015-03-22
Fix forgetting to close handle in ACE_OS::stat.

likema:
In ACE_OS::stat WinCE case, It forgets to close the handle that open by FindFirstFile.


jwillemsen:
Looks ok, will merge at some point

PR #45, opened 2015-03-22
Fix connecting timeout without returning -1 in ACE_Cached_Connect_Strategy_T

likema:
ACE_Cached_Connect_Strategy_T::cached_connect timeout but still return 0.

### DIFF
--- a/ACE/ace/Cached_Connect_Strategy_T.cpp
+++ b/ACE/ace/Cached_Connect_Strategy_T.cpp
@@ -271,7 +271,10 @@ ACE_Cached_Connect_Strategy_Ex<SVC_HANDLER, ACE_PEER_CONNECTOR_2, CACHING_STRATE
       // connect().
 
       if (errno == EWOULDBLOCK || errno == ETIMEDOUT)
-        errno = ENOTSUP;
+        {
+          errno = ENOTSUP;
+          return -1;
+        }
       else if (ACE::out_of_handles (errno) || errno == EADDRINUSE)
         {
           // If the connect failed due to the process running out of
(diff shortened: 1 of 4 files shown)


jwillemsen:
Thanks for the pull request, can you extend/add a test under ACE/tests that reproduced this problem?

likema:
OK, I will do it later.

likema:
It is hard to simulate EWOULDBLOCK or ETIMEDOUT. 

```
if (errno == EWOULDBLOCK || errno == ETIMEDOUT)
        {
          errno = ENOTSUP;
          return -1;
        }
```

In man connect,

* *EAGAIN* No  more  free  local  ports  or  insufficient  entries  in  the  routing  cache.   For  AF_INET  see  the description of  /proc/sys/net/ipv4/ip_local_port_range ip(7) for information on how to increase the number of local ports.
* *ETIMEDOUT* Timeout while attempting connection.  The server may be too busy to accept new connections.  Note that for IP sockets the timeout may be very long when syncookies are enabled on the server.

I have written a test with ACE_Synch_Options timeout, but it reports errno is ETIME rather than EAGAIN or ETIMEDOUT.

jwillemsen:
New test also needs to be added to tests/run_test.lst so that it runs as part of the test suite

mitza-oci:
Since the change was reverted, is there a way to reopen the pull request?

jwillemsen:
As far as I remember you have to create a new pull request

likema:
Why revert it?

jwillemsen:
The pull request was reverted through request #143 because the new unit test was failing on most of the platforms.

PR #57, opened 2015-04-20
Feature/fix windows process

chenjianlong:


shuston:
Could you please link this to some sort of trouble report? Also, please add a test case that shows the problem you are fixing. None of the tests cases run to date show a problem in this area.

jwillemsen:
See http://bugzilla.dre.vanderbilt.edu/ for the bugzilla system to report an issue

jwillemsen:
Closing pull request because it is not clear what problem is fixed and there is no test case

PR #67, opened 2015-04-29
ACE_CDR::Fixed corrected a bug and short-circuit division by 1.

mitza-oci:


mitza-oci:
Travis CI error "Your test run exceeded 50 minutes." (only for ACETESTS=1), will merge anyway.

PR #71, opened 2015-05-05
Added truncate_cast because suseconds_t is 32bit on Windows x86

jwillemsen:
* ACE/ace/Time_Value.h:


jwillemsen:
Merging right now because jenkins doesn't use a gcc compiler that has C++11 support

PR #76, opened 2015-05-11
Fixed coverity issues 1297401 and 1297404

mitza-oci:


ACE/ace/CDR_Base.cpp
```diff
@@ -948,8 +951,8 @@ void ACE_CDR::Fixed::normalize (UShort min_scale)
 
 ACE_CDR::Fixed ACE_CDR::Fixed::from_string (const char *str)
 {
-  const bool negative = str && *str == '-';
-  if (negative || (str && *str == '+'))
+  const bool negative = *str == '-';
```
shuston: It seems like this removes a valid check for zero-valued str pointer, which is probably a bad idea. In fact, CID 1297401 complains that a zero-valued str is passed to strspn, below. It seems better to check for zero str before doing any of these other checks and bail out right away.

mitza-oci: I think Coverity lets you treat it as a precondition as long as you're consistent.  I'd rather treat it as a precondition to make this function act like any of the std::str* functions.

mitza-oci:
Let's see how this version works with the next Coverity scan.  If there are still problems we can fix before the next release.

ACE/ace/CDR_Base.cpp
shuston: Coverity is still flagging it as an issue. If there's a way to specify a known precondition, could you do that, else fix the code to check null str first and bail before doing content checks?

mitza-oci: I'll take a look. Does anyone have a link to coverity docs?

On May 19, 2015 5:59 PM, "Steve Huston" <notifications@github.com> wrote:
>
> In ACE/ace/CDR_Base.cpp:
>
> > @@ -948,8 +951,8 @@ void ACE_CDR::Fixed::normalize (UShort min_scale)
> >
> >  ACE_CDR::Fixed ACE_CDR::Fixed::from_string (const char *str)
> >  {
> > -  const bool negative = str && *str == '-';
> > -  if (negative || (str && *str == '+'))
> > +  const bool negative = *str == '-';
>
> Coverity is still flagging it as an issue. If there's a way to specify a
known precondition, could you do that, else fix the code to check null str
first and bail before doing content checks?

mitza-oci: I went back to coverity's web interface and I couldn't see any outstanding issues in CDR_Base.  I'm going to consider this one closed unless I hear otherwise.

shuston: Sounds good – thanks for following thi sup, Adam!

PR #106, opened 2015-06-29
Added support to use the wait operations of the ACE process manager u…

jwillemsen:
…sing std::chrono::duration as alternative to ACE_Time_Value

    * ACE/ace/Process_Manager.h:
    * ACE/tests/Process_Manager_Test.cpp:

### DIFF
--- a/ACE/tests/Process_Manager_Test.cpp
+++ b/ACE/tests/Process_Manager_Test.cpp
@@ -437,7 +437,11 @@ run_main (int argc, ACE_TCHAR *argv[])
                               mgr,
                               1,
                               4);
+#if defined (ACE_HAS_CPP11)
+  result = mgr.wait (0, std::chrono::seconds (4), &exitcode);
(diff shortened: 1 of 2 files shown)


ACE/tests/Process_Manager_Test.cpp
```diff
@@ -437,7 +437,11 @@ run_main (int argc, ACE_TCHAR *argv[])
                               mgr,
                               1,
                               4);
+#if defined (ACE_HAS_CPP11)
```
shuston: This change removes the valid tests for ACE_Time_Value-using API. Maybe revert one of these changes?

jwillemsen: The C++11 template methods do only a convert to ACE _Time_Value and call the regular operations, so even not called explicitly we do test all underlying code, that is why I changed both calls in this test to use the C++11 version.

PR #118, opened 2015-07-22
TAO_IDL: generate "if" instead of "switch" for bool-discriminated unions.

mitza-oci:
Eliminates warnings in generated code with newer clang and gcc.


jwillemsen:
This causes some new warnings, see http://buildlogs.theaceorb.nl/centos70_atcd_release_ndds510rev0/index.html

jwillemsen:
Also TAO/tests/Bug_3821_Regression/run_test.pl is now failing

mitza-oci:
I see why 3821 is failing.  The helpful comment in the tao_idl sources says that the code was only generated for enum-discriminated unions but it should be there for bool too.

PR #122, opened 2015-07-23
TAO_IDL: fixed regression test 3821 failure, cleaned up warnings.

mitza-oci:


jwillemsen:
Looks we have 2 warnings left, see http://buildlogs.theaceorb.nl/fc21_taox11_corbae_micro/

mitza-oci:
Only one warning (on 2 lines): "be/be_visitor_union/union.cpp:127:27: warning: ‘other_val’ may be used uninitialized in this function [-Wmaybe-uninitialized]"
This is a bug in GCC -- newer versions don't have it.  Do we usually change the code anyway?

jwillemsen:
We have a zero warning policy so please how the code can be changed so that the warning is gone, we see this in a lot of our builds

jwillemsen:
other_val is not initialized on line 127, so it could be it gets a random value when line 138 is not triggered

mitza-oci:
It's a bug in GCC, for example http://buildlogs.theaceorb.nl/suse132_atcd_debug uses GCC 4.8.3 has this warning.  I developed the latest changes on GCC 5.2.0 and didn't see the warning.  The problem is that (older GCC) doesn't notice that other_val is only read when has_other is true.  All code paths where has_other is true have definitely assigned a value to other_val.  Our GCC 5.2.0 build http://scoreboard.ociweb.com/doc_trunk_wiro_linux_gcc-5.2_d1o0/ shows no warnings.

I don't mind adding a simple initializer to silence this bad warning, but we have many other warnings throughout the codebase for example on clang (which is really good about warnings) at http://scoreboard.ociweb.com/doc_trunk_wiro_linux_clang_d1o0/2015_07_24_11_43_Brief.html

jwillemsen:
Please add an initializer so that at least with GCC the warnings is gone

PR #127, opened 2015-08-10
Allow spaces in the temp dir used for tao_idl's invocation of the preprocessor

mitza-oci:


TAO/TAO_IDL/driver/drv_preproc.cpp
```diff
@@ -138,8 +138,23 @@ DRV_cpp_putarg (const char *str)
...
+      ACE_TCHAR *buf = 0;
+      ACE_NEW_NORETURN (buf, ACE_TCHAR[ACE_OS::strlen (str) + 3]);
+      if (buf)
+        {
+          buf[0] = ACE_TEXT ('"');
+          ACE_OS::strcpy (buf + 1, ACE_TEXT_CHAR_TO_TCHAR (str));
+          ACE_OS::strcat (buf, ACE_TEXT ("\""));
+          DRV_arglist[DRV_argcount++] = buf;
+        }
+    }
```
jwillemsen: What when allocation has failed and buf is zero?

mitza-oci: Before this change (and in the "else" clause) a null pointer is stored in DRV_arglist which terminates the list early.  tao_idl will fail but not crash (the preprocessor doesn't have an input file).  In this case, the preprocessor still doesn't get an input file and should fail, and tao_idl still doesn't crash.

PR #141, opened 2015-09-14
Add ACE support for MacOS X 10.11 El Capitan

felixbuenemann:
This simply points to the Yosemite configs, which seems to work fine.

### DIFF
--- a/ACE/ace/config-macosx-elcapitan.h
+++ b/ACE/ace/config-macosx-elcapitan.h
@@ -1,0 +1,6 @@
+#ifndef ACE_CONFIG_MACOSX_ELCAPITAN_H
+#define ACE_CONFIG_MACOSX_ELCAPITAN_H
+
+#include "ace/config-macosx-yosemite.h"
+
+#endif // ACE_CONFIG_MACOSX_ELCAPITAN_H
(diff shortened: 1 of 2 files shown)


jwillemsen:
Looks ok, will wait for the check builds to finish

felixbuenemann:
Btw. while trying to build ACE+TAO for the r2corba gem I'm now running into lots of errors when building TAO similar to:
```
dyld: Library not loaded: libTAO_IDL_BE.dylib
  Referenced from: /Users/felix/dev/gems/r2corba/ACE/ACE_wrappers/TAO/tao/TransportCurrent/../../../bin/tao_idl
  Reason: image not found
make[1]: *** [TCS.h] Trace/BPT trap: 5
```
My guess is that the build process depends on exporting DYLD_LIBRARY_PATH across shell boundaries, which is no longer possible due to security enhancements in Mac OS X 10.11. So while it's still possible to use DYLD_LIBRARY_PATH, you can only use it within the same shell. See rust-lang/cargo#1816 for a better description of the limitations.

jwillemsen:
If you have a fix for the mentioned issue please open a new pull request

PR #146, opened 2015-09-17
Add static ACE_SSL_Context::close for deleting it.

likema:
Because ACE_SSL_Context is a unmanaged singleton, it would not
destruct without calling static close method explicitly.

ACE_SSL_Initializer can load and unload ACE_SSL_Context, but ACE_SSL_Context could not be closed without it.

### DIFF
--- a/ACE/ace/SSL/SSL_Context.cpp
+++ b/ACE/ace/SSL/SSL_Context.cpp
@@ -135,6 +135,12 @@ ACE_SSL_Context::instance (void)
 }
 
 void
+ACE_SSL_Context::close (void)
+{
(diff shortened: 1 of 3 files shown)


jwillemsen:
The ACE_SSL_Initializer::fini does call close of the unmanaged singleton, just as your new static close method does, at the end your changes don't add anything to the shutdown sequence as far as I can determine. Do I miss something?

likema:
If I want to ACE_SSL_Context without ACE_SSL_Initializer, no way to close ACE_SSL_Context.

This is why I give this PR.

PR #150, opened 2015-10-19
Update the Impl Repo for better heavy use performance

pmesnier:


TAO/orbsvcs/ImplRepo_Service/ImR_Activator_i.cpp
```diff
@@ -139,8 +139,13 @@ ImR_Activator_i::init_with_orb (CORBA::ORB_ptr orb, const Activator_Options& opt
         "ImR_Activator");
       ACE_ASSERT (! CORBA::is_nil(this->imr_poa_.in ()));
 
+      obj = orb->resolve_initial_references ("POACurrent");
+      ACE_ASSERT (! CORBA::is_nil (obj.in ()));
```
jwillemsen: Is an assert the best way to handle this? The assert will abort the process and on some embedded platforms just make the system completely unresponsive

TAO/orbsvcs/ImplRepo_Service/ImR_Activator_i.cpp
```diff
@@ -531,7 +559,7 @@ ImR_Activator_i::handle_exit_i (pid_t pid)
         {
           this->locator_->child_death_pid (name.c_str(), pid);
         }
-      catch (const CORBA::Exception &ex)
+      catch (CORBA::Exception &ex)
```
jwillemsen: Why remove const here?

TAO/orbsvcs/ImplRepo_Service/ImR_Activator_i.cpp
pmesnier: perhaps not, I was following the pattern already present.

PR #152, opened 2015-10-21
Updates for VxWorks 7 RTP and Kernel.

mitza-oci:


ACE/ace/os_include/os_pthread.h
```diff
@@ -45,9 +45,13 @@
 
 
 #if !defined (ACE_LACKS_PTHREAD_H)
+# if !defined ACE_VXWORKS || ACE_VXWORKS < 0x700
```
jwillemsen: Maybe add a macro which is set in the os config file, hopefully with a next vxworks version we can then undefine that

mitza-oci: The original code was always adding "extern C" which makes a bad assumption.  It would be nice to have a configuration macro so that only platforms that need an "extern C" would get one.  That way when there are 0 such platforms the code should be removed.  VxWorks7 seems to be the only platform that actually breaks with it there.

jwillemsen: Would there really be a platform that needs this? Probably this was in the ancient ACE_OS code and nobody tried to remove it until now

mitza-oci: See pull request #155

PR #153, opened 2015-10-22
Removed references to CIAO and DAnCE.

ops:
This package contains only ACE and TAO.

### DIFF
--- a/ACE/rpmbuild/ace-tao.spec
+++ b/ACE/rpmbuild/ace-tao.spec
@@ -1,7 +1,6 @@
 # Set the version number here.
 %define ACEVER  6.3.3
 %define TAOVER  2.3.3
-%define CIAOVER 1.3.3
 
 # Conditional build
 # Default values are
@@ -9,7 +8,7 @@
 #                    --with opt          (Optimized build)
 #                    --with zlib         (Zlib compressor)
 #                    --with bzip2        (Bzip2 compressor)
-#                    --with tao          (TAO/CIAO)
+#                    --with tao          (TAO)
 #                    --without fltk      (No ftlk support)
 #                    --without tk        (No tk support)
 #                    --without xt        (No xt support)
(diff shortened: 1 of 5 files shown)


ACE/rpmbuild/ace-tao.spec
```diff
@@ -1,7 +1,6 @@
 # Set the version number here.
 %define ACEVER  6.3.3
 %define TAOVER  2.3.3
-%define CIAOVER 1.3.3
```
jwillemsen: What about the bin/make_release.py script, that updates also this file and updates CIAOVER, any need for a change there?

ops: Right. I'll update make_release.py as well.

ACE/rpmbuild/ace-tao.spec
```diff
@@ -75,7 +74,7 @@ Group:        Development/Libraries/C and C++
 URL:          http://www.dre.vanderbilt.edu/~schmidt/ACE.html
 License:      DOC License
 %if 0%{?_with_tao:1}%{?_without_tao:0}
-Source0:      http://download.dre.vanderbilt.edu/previous_versions/ACE+TAO+CIAO-src-%{ACEVER}.tar.gz
+Source0:      http://download.dre.vanderbilt.edu/previous_versions/ACE+TAO-src-%{ACEVER}.tar.gz
```
jwillemsen: I do agree this is a better source tarball to take, but please check also the debian directory where ACE+TAO+CIAO as tarball is used, also check the make_release.py script for the change of the source. Else we need two different gz files on OBS which also wastes space

ops: OK, I'll take a look. Do we prefer .bz2 or .gz tarball?

jwillemsen: Not sure if some packaging only works with one version, maybe just use .gz

PR #156, opened 2015-11-07
Remove version specific OpenSSL methods

kroeckx:
The SSLv23_* methods are the only one that support multiple versions.

This is https://bugs.debian.org/804326

### DIFF
--- a/ACE/ace/SSL/SSL_Context.cpp
+++ b/ACE/ace/SSL/SSL_Context.cpp
@@ -245,26 +245,6 @@ ACE_SSL_Context::set_mode (int mode)
 
   switch (mode)
     {
-#if !defined (OPENSSL_NO_SSL2)
-    case ACE_SSL_Context::SSLv2_client:
-      method = ::SSLv2_client_method ();
-      break;
-    case ACE_SSL_Context::SSLv2_server:
-      method = ::SSLv2_server_method ();
-      break;
-    case ACE_SSL_Context::SSLv2:
-      method = ::SSLv2_method ();
-      break;
-#endif /* OPENSSL_NO_SSL2 */
-    case ACE_SSL_Context::SSLv3_client:
-      method = ::SSLv3_client_method ();
-      break;
-    case ACE_SSL_Context::SSLv3_server:
-      method = ::SSLv3_server_method ();
-      break;
-    case ACE_SSL_Context::SSLv3:
-      method = ::SSLv3_method ();
-      break;
     case ACE_SSL_Context::SSLv23_client:
       method = ::SSLv23_client_method ();
       break;
@@ -274,39 +254,8 @@ ACE_SSL_Context::set_mode (int mode)
     case ACE_SSL_Context::SSLv23:
       method = ::SSLv23_method ();
       break;
-    case ACE_SSL_Context::TLSv1_client:
-      method = ::TLSv1_client_method ();
-      break;
-    case ACE_SSL_Context::TLSv1_server:
-      method = ::TLSv1_server_method ();
-      break;
-    case ACE_SSL_Context::TLSv1:
-      method = ::TLSv1_method ();
-      break;
-#if defined(TLS1_1_VERSION) && (TLS_MAX_VERSION >= TLS1_1_VERSION)
-    case ACE_SSL_Context::TLSv1_1_client:
-      method = ::TLSv1_1_client_method ();
-      break;
-    case ACE_SSL_Context::TLSv1_1_server:
-      method = ::TLSv1_1_server_method ();
-      break;
-    case ACE_SSL_Context::TLSv1_1:
-      method = ::TLSv1_1_method ();
-      break;
-#endif
-#if defined(TLS1_2_VERSION) && (TLS_MAX_VERSION >= TLS1_2_VERSION)
-    case ACE_SSL_Context::TLSv1_2_client:
-      method = ::TLSv1_2_client_method ();
-      break;
-    case ACE_SSL_Context::TLSv1_2_server:
-      method = ::TLSv1_2_server_method ();
-      break;
-    case ACE_SSL_Context::TLSv1_2:
-      method = ::TLSv1_2_method ();
-      break;
-#endif
     default:
-      method = ::SSLv3_method ();
+      method = ::SSLv23_method ();
       break;
     }
 
@@ -492,16 +441,7 @@ ACE_SSL_Context::load_trusted_ca (const char* ca_file,
 
   // For TLS/SSL servers scan all certificates in ca_file and ca_dir and
   // list them as acceptable CAs when requesting a client certificate.
-  if (mode_ == SSLv23
-      || mode_ == SSLv23_server
-      || mode_ == TLSv1
-      || mode_ == TLSv1_server
-#if !defined (OPENSSL_NO_SSL2)
-      || mode_ == SSLv2
-      || mode_ == SSLv2_server
-#endif /* !OPENSSL_NO_SSL2 */
-      || mode_ == SSLv3
-      || mode_ == SSLv3_server)
+  if (mode_ == SSLv23 || mode_ == SSLv23_server)
     {
       // Note: The STACK_OF(X509_NAME) pointer is a copy of the pointer in
       // the CTX; any changes to it by way of these function calls will
--- a/ACE/ace/SSL/SSL_Context.h
+++ b/ACE/ace/SSL/SSL_Context.h
@@ -104,26 +104,9 @@ public:
 
   enum {
     INVALID_METHOD = -1,
-#if !defined (OPENSSL_NO_SSL2)
-    SSLv2_client = 1,
-    SSLv2_server,
-    SSLv2,
-#endif /* !OPENSSL_NO_SSL2 */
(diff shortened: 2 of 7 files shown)


jwillemsen:
We can't just remove the old version specific OpenSSL methods, it could be that users depend on them. Instead of removing them, please ifdef them using for exasmple the OPENSSL_NO_SSL3 define.

kroeckx:
They will be removed at some point, so you will have to fix it at some point.

If your users are making use of SSLv3, using the define will still break them.  (And it should break them because it's just not secure.)

jwillemsen:
We have users in all kind of environments we should limit the impact for them. They aren't always able to upgrade their OS and OpenSSL versions due to all kind of circumstances

kroeckx:
I'm not sure what you're saying.  The OpenSSL code just encourages bad practices, and then you have various projects doing things that doesn't make sense because they don't understand it.  So we will remove support for that and in some future OpenSSL version it's just not going to be available.

You can either start fixing this now, or wait until it's broken. But you'll have to do this at some point.

shuston:
I think it's best to do it now - the protocol restriction has been common for quite a while now. If users need to stay at older openssl APIs they can stay at older ACE versions too. As long as this is explained to users up front so they're not taken completely by surprise, I'm in favor of incorporating this request.

jwillemsen:
@kroeckx Could you document this breaking change in ACE/NEWS under changes for 6.3.4? Having thought more about this, if you add that we can merge this

ACE/ace/SSL/SSL_Context.cpp
```diff
@@ -274,39 +254,8 @@ ACE_SSL_Context::set_mode (int mode)
     case ACE_SSL_Context::SSLv23:
       method = ::SSLv23_method ();
       break;
-    case ACE_SSL_Context::TLSv1_client:
```
obruns: According to the OpenSSL/master man page for [SSL_CTX_new](https://www.openssl.org/docs/manmaster/ssl/SSL_CTX_new.html) all of the `TLSv1*` functions removed in this hunk exist in OpenSSL and `SSLv23_method()` added below is deprecated in favour of the `TLSv1*` ones.

ACE/ace/SSL/SSL_Context.cpp
```diff
@@ -274,39 +254,8 @@ ACE_SSL_Context::set_mode (int mode)
...
-    case ACE_SSL_Context::TLSv1_2_server:
-      method = ::TLSv1_2_server_method ();
-      break;
-    case ACE_SSL_Context::TLSv1_2:
-      method = ::TLSv1_2_method ();
-      break;
-#endif
     default:
-      method = ::SSLv3_method ();
+      method = ::SSLv23_method ();
```
obruns: The commit message fails to point out the implications as documented in the OpenSSL/v1.0.2 man page for [SSL_CTX_New](https://www.openssl.org/docs/man1.0.2/ssl/SSL_CTX_new.html):

> If the cipher list does not contain any SSLv2 ciphersuites (the default cipher list does not) or extensions are required (for example server name) a client will send out TLSv1 client hello messages including extensions and will indicate that it also understands TLSv1.1, TLSv1.2 and permits a fallback to SSLv3. A server will support SSLv3, TLSv1, TLSv1.1 and TLSv1.2 protocols. This is the best choice when compatibility is a concern.

If a user of ACE_SSL wants to restrict security to e.g. TLSv1.2 his only option is to set the corresponding cipher list
by getting the underlying `SSL_CTX` via `ACE_SSL_Context::context()` and pass it to the corresponding OpenSSL function.

IMO, that raises the bar to use strong encryption with ACE_SSL by default.

ACE/ace/SSL/SSL_Context.cpp
kroeckx: On Wed, Dec 16, 2015 at 12:46:23PM -0800, obruns wrote:
> According to the OpenSSL/master man page for [SSL_CTX_new](https://www.openssl.org/docs/manmaster/ssl/SSL_CTX_new.html) all of the `TLSv1*` functions removed in this hunk exist in OpenSSL and `SSLv23_method()` added below is deprecated in favour of the `TLSv1*` ones.

There are new TLS_* methods, and SSLv23_* is a  define for those
new methods.  You should never use the TLSv1_* methods.

ACE/ace/SSL/SSL_Context.cpp
kroeckx: You seem to be misunderstanding something, but I'm not sure what.

First, that part of the documentation is removed because the SSLv2
support has been removed.  That was a section about using the
SSLv2 compatible client hello.  You really don't want that since
it doesn't support important extentions, some of which were always
enabled.  So it was actually impossible to send such a client
hello in the first place.  The 0.9.8 version was the last one that
actually send such SSLv2 compatible client hellos.  It didn't even
support TLS 1.1 and it support will end the end of this month.

Second, As it point out, it will support multiple versions of the
protocol.  It's the only option that supports multiple protocols.
Like it also point out is that if you want to restrict the
supported protocols you should tell it to use the options.
There is actually an upcomming new API that lets you set a minimum
version instead of using all those settings.

This is a part that you might want to add at some point, to be
able to disable SSLv3 for instance.  That would have as effect
that if the other side only speaks SSLv3 that you refuse to talk
to it.

Third, I don't recommend using the ciphers to select a protocol,
it's not really meant for that.

Fouth, by default you will get TLS 1.2 if both side support it
instead of the insecure SSLv3.  You now get the strong security
by default.  If TLS 1.3 support is added you will also
automatically upgraded to it.

obruns: On Dec 16 15:36 -0800, Kurt Roeckx wrote:
> >      default:
> > -      method = ::SSLv3_method ();
> > +      method = ::SSLv23_method ();
> 
> You seem to be misunderstanding something, but I'm not sure what.

Thank you @kroeckx for taking the time to explain. I was indeed
confusing both the OpenSSL and the ACE_SSL APIs.

> 
> First, that part of the documentation is removed because the SSLv2
> support has been removed.  That was a section about using the
> SSLv2 compatible client hello.  You really don't want that ...

Sure. My point in the quote was that `SSLv23_*` supports anything, while
not allowing a restriction to e.g. TLSv1.2. What I had overlooked was
the hint at `SSL_CTX_set_options()` in combination with `SSL_OP_NO_SSLv3`
and friends.

> ...

> Second, As it point out, it will support multiple versions of the
> protocol.  It's the only option that supports multiple protocols.
> Like it also point out is that if you want to restrict the
> supported protocols you should tell it to use the options.

Thanks for pointing that out (also see above).

> There is actually an upcomming new API that lets you set a minimum
> version instead of using all those settings.
> 
> This is a part that you might want to add at some point, to be
> able to disable SSLv3 for instance.  That would have as effect
> that if the other side only speaks SSLv3 that you refuse to talk
> to it.

Exactly that is my key concern: I don't want to talk to a client (or
server for that matter) that tries to force insecure protocols on me.
But isn't it all there in stable OpenSSL already? I was doing this:

```cpp
ACE_SSL_Context tlsContext;
/*
 * Calls SSL_CTX_set_options() with anything but SSL_OP_NO_TLSv1_2
 */
int ret = tlsContext.filter_versions("tlsv1.2"); // TLSv1.2 only
```

Connecting to a server using the following command line fails as
expected (client wants TLSv1, server wants TLSv1.2):

    openssl s_client -connect [ip:port] -tls1

It worked the same if I had used an ACE_SSL-based client against a
`openssl s_server`.

> Third, I don't recommend using the ciphers to select a protocol,
> it's not really meant for that.

Sure, wasn't meant to read that way. Took a shortcut and didn't
reconsider my wording.

> 
> Fouth, by default you will get TLS 1.2 if both side support it
> instead of the insecure SSLv3.  You now get the strong security
> by default.  If TLS 1.3 support is added you will also
> automatically upgraded to it.

You made the key point that I hadn't seen: Using `TLSv1_2_method()`
pins TLSv1.2 without forward compatibility. Unfortunately, the way
`ACE_SSL_Context::filter_versions()` is implemented right now, it
isn't forward-compatible, either.
So this is what the new API, you referred to above, is for?

kroeckx: On Thu, Dec 17, 2015 at 03:28:37PM -0800, obruns wrote:
> Exactly that is my key concern: I don't want to talk to a client (or
> server for that matter) that tries to force insecure protocols on me.
> But isn't it all there in stable OpenSSL already? I was doing this:
> 
> ```cpp
> ACE_SSL_Context tlsContext;
> /*
>  * Calls SSL_CTX_set_options() with anything but SSL_OP_NO_TLSv1_2
>  */
> int ret = tlsContext.filter_versions("tlsv1.2"); // TLSv1.2 only
> ```

I have no idea what that does, but I assume it sets things like
SSL_OP_NO_SSLv3.

> Unfortunately, the way
> `ACE_SSL_Context::filter_versions()` is implemented right now, it
> isn't forward-compatible, either.
> So this is what the new API, you referred to above, is for?

It's now available in the master branch:
https://www.openssl.org/docs/manmaster/ssl/SSL_CTX_set_min_proto_version.html

All relevant documentation should have been updated.

PR #157, opened 2015-11-07
Bug 4207

milan-mpathix:
Bug fix for Bug 4207:

- added test case exposing the bug
- fixed memory leak in TAO/tao/Cache_Entries
- fixed semantics of ACE_Multi_Hash_Map_Manager, to preserve pointers of
  existing elements when new elements are added or removed
- Transport_Cache_Manager uses multi-hash map instead of map of indexed elements
- removed index_ from Cache_ExtId


TAO/tao/Cache_Entries_T.inl
```diff
@@ -128,6 +126,9 @@ namespace TAO
   {
     if (this != &rhs)
       {
+        if (this->is_delete_)
+          delete this->transport_property_;
```
jwillemsen: Shouldn't this->transport_property_ be set to 0?

TAO/tao/Transport.cpp
```diff
@@ -164,6 +163,8 @@ TAO_Transport::TAO_Transport (CORBA::ULong tag,
 #endif /* TAO_HAS_TRANSPORT_CURRENT == 1 */
   , flush_in_post_open_ (false)
 {
+  cache_map_entry_.entry_ = 0;
+  cache_map_entry_.int_id_ = 0;
```
jwillemsen: Shouldn't the constructor of cache_map_entry_ initialize its members to 0?

milan-mpathix: Transport::cache_map_entry_ needs to be POD, for qsort used to sort those elements in Transport_Cache_Manager_T to be able to memcpy its memory. This is the reason I avoided temptation to add any functions, including the default constructor.

TAO/tao/Cache_Entries_T.inl
milan-mpathix: The very next line in the same file assigns the value to this->transport_property_. I will add setting to 0 to stay exception neutral

jwillemsen:
Have you compiled and ran all TAO unit tests? That is something we can't do on travis/snap, maybe you can setup a branch build to show that all tests do work without problems at runtime? This change is in the core of TAO and it could be that there are unforeseen issues.

milan-mpathix:
I will re-run all the tests, and verify it compiles/runs with versioned namespaces.

milan-mpathix:
Please see the scoreboard:
http://support.mpathix.com/tao_scoreboard/index.html

TAO/tao/Transport.cpp
milan-mpathix: It appears that it would be more efficient to use 'partial_sort' instead of full sort, since we need to find only the known number of purgable entries and sort them by purging order.

I am not too sure if all platforms supported by TAO actually have std::partial_sort though, and would the default assumption be that it exists, or it doesnt (ACE_HAS_PARTIAL_SORT or ACE_LACKS_PARTIAL_SORT...)

In any case, by not using qsort, I can also introduce default constructor to HASH_MAP_ENTRY_REF class, as suggested by @jwillemsen.

jwillemsen: I would assume that by default a platform has it, only platforms that really lack it should get a ACE_LACKS_STD_PARTIAL_SORT

milan-mpathix:
All patches for this pull request are complete. Please review.

PR #158, opened 2015-11-10
ImR deadlock resolution, AV service compiler warning, Unbounded_Set of refcounted objects fix

pmesnier:


TAO/bin/imr_tests.lst
```diff
@@ -0,0 +1,58 @@
+# $Id$
```
jwillemsen: This will trigger a fuzz error, $Id$ is from the old svn repo

TAO/orbsvcs/ImplRepo_Service/AsyncAccessManager.cpp
```diff
@@ -283,11 +283,27 @@ AsyncAccessManager::update_status (ImplementationRepository::AAM_Status s)
 }
 
 void
-AsyncAccessManager::activator_replied (bool success)
+AsyncAccessManager::activator_replied (bool success, pid_t pid)
 {
   if (success)
     {
-      this->update_status (ImplementationRepository::AAM_WAIT_FOR_RUNNING);
+      if (pid == 0)
```
jwillemsen: Shouldn't this be ACE_INVALID_PID?

pmesnier: No, pid use is optional inside the locator based on context. Early on in this generation of the ImR pids were not used at all. To distinguish between "pid expected but not valid" and "pid value not available" I use 0 for the latter case.

jwillemsen: Ok, hopefully there is no OS where pid 0 is a valid value

pmesnier: While pid 0 may be a kernel pager or scheduler process, using 0 to mean "no pid" is a common thing. Consider posix fork() where the function returns 0 in the child process or -1 (invalid pid) on an error.

jwillemsen:
Looks we missed one warning, @pmesnier , can you check? http://buildlogs.theaceorb.nl/win_mingw64_64_atcd/index.html

pmesnier:
Strange that it flagged only that use, there are other examples of logging pids throughout that and other sources.

I'm preparing some other ImR changes, I'll address this when I commit those.


> On Nov 18, 2015, at 7:40 AM, Johnny Willemsen <notifications@github.com> wrote:
> 
> Looks we missed one warning, @pmesnier , can you check? http://buildlogs.theaceorb.nl/win_mingw64_64_atcd/index.html
> 
> —
> Reply to this email directly or view it on GitHub.
> 

--
Phil Mesnier
Principal Software Engineer and Partner,   http://www.ociweb.com
Object Computing, Inc.                     +01.314.579.0066 x225

PR #159, opened 2015-11-12
Future Airborne Capability Environment (FACE) configurations for ACE

mitza-oci:
The Open Group publishes the FACE spec at http://www.opengroup.org/face, which
includes two safety profiles known as "base" and "extended."  These profiles
define a subset of platform functions which may be called by the application.
Since they are defined in terms of POSIX, only POSIX-like platforms can be used
for safety profiles.  We are using Linux and LynxOS-178 as our test platforms.

There are a number of changes to ACE involved in safety profiles, but the main
themes are replacing unsafe functions like sprintf with safer alternatives like snprintf
and controlling dynamic allocation with a user-replaceable allocator.

TAO does not build against a safety-enabled ACE.  This may be an interesting
future project but is not in scope for our current project.  In case anyone is
interested, OpenDDS does now have its own safety profile which is a subset (kind
of like minimum CORBA or CORBA/e compact/micro) that uses safety-enabled ACE.


jwillemsen:
I see that our X11 builds fail on Windows after this merge, we do use versioned namespaces, see http://buildlogs.theaceorb.nl/win_msvc14_32_ciaox11/

mitza-oci:
Will a normal ACE build with versioned namespaces enabled fail the same way?  Or is the problem only visible when a versioned ACE build is used in a certain way in an external project?

jwillemsen:
Looks it is related to Windows, linux with versioned namespaces works

mitza-oci:
OK, I'll try a vc14 build with versioned namespace.

mitza-oci:
See pull request #160.

jwillemsen:
I see a new warning with MinGW, see http://buildlogs.theaceorb.nl/win_mingw64_32_atcd/index.html

mitza-oci:
Some issues were fixed in pull request #162, I was focusing on compile errors and not warnings.  I'll have another round of changes soon (probably tomorrow).

mcorino:
These changes seem to have completely broken domain name resolution on Android platforms. Our TAO(X11) Android tests now all always fail unless we force them to use IP addresses instead of host names. Building the same tests against the x.3.3 release shows no failures.

mitza-oci:
Do you have IPv6 enabled?  Does it work if you define ACE_LACKS_GETNAMEINFO and ACE_LACKS_GETADDRINFO?  This should essentially give you the same code as 6.3.3, but it will still use functions that have been removed (not just deprecated) by POSIX.
I don't see any android runtime testing on the scoreboard.  It would be good to have some if you can contribute it.

mcorino:
Yes, IPv6 support is enabled in the build.
We're good for now as we determined 6.3.3 is the stable release we can work with.
We'll might get back to this when a business case arises.

PR #184, opened 2016-01-05
VxWorks support and getmacaddress

mitza-oci:


jwillemsen:
Looks we have a bugzilla issue related to this, http://bugzilla.dre.vanderbilt.edu/show_bug.cgi?id=3838, can you update that?

PR #214, opened 2016-03-16
Fix for new persistence failures that appeared after earlier commits.

pmesnier:
These failures are actually quite old (circa 2013) but were masked. OCI TAO addressed the problem by first fixing the errant PG_Group_List_Store file lock, and then simply ignoring the "Error" of not creating a backup for empty persistence files. However this causes the FT_Naming/Failover/Persistence test to fail, so this commit now fixes that behavior, so a backup file is once again created as soon as the primary persistence file is created.


jwillemsen:
Some new compile errors, see http://scoreboard.ociweb.com/doc_master_weevil_win2003_vc71_s1d1o0/index.html

PR #228, opened 2016-04-06
Fixed ssliop corbaloc parser, when there is a comma delimited list

milan-mpathix:
... use the same logic as in TAO/tao/Transport_Connector.cpp

Allows using corbaloc:iiop:host:2222,iiop:host:2223/ObjectKey with SSLIOP.

### DIFF
--- a/TAO/orbsvcs/orbsvcs/SSLIOP/SSLIOP_Connector.cpp
+++ b/TAO/orbsvcs/orbsvcs/SSLIOP/SSLIOP_Connector.cpp
@@ -286,7 +286,7 @@ TAO::SSLIOP::Connector::corbaloc_scan (const char *endpoint, size_t &len)
        }
        len = ACE_OS::strlen (endpoint);
    }
-   else if (slash_pos != 0 || comma_pos > slash_pos)
+   else if (comma_pos == 0 || comma_pos > slash_pos)
    {
        // The endpoint address does not extend past the first '/' or ','
        len = slash_pos - endpoint;


jwillemsen:
Can you extend one of the ssliop unit tests for this?

milan-mpathix:
Hm, this patch is not related to transport cache. It only allows ssliop corbaloc to actually have multiple profiles:

orb->string_to_object ("corbaloc:iiop:host1:9999,iiop:host2:9999/ObjectKey") currently throws INVALID_OBJ_REF if the process has SSLIOP_Factory, due to the parser trying to convert usehost1:9999,iiop:host2:9999 (instead of separating into 2 separate profiles)...

I thought to update TAO/orbsvcs/tests/Security/ssliop_corbaloc... but this test is not run automatically, and timeouts in 2 out of 5 tests it performs in my setup.

jwillemsen:
I thought about a test that also validates that the ORB listens on two ports, something that could be validated through the transport cache. Could you see why that test fails and add it to the test suite so that it runs automatically?

milan-mpathix:
I fixed TAO/orbsvcs/tests/Security/ssliop_corbaloc test, and added it to TAO/bin/tao_other_tests.lst so it gets executed automatically. Please verify the entry in TAO/bin/tao_other_tests.lst for conditions required.

Notes: TAO/orbsvcs/tests/Security/ssliop_corbaloc is not what it claims to be:

- it doesn't simply test if corbaloc/ssliop works
- it starts a naming service on ssl endpoint, and tests various naming service related operations, using
   corbaloc:sslop reference to a client
- it invokes RPC to an object hosted in the client process. For this to work with SSLIOP, client needs to run ORB event loop
- I added a separate test case where client uses corbaloc:ssliop:host:PORT1,iiop:host:PORT2/NamingService. This test would fail before this pull request, due to string_to_object in client throwing INV_OBJREF regardless of naming service running at all. It can be reproduced in 6114e2b1df60ec798efd27da8bf07ce742cf3b5d.

TAO/orbsvcs/tests/Security/ssliop_corbaloc/client.cpp
```diff
@@ -1071,5 +1078,6 @@ ACE_TMAIN(int argc, ACE_TCHAR *argv[])
   if (cosnaming_client.init (argc, argv) == -1)
     return 1;
 
+  cosnaming_client.activate();
```
jwillemsen: When you add threading the main should use the ACE thread manager to wait on the worker threads to be finished to not get strange crashes at shutdown

jwillemsen:
To allow unsecure collocated calls with ssliop you have to enable that, see TAO/orbsvcs/tests/Security/Bug_2908_Regression/MessengerServer.cpp. Isn't that better than to add threading to the test (which also makes it harder to debug)

milan-mpathix:
I am not too familiar with collocated ssliop. I tried briefly to set default_collocated_decision, but the resulting client still hangs.

jwillemsen:
Integrated into #732

PR #229, opened 2016-04-08
Bug 4152 for windows and log walker 

pmesnier:


TAO/utils/logWalker/Invocation.cpp
```diff
@@ -232,13 +232,34 @@ Invocation::new_line (ostream &strm, int indent, int offset, bool add_nl, bool s
 }
 
 void
-Invocation::dump_start_line (ostream &strm, size_t indent)
+Invocation::dump_rel_time (ostream &strm, const ACE_Time_Value &tv, const ACE_Time_Value& start)
+{
+  ACE_Time_Value reltime = tv - start;
```
jwillemsen: ACE_Date_Time could do all this math for you as far as I remember

pmesnier: I originally tried using the ACE supplied functions to format the time, but they _really_ wanted to treat it as wall time since 1970, yielding inappropriate output. My intent is to display a duration, so hour will always increment, rather than wrapping at midnight.

jwillemsen: Ok, that could be, ACE_Time_Value/ACE_Date_Time has its limitations

PR #234, opened 2016-04-13
Fix compile for Solaris Studio 12.4

jodar1:
Solaris Studio 12.4 now implements symbol lookup correctly.
Added #define ACE_ANY_OPS_USE_NAMESPACE to
config-sunos5.10.h when 12.4 or later compiler is detected.

### DIFF
--- a/ACE/ace/config-suncc-common.h
+++ b/ACE/ace/config-suncc-common.h
@@ -48,5 +48,10 @@
 
 #define ACE_TEMPLATES_REQUIRE_SOURCE
 
+// Solaris Studio 12.4 implements symbol lookup correctly.
+#if defined (__SUNPRO_CC) && (__SUNPRO_CC >= 0x5130)
+#define ACE_ANY_OPS_USE_NAMESPACE
+#endif
+
 #include /**/ "ace/post.h"
 #endif /* ACE_SUNCC_COMMON_H */
--- a/ACE/ace/config-sunos5.10.h
+++ b/ACE/ace/config-sunos5.10.h
@@ -59,4 +59,9 @@
 
 #define ACE_HAS_SOLARIS_ATOMIC_LIB
 
+// Solaris Studio 12.4 implements symbol lookup correctly.
+#if defined (__SUNPRO_CC) && (__SUNPRO_CC >= 0x5130)
+#define ACE_ANY_OPS_USE_NAMESPACE
+#endif
+
 #endif /* ACE_CONFIG_H */
--- a/ACE/ace/config-sunos5.10.h
+++ b/ACE/ace/config-sunos5.10.h
@@ -59,9 +59,4 @@
 
 #define ACE_HAS_SOLARIS_ATOMIC_LIB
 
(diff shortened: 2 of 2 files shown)


mitza-oci:
Thank you for the PR.  Are you able to host a continuous integration build using this compiler (see http://www.dre.vanderbilt.edu/scoreboard/)?  The most recent Sun CC we have is "CC: Sun C++ 5.12 SunOS_sparc 2011/11/16" which appears to be "Studio" 12.3.

jodar1:
I'm sorry, but the machines I am working with are behind a corporate firewall and would not be available for continuous integration.

mitza-oci:
All of our continuous integration build slaves are behind a corporate firewall.  Rsync is used to publish results to the public web.

jodar1:
This is new to me, but I will see if I can get in contact with someone that might be able to set this up.  What are the technical requirements and are you willing to do a click through agreement?

mitza-oci:
There is a separate repository with continuous integration tools and scripts at [autobuild](https://github.com/DOCGroup/autobuild).  You can run builds via cron or with a CI tool like Jenkins.  The output of the build consists of a few text files (HTML and plaintext) that need to be made available at some public URL.  The central scoreboard that runs at Vanderbilt consolidates results from all distributed builds.

I'm not sure what a "click through agreement" means in this context.  As part of community support I wouldn't think this would be possible.  We also offer commercial support, see http://www.ociweb.com/services/commercial-product-support/.

jwillemsen:
Had a quick look at the proposed patch, but is the file config-suncc-common.h not a better candidate for this check, looks not related to the sunos version

jodar1:
I put it in the config-sunos5.10.h file because Solaris Studio 12.4 is only available for 5.10 and 5.11, but if you think it is a better fit in config-suncc-common.h that is fine by me.

jodar1:
I have pushed the update to move the fix to config-suncc-common.h.  Also, here is the response I got about continuous integration:

> Oracle SPARC systems are available for use by community. The interfaces are being extended to accomodate continuous integration needs of our partners. The best way to start the conversation is to register for use at:
> https://swisdev.oracle.com/   and click on the "Register" button
> or just email us at: swisdev_help_ww_grp@oracle.com

mitza-oci:
In reading your message I must be missing some context since I don't know what question you asked (and to whom) that resulted in the response you quote.

That's generous of Oracle to provide systems for community use.  I'm not sure how their system works and how we could integrate with it.

Will you be working on setting up an ACE build on the Oracle infrastructure?

jodar1:
What I posted was the response to this question:
> Is it possible for Oracle to contribute a continuous integration server to help support an open source project building on an Oracle OS (Solaris 5.11) using an Oracle compiler (Solaris Studio 12.4)?

The SWiSDev lab is able to provide SPARC M7 VMs with Solaris Studio 12.4.  To register one needs to be an Oracle customer, a member of the Oracle Partner Network, a University Researcer, or a University Student.
I am actually none of the above, but hopefully someone else contributing to this project qualifies.

PR #254, opened 2016-05-06
In kfreebasd atop is a macro which we need to undefine

jwillemsen:
* ACE/ace/config-kfreebsd.h:


ops:
There is a much better way handle system functions which are defined as macros, see 
[https://github.com/DOCGroup/ACE_TAO/blob/master/ACE/ace/OS_NS_stdio.h]

PR #255, opened 2016-05-09
feature/cross-mingw

likema:
Fix cross building only ACE by mingw in Ubuntu 12.04 and 14.04

### DIFF
--- a/ACE/ace/CDR_Base.inl
+++ b/ACE/ace/CDR_Base.inl
@@ -143,14 +143,10 @@ ACE_CDR::swap_8 (const char* orig, char* target)
     bswap_64 (*reinterpret_cast<uint64_t const *> (orig));
 #elif (defined (__amd64__) || defined (__x86_64__)) && defined(__GNUG__) \
     && !defined(ACE_LACKS_INLINE_ASSEMBLY)
-  ACE_REGISTER unsigned long x =
-    * reinterpret_cast<const unsigned long*> (orig);
-#if defined (__MINGW64__)
-  asm ("bswap %1" : "=r" (x) : "0" (x));
-#else
+  ACE_REGISTER ACE_UINT64 x =
+    * reinterpret_cast<const ACE_UINT64*> (orig);
   asm ("bswapq %1" : "=r" (x) : "0" (x));
-#endif /* __MINGW__ */
-  *reinterpret_cast<unsigned long*> (target) = x;
+  *reinterpret_cast<ACE_UINT64*> (target) = x;
 #elif defined(ACE_HAS_PENTIUM) && defined(__GNUG__) \
     && !defined(ACE_LACKS_INLINE_ASSEMBLY)
   ACE_REGISTER unsigned int i =
--- a/ACE/ace/CDR_Base.inl
+++ b/ACE/ace/CDR_Base.inl
@@ -145,7 +145,11 @@ ACE_CDR::swap_8 (const char* orig, char* target)
     && !defined(ACE_LACKS_INLINE_ASSEMBLY)
   ACE_REGISTER unsigned long x =
     * reinterpret_cast<const unsigned long*> (orig);
+#if defined (__MINGW64__)
+  asm ("bswap %1" : "=r" (x) : "0" (x));
+#else
   asm ("bswapq %1" : "=r" (x) : "0" (x));
+#endif /* __MINGW__ */
   *reinterpret_cast<unsigned long*> (target) = x;
 #elif defined(ACE_HAS_PENTIUM) && defined(__GNUG__) \
     && !defined(ACE_LACKS_INLINE_ASSEMBLY)
--- a/ACE/ace/SSL/SSL_SOCK_Acceptor.cpp
+++ b/ACE/ace/SSL/SSL_SOCK_Acceptor.cpp
@@ -133,7 +133,7 @@ ACE_SSL_SOCK_Acceptor::ssl_accept (ACE_SSL_SOCK_Stream &new_stream,
           ACE_ASSERT (rd_handle.num_set() == 1 || wr_handle.num_set () == 1);
 #if defined (__MINGW64__)
 # pragma GCC diagnostic push
-# pragma GCC diagnostic ignored "-fpermissive"
+# pragma GCC diagnostic warning "-fpermissive"
 #endif /* __MINGW64__ */
           status = ACE::select (int (handle) + 1,
                                 &rd_handle,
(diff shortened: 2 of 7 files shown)


ACE/ace/CDR_Base.inl
```diff
@@ -145,7 +145,11 @@ ACE_CDR::swap_8 (const char* orig, char* target)
     && !defined(ACE_LACKS_INLINE_ASSEMBLY)
   ACE_REGISTER unsigned long x =
     * reinterpret_cast<const unsigned long*> (orig);
+#if defined (__MINGW64__)
```
mitza-oci: From what I can tell on MinGW-W64 targeting x86_64 (no matter the host), sizeof(long) == 4.  An 8-byte swap is not going to work on a long pointer unless it's explicitly qualified as "bswapq" instead of "bswap" at the assembly level.

mitza-oci: This code is not compiled on native MinGW due to...
ace/config-win32-mingw64.h:82:#   define ACE_LACKS_INLINE_ASSEMBLY
...if that's not active for the cross-compiler, something may need to be corrected with the configuration.

likema: What about the following?

```cpp
...
#elif (defined (__amd64__) || defined (__x86_64__)) && defined(__GNUG__) \
    && !defined(ACE_LACKS_INLINE_ASSEMBLY)
  ACE_REGISTER ACE_UINT64 x = 
    * reinterpret_cast<const ACE_UINT64*> (orig);
  asm ("bswapq %1" : "=r" (x) : "0" (x));
  *reinterpret_cast<ACE_UINT64*> (target) = x;
#elif defined(ACE_HAS_PENTIUM) && defined(__GNUG__) \
    && !defined(ACE_LACKS_INLINE_ASSEMBLY)
...
```

likema: If ACE_LACKS_INLINE_ASSEMBLY is defined, the above code will not be compiled.

mitza-oci: Correct, that's what I said in the most recent comment.  Since the code is not relevant to MinGW it should not be changed in this pull request.

I also tried building the master branch using "x86_64-w64-mingw32-g++ (GCC) 4.8.2" on Ubuntu 14.04.  This didn't get very far because the special pthread.h for MinGW defines strok_r as a macro which confuses ACE's code.  The same pthread.h in *native* MinGW doesn't have this problem.

likema: You are right. I made patches for Ubuntu 12.04/14.04 x86_64-w64-mingw32-g++. 

I don't know how to report them to Debian/Ubuntu upstream. So I made PPA for them:

https://launchpad.net/~likemartinma/+archive/ubuntu/devel

They can make ACE cross compiling successfully. 

I used them into our company product, an enterprise level heterogeneous backup and recovery suite, just like Veritas NetBackup. It works very well so far.

BTW, thanks ACE and you guys. It makes our product catching up with Veritas NetBackup in Chinese market.

likema: I commited new changes just now.

mitza-oci:
If there are changes that Ubuntu needs to make, you can file a bug at https://launchpad.net/ubuntu/+source/mingw-w64/+bugs.

I think it will be difficult to justify adding changes to ACE that won't build with the official packages.  Another option is to update the ACE headers to guard against the extra strtok_r macro coming from the platform headers.

likema:
OK. I will try to submit the mingw patches to Ubuntu later. 

Not only strtok_r, the following patches can not be bypassed by config.h

- https://gitlab.com/ubuntu-trusty/mingw-w64/commit/8c1dade8c30ffd0a3abbebda7e8e0f2ce09e2c34
- https://gitlab.com/ubuntu-trusty/mingw-w64/commit/706d83c410575bf1e6096d36a5dec40d04ac9c0f
- https://gitlab.com/ubuntu-trusty/mingw-w64/commit/403a71e6bab2626ef50b409181cc7e26dc2e6ab6

likema:
No matter what issues that mingw cross deb have, this PR contains ACE issues.

likema:
mingw patches is submitted

- https://bugs.launchpad.net/ubuntu/+source/mingw-w64/+bug/1580735
- https://bugs.launchpad.net/ubuntu/+source/mingw-w64/+bug/1580737

mingw-w64 are so much different in each Ubuntu distribution that I am not sure whether they will be accepted by upstream.

mitza-oci:
Looks like the Debian/Ubuntu maintainers may need to bring in a change from the upstream MinGW source: See 9488f917d4008775041280abea5fd6263dca79e8 in git://git.code.sf.net/p/mingw-w64/mingw-w64

Since that may take a lot of time, and it's a process we don't control, I'd suggest moving ahead with changing ACE headers to work with the unmodified Ubuntu packages.

likema:
The **ace/config.h** for mingw cross compilation:

```cpp
#ifndef _FILE_OFFSET_BITS
#       define _FILE_OFFSET_BITS 64
#endif /* _FILE_OFFSET_BITS */

#define DISABLE_TIME_REENTRANT 1

#if defined(_WIN32) && !defined(_WIN64)
#   ifndef _USE_32BIT_TIME_T
#       define _USE_32BIT_TIME_T
#   endif /* _USE_32BIT_TIME_T */
#endif /* define(_WIN32) && !defined(_WIN64) */

#define ACE_LACKS_GRP_H

#include "config-win32.h"
```

**DISABLE_TIME_REENTRANT** is for my PPA mingw-w64.

ops:
It is not necessary to patch mingw (or any other platform) due to strtok_r being defined as a macro. There's a nice pattern how to handle this case, see https://github.com/DOCGroup/ACE_TAO/blob/master/ACE/ace/OS_NS_stdio.h

likema:
It can not be solved just by changing ACE headers. e.g.

localtime_r is defined as a macro in mingw, but ACE_OS::localtime_r is a function.

Not only localtime_r but also other time *_r functions.

likema:
In fact , I haven't encountered any errors by strtok_r.

ops:
Check how this is handled if clearerr is a macro

likema:
Even adopt OS_NS_stdio.h pattern, any code using time *_r functions may undefined *_r macro before, e.g. 

```cpp
#undef localtime_r

ACE_OS::localtime_r(....)
```

That will bring much inconvenience for coding.

likema:
The key problem is that *_r macros has infected many system C headers in mingw.

ACE/ace/SSL/SSL_SOCK_Acceptor.cpp
```diff
@@ -131,11 +131,18 @@ ACE_SSL_SOCK_Acceptor::ssl_accept (ACE_SSL_SOCK_Stream &new_stream,
         {
           // Must have at least one handle to wait for at this point.
           ACE_ASSERT (rd_handle.num_set() == 1 || wr_handle.num_set () == 1);
+#if defined (__MINGW64__)
+# pragma GCC diagnostic push
```
jwillemsen: Why is this needed?

ops:
@likema Could you test my [repo](https://github.com/ops/ACE_TAO) if that fixes your problem with macros?

jwillemsen:
Change in ace/OS_NS_sys_stat.h has been merged to master as part of a different pull request. I am not confident that the other changes are necessary, so closing.

PR #256, opened 2016-05-11
feature/cross-aarch64-mips64el

likema:
Fix compilation warning in

* Ubuntu 14.04/14.10 aarch64 cross compilation.
* Native compilation in Linux mips64el

### DIFF
--- a/ACE/ace/config-linux.h
+++ b/ACE/ace/config-linux.h
@@ -307,7 +307,7 @@
 # define ACE_HAS_STRBUF_T
 #endif
 
-#if defined (__ia64) || defined(__alpha) || defined (__x86_64__) || defined(__powerpc64__)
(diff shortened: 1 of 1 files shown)


jwillemsen:
Would you be able to contribute an autobuild that shows that without this patch this doesn't work, see https://github.com/DOCGroup/autobuild for the autobuild repository?

likema:
Thanks.

I read autobuild but I don't understand how to add aarch64 and mips64el into it.

mips64el needs cross compiler but no linux distributions supply it. I made it by crosstool-ng.

jwillemsen:
With autobuild the idea is that you run a regular build job that compiles ACE+TAO and post the build logs on a public webserver. The logs are then used by the integrated scoreboard to keep track of compile and test results. The autobuild tooling makes it easier to setup a build, prettify the logs, and combine all in a single integrated view.

PR #264, opened 2016-06-17
Removed include of ace/Auto_Ptr.h where not needed

mitza-oci:


jwillemsen:
Oops, merged by accident, when wanting to just review

jwillemsen:
Looks minor changes, let us just watch master this weekend

mitza-oci:
OK, I have built all of ACE+TAO with tests/examples on Ubuntu.  I also built TAO_ACE.mwc on vc14.  If anything fails it will be in sections of code I don't have enabled like SSL.

jwillemsen:
When C++11 is enabled we can use std::unique_ptr on most places and guard the code with some ifdefs to get rid of most of the warnings.

mitza-oci:
Yes, there are certainly a few more tasks to do in order to make ACE look reasonable on GCC 6.1 or another C++11/14-enabled compiler.  Right now I'm trying to cut down on the warnings that come from ACE's headers so they impact downstream projects too.

OpenDDS's core only has 4 classes that use auto_ptr.  I may start experimenting there with making a typedef that resolves to auto_ptr or unique_ptr depending on the configuration.

mitza-oci:
Also see PR #265.  ACE's netsvcs are not part of our normal scoreboard builds so I didn't see this in the first round of changes.

PR #265, opened 2016-06-17
Auto_Ptr.h fixup for ACE netsvcs

mitza-oci:


jwillemsen:
ACE/TAO are compiling ok again, thanks!

PR #289, opened 2016-09-05
Revert singleton changes

jwillemsen:
@mitza-oci I am creating this new pull request to revert the changes of issue #284, after your changes the Based_Pointer_Test and the AVStream tests fail on Windows and Linux.

FYI, we have ported AXCIOMA to ARM Linux using ACE/TAO 6.4.0 without having to made these changes.


jwillemsen:
@mitza-oci The checks are all passed, I would like to propose to merge this pull request so that you can retest the changes on various platforms in parallel and we have a stable master for other pending work items

mitza-oci:
The code in question is invalid C++ and should not be included in ACE.  In the past it seems like people were finding work-arounds and not actual fixes when building with GCC for ARM.  (see for example https://iotprotocols.wordpress.com/2014/11/04/how-to-build-an-iot-scenario-using-dds-on-a-raspberry-pi-with-node-jswebsocket/)

I will work on a new PR to fix this.

jwillemsen:
Maybe you can add a fuzzer to check code for invalid macro usage at the moment you have a fix

PR #296, opened 2016-09-21
Use explicit template class instantiation for ACE_*_SINGLETON_DECLARE…

huangminghuang:
… with GCC 4 and above

This prevents multiple definition of typeinfo symbols during linking,
only seen on arm-linux-gnueabihf-g++.


mitza-oci:
@jwillemsen this is the updated fix for the problem in #284 that was reverted in #289 

We have a scoreboard build for Raspberry Pi that is currently failing, and should be working once we merge this change.

jwillemsen:
Looks DAnCE/CIAO are broken after these changes

TAO/orbsvcs/orbsvcs/AV/AV_Core.cpp
```diff
@@ -1146,8 +1146,11 @@ TAO_AV_Core::get_control_flowname(const char *flowname)
...
-#if defined (ACE_HAS_EXPLICIT_STATIC_TEMPLATE_MEMBER_INSTANTIATION)
-template ACE_Singleton<TAO_AV_Core, ACE_Null_Mutex> *ACE_Singleton<TAO_AV_Core, ACE_Null_Mutex>::singleton_;
-#endif /* ACE_HAS_EXPLICIT_STATIC_TEMPLATE_MEMBER_INSTANTIATION */
 
 TAO_END_VERSIONED_NAMESPACE_DECL
+
+ACE_BEGIN_VERSIONED_NAMESPACE_DECL
+
+ACE_SINGLETON_TEMPLATE_INSTANTIATE(ACE_Singleton, TAO_AV_Core, ACE_Null_Mutex)
+
```
jwillemsen: Why is this one within an ACE begin/end versioned namespace block and all others not?

huangminghuang:
Because this one tried to put the singleton inside the TAO namespace and it didn’t work with explicit instantiation. Others didn’t use namespace in the first place; therefore turning on the versioned namespace have no effect on them.

Huang-Ming

mitza-oci:
On 9/24/2016 3:25 PM, Johnny Willemsen wrote:
>
> Looks DAnCE/CIAO are broken after these changes
>

We'll take a look at DAnCE and CIAO.  Our build VM that has those builds 
is offline (coming back soon) so I'll build locally.

TAO/orbsvcs/orbsvcs/AV/AV_Core.cpp
mitza-oci: With versioned namespaces enabled, the template ACE_Singleton belongs to the namespace ACE_x_y_z.  The explicit instantiation should be in that same namespace.  Looks like GCC also allows it to appear in the global namespace though I'm not sure this is technically legal C++.  I think the only instances of this we have are in tests or examples so I'm not too concerned but we could change them for consistency.

PR #298, opened 2016-09-21
Better Mac OSX support

huangminghuang:
The pull request bypasses the limitation of the default System Integration Protection mode introduced since Mac OS X El Captitan and adds config-macosx-all.h/platform-macosx-all.GNU files to detect the Mac OS version.


jwillemsen:
Why not remove -all from the filename, looks much cleaner for the user

mitza-oci:
> Why not remove -all from the filename, looks much cleaner for the user

We discussed that but the repository already has config-macosx.h which is for 10.2, so it might be confusing to users.

jwillemsen:
10.2 is ancient and probably nobody includes that file directly, why not rename it to include the macosx name of 10.2?

PR #299, opened 2016-09-22
Remove null reference checks

huangminghuang:


jwillemsen:
When you remove these checks probably bugzilla 2234 will be introduced back, see http://bugzilla.dre.vanderbilt.edu/show_bug.cgi?id=2234

mitza-oci:
I don't think it's possible for a conforming C++ compiler to treat this any differently.  Taking the address of an object will always yield non-0.  If it's 0 then there's a C++ extension at work that allows binding a reference to null.  Perhaps we should discuss more on bugzilla, but this change itself seems right to me and will eliminate a ton of warnings from clang and newer GCCs.

jwillemsen:
Yes, the code does look strange and doesn't confirm to C++ to my idea, but probably another solution for 2234 has to be found

mitza-oci:
I haven't seen the 2234 regression test fail with this change.  I ran it locally and we have it in TAO 2.2a.

jwillemsen:
That sound great, did see these warnings some weeks ago but didn't had time to revert and retest with 2234

PR #316, opened 2016-10-26
Added a guard to prevent multiple threads accessing class member simu…

ops:
…ltaneously

### DIFF
--- a/TAO/orbsvcs/orbsvcs/Event/ECG_Mcast_EH.cpp
+++ b/TAO/orbsvcs/orbsvcs/Event/ECG_Mcast_EH.cpp
@@ -116,9 +116,7 @@ TAO_ECG_Mcast_EH::shutdown (void)
 int
 TAO_ECG_Mcast_EH::handle_input (ACE_HANDLE fd)
 {
-  // We don't use ACE_GUARD here. The lock can be released
-  // before calling handle_input ()
-  this->lock_.acquire ();
+  ACE_GUARD_RETURN (TAO_SYNCH_MUTEX, ace_mon, this->lock_, -1);
   size_t const subscriptions_size = this->subscriptions_.size ();
   for (size_t i = 0; i != subscriptions_size; ++i)
     {
@@ -129,7 +127,6 @@ TAO_ECG_Mcast_EH::handle_input (ACE_HANDLE fd)
           return this->receiver_->handle_input (*socket);
         }
     }
-  this->lock_.release ();
   return -1;
 }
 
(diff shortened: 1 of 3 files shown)


jwillemsen:
During the lock is hold a lot of code is done, including calls to the reactor and socket operations, wouldn't it be better to rework the code so that the lock is only hold at the moment the class members are modified, not when you make calls to other ACE parts

ops:
@jwillemsen Could you review the patch again?

jwillemsen:
@ops On my list, but currently no time to review in detail

TAO/orbsvcs/orbsvcs/Event/ECG_Mcast_EH.cpp
```diff
@@ -97,6 +97,7 @@ TAO_ECG_Mcast_EH::shutdown (void)
   // Indicates that we are in a shutdown state.
   this->receiver_ = 0;
 
+  ACE_GUARD_RETURN (TAO_SYNCH_MUTEX, ace_mon, this->lock_, -1);
```
ops: OK

TAO/orbsvcs/orbsvcs/Event/ECG_Mcast_EH.cpp
```diff
@@ -197,6 +200,7 @@ int
 TAO_ECG_Mcast_EH::delete_unwanted_subscriptions (
     Address_Set& multicast_addresses)
 {
+  ACE_GUARD_RETURN (TAO_SYNCH_MUTEX, ace_mon, this->lock_, -1);
```
pmesnier: Hi Olli,
Can you explain your motivation for this change? It looks like you've seen situations where handle input and delete_unwanted_subscriptions have raced change the state of th subscription array. Is that correct?

If so, I first suggest editing the title of this pull request to be more specific. Second, Johnny is correct, the locks shouldn't be held  during invocations that might be blocking. I have more to say, but let me know if I am missing the point first.

ops: Hi Phil,
Yes, your assumption about the race is correct.

pmesnier: OK, so here's my additional thoughts:

The real problem is that removing members from the subscription array reorders the array. So it seems the delete_unwanted_members should be refactored so that it is a two-phase process: First, while under the lock, collect the entries from subscription array that need to be removed to a new stack-based list, reordering the subscriptions array at will. Then after releasing the lock on the subscription array, you can safely clean up the resources of the removed members.

While in handle_input, use the guard to lock the subscription array while searching, but when the correct receiver is found, use guard.release() to unlock the array again before calling the receiver's handle input.

Note that this still makes it possible for a subscription member to be closed down while its associated peer data is being received, but that's a separate issue. 

As for the pull request title, how about something like "Synchronize access to the ECH Mcast subscription list?"

ops: Thanks for your ideas. I think I'll close this PR and make a new based on your proposal. And I need to make a separate PR for TTY changes that slipped in.

jwillemsen: Thanks @pmesnier

PR #320, opened 2016-10-31
Attempt to test FACE configuration

jwillemsen:
* .travis.yml:

PR #322, opened 2016-11-01
ace/Synch.h sanitizations

esohns:
ACE implements abstractions for synchronization primitives, supporting platform features and emulating functionality that may not be available natively (e.g. recursive mutexes, absolute/relative timeouts, thread mutexes, (DLL-)singletons, etc). This abstraction is mostly implemented using macros in `ace/Synch_Traits.h` that translate into type declarations specifying the traits of 'strategized' class templates (e.g. `ACE_Task_T`, `ACE_GUARD`) as well as regular class members and variables (e.g. `ACE_SYNCH_MUTEX`). I.e., rather than specifying a functionality by class, developers use a macro instead (e.g. `ACE_SYNCH_MUTEX` for `ACE_Thread_Mutex`). [Notice also how ACE might start to integrate `std::mutex` using (sets of) configuration macros such as `ACE_SYNCH_USE_STD_MUTEX` this way]
These type declarations are sometimes templates themselves, leading to issues on some platforms, when the compiler cannot see their definitions at build time. This is being addressed by the meta-header `ace/Synch.h`, which `#include`s all necessary declarations in the 'right' order. However, this approach introduces subtle issues (specifically, 'multiple definitions' linkage trouble for the template specializations, using MSVC).
The proposed solution is two-fold:
- replace `ace/Synch.h` with `ace/Synch_Traits.h` in all ACE headers, and `#include`ing `ace/Synch.h` in each translation unit (`.cpp`) using any synchronization functionality at all (Note that this is a requirement, as compilers generally will not translate under-defined template classes). Note how this measure effectively moves the issue into each translation unit, where it can be solved individually
- use the concept of `extern template`s (introduced in _C++11_) whenever possible

This way, the two-step approach (i.e. pre-processing + strategized primitives) towards abstracting synchronization described above becomes less error-prone and can be developed more easily.

This pull request addresses point one


esohns:
pulled #320 and resolved the merge conflicts

ACE/protocols/ace/RMCast/RMCast.mpc
```diff
@@ -1,13 +1,24 @@
...
   avoids = ace_for_tao
-  sharedname   = ACE_RMCast
+  sharedname = ACE_RMCast
   dynamicflags += ACE_RMCAST_BUILD_DLL
 
   specific {
     install_dir = ace/RMCast
   }
 
+  // *TODO*: this is essentially a workaround to be removed ASAP. On modern
```
jwillemsen: We also have other compilers than msvc on windows, probably this has to be specific to the microsoft compiler

jwillemsen:
Can you check the systems from which you commit, one seems to use an invalid email address.

ACE/protocols/ace/RMCast/RMCast.mpc
esohns: Yes. Any suggestion how to do this ?

mitza-oci:
Please check .gitignore, the changes don't make sense to me.  The semantics of [] are "pick one character from the list."

ACE/protocols/ace/RMCast/RMCast.mpc
mitza-oci: See the MPC documentation for how to write a section that only applies to MSVC.

esohns:
the changes to .gitignore reflect the fact that the rule `*.[ip]db` did not filter the `.ipdb` files generated by recent MSVC compilers. I don't know whether this introduces inconveniences on other systems. Thank you for bringing this to my attention.

ACE/protocols/ace/RMCast/RMCast.mpc
esohns: Thank you. Fixed in commit 199cd41

ACE/TAO/rules.tao.GNU
```diff
@@ -0,0 +1 @@
+../../TAO/TAO_IDL/rules.tao.GNU
```
jwillemsen: Is this necessary?

jwillemsen:
I see that you add an include of ace/Synch.h to a lot of files, this means that you pull in all possible headers for synchronization which looks an overkill, a lot of time has been spend on making it possible to just include the wrappers needed, now you include everything again, resulting in more compile-time and potential footprint. Is there no other solution?

ACE/TAO/rules.tao.GNU
esohns: No. Removed in commit 7723fd6

esohns:
ace/Synch.h may become obsolete altogether at some point. It was being pulled in indirectly via other headers before this change, so I doubt that the compilation time has increased significantly, and the footprint probably hasn't changed.
Now that it is included directly only by those translation units requiring synchronization, changes are easier to gradually implement what you expressed: to effectively only include the wrappers that are actually required. The next step in this direction will require more thought and effort however. Perhaps new (or intermediate-) (meta-)headers are in order to separate and encapsulate the functionality (e.g. into mutex/condition, process/thread scope, ...), then ace/Synch.h could be replaced, or modified by (ACE-) configuration macros, compiler flags or other means, to include only necessary headers.

PR #334, opened 2016-11-20
Fix for TTY_IO parity settings

ops:


jwillemsen:
Do you have some background information for the changes?

ops:
It was not possible to set parity settings correctly due to bugs in the TTY_IO code.

jwillemsen:
@ops You are using this patch already in your system without problems?

ops:
Yes, tested on Linux

PR #335, opened 2016-11-23
Synchronize access to the ECG_Mcast_EH subscription list

ops:


TAO/orbsvcs/orbsvcs/Event/ECG_Mcast_EH.cpp
```diff
@@ -97,30 +98,35 @@ TAO_ECG_Mcast_EH::shutdown (void)
...
 TAO_ECG_Mcast_EH::handle_input (ACE_HANDLE fd)
 {
+  ACE_GUARD_RETURN (TAO_SYNCH_MUTEX, ace_mon, this->lock_, -1);
   size_t const subscriptions_size = this->subscriptions_.size ();
   for (size_t i = 0; i != subscriptions_size; ++i)
     {
       ACE_SOCK_Dgram_Mcast *socket = this->subscriptions_[i].dgram;
       if (socket->get_handle () == fd)
         {
+          this->lock_.release ();
```
pmesnier: You should use   ace_mon.release()  instead of lock_.release() since ace_mon is currently in control of the lock status.

pmesnier:
Hi Olli,

Apart from the nit regarding handle_input, I think this is probably good enough.

One other thing you might consider is that in the spirit of "don't repeat yourself" you could refactor the shutdown method to move all of the subscriptions to a list like you do in delete_unwanted_subscriptions, then you can refactor the unregister/socket closure into a single function called by by both shutdown and delete_unwanted_subscriptions.

Thanks,
Phil

ops:
Hi Phil,

I thought refactoring cleanup code but decided to minimize changes this time. Maybe I do it with another PR.

Thanks for your help!

pmesnier:
In this case, factoring the common code into its own private or protected method seems like reducing complexity. Also given that you are touching both of those methods to begin with, now seems like the perfect time to do it. However, given that you didn't create the duplication, I understand your reluctance to go further that you have so far.

I'm fine with accepting this after the checks all pass.

jwillemsen:
I would be in favor of refactoring the shutdown method also right now. The code nows calls into the reactor with a lock hold, could potentially trigger other issues

pmesnier:
Upon a second look, I see that shutdown also does this prior to the synchronized block:
    // Indicates that we are in a shutdown state.
    this->receiver_ = 0;

Now, if it is possible for the shutdown to happen concurrently with any of the other methods, which your use of a lock implies. I suggest that rather than locking the subscription clean up loop, you instead grab the lock prior to setting receiver_ to 0, then in all the places where receiver_ is used, do a check for receiver_ is null while holding the lock.

Otherwise it could be possible that while handle input holds the lock to iterate over the subscriptions, another thread calls shutdown, nulling the receiver_ then waiting on the lock to clean up the sockets. Then the thread in handle input finds the subscription it wants and blammo - a null pointer reference when it goes to use the receiver_.

However if shutdown uses the lock to guard the receiver_ = 0 assignment, and then handle_input checks to see if receiver_ is null inside the guard and before the loop, then that is avoided. 

However again, if a thread is in handle_input first, grabs the lock and starts to iterate before a second thread enters shutdown, then at the point where handle_input releases the lock before using the receiver_, a race happens. Thread 1 could use the receiver OK, or thread 2 could take the lock and null the receiver prior to thread1 making the call.

I have an idea about that too, but it's thanksgiving and I have to start preparing the house.

ops:
Good catch Phil! What about adding a new member variable shutdown to control shutdown state and remove receiver_ = 0 assignment from shutdown() ?

pmesnier:
Hi Olli,

Looking only at the ECG_Mcast_EH.cpp code, there is still a race because in addition to the receiver reference, handle input also uses one of the members of the subscriptions list as well. So for the same reason setting release to 0 is bad, so is deleting the subscriptions. Likewise with the delete_unwanted_subscriptions method, it could whack a subscription holding the socket reference currently used by handle_input.

So here's my solution:

1. Make the subscription values reference counted and add a destructor to the subscription struct which is responsible for the dgram close and delete. Use a boost shared_ptr or an ACE_Refcounted_Auto_ptr as the array member and as the temporary value used inside of handle_input.

2. Simplify shutdown() and delete_unwanted_subscriptions() to just unregister the specific subscription(s) from the reactor and release the reference. This way if a particular subscription happened to send a dgram concurrent with the desire to get rid of that subscription, there's no conflict.

I think that gets rid of all the race conditions.

jwillemsen:
Please don't add a dependency on boost, that causes a dependency on all platforms. Regret that we can't use C++11 everywhere, that would simplify a lot of code but only some platforms/compilers now have adequate C++11 support.

pmesnier:
Hi Olli,

Are you still working on this? Do you need more from  me to complete the patch? I agree with Johnny, the solution should be based on ACE components rather than boost.

ops:
Hi Phil,
Yes, but something more urgent came up. Hopefully, I'll get back on this in a week or two.

jwillemsen:
@ops I am closing this for the moment, please open a new PR when you have worked on it again

PR #346, opened 2017-01-03
Added ACE_DES_FREE_THIS which doesn't check the pointer for nil becau…

jwillemsen:
…se this can't be nil, fixes compiler warnings with newer GCC versions

    * ACE/ace/Global_Macros.h:
    * ACE/ace/Message_Block.cpp:
    * TAO/tao/Asynch_Queued_Message.cpp:
    * TAO/tao/Synch_Queued_Message.cpp:

### DIFF
--- a/ACE/ace/Global_Macros.h
+++ b/ACE/ace/Global_Macros.h
@@ -254,6 +254,13 @@ ACE_END_VERSIONED_NAMESPACE_DECL
       } \
    while (0)
 
+# define ACE_DES_FREE_THIS(POINTER,DEALLOCATOR,CLASS) \
+   do { \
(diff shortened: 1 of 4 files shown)


ACE/ace/Global_Macros.h
```diff
@@ -254,6 +254,13 @@ ACE_END_VERSIONED_NAMESPACE_DECL
       } \
    while (0)
 
+# define ACE_DES_FREE_THIS(POINTER,DEALLOCATOR,CLASS) \
```
mitza-oci: Why does it need a POINTER parameter if it's always "this"?

jwillemsen:
Yes, you are right, will update this and make a new PR

PR #360, opened 2017-01-24
ACE_HAS_GCC_ATOMIC_BUILTINS should be defined for clang

Zaffy:
Making pull request upon request.  
More info can be found on [mailing lists](http://list.isis.vanderbilt.edu/pipermail/ace-bugs/2017-January/000032.html).

### DIFF
--- a/ACE/ace/config-linux.h
+++ b/ACE/ace/config-linux.h
@@ -146,9 +146,6 @@
   // this must appear before its #include.
 # define ACE_HAS_STRING_CLASS
 # include "ace/config-g++-common.h"
-# ifdef __clang__
-#  undef ACE_HAS_GCC_ATOMIC_BUILTINS
-# endif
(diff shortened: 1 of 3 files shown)


jwillemsen:
Do you know which clang version did introduce the necessary support? For gcc we have this define dependent on the g++ version, probably something similar is necessary for clang.

Zaffy:
Honestly I don't, sorry.

But clang should be highly compatible with gcc so if the checks pass and the macro gets defined then it probably should stay defined.

I don't know what was the original reason for add such code that forcibly undefines it for clang but even if it was legit, it was back in 2010 and clang did a huge step in development.

jwillemsen:
Could you merge master into your branch and push again? I have enable clang on travis-ci as basic test to see if removing this define doesn't break basic support.

jwillemsen:
Thanks!

Zaffy:
Thank You, too! :+1:

Issue #363, opened 2017-02-08
Native Windows Condition Variables do not set errno to ETIME


mathersm:
From the documentation associated with the ACE_Condition<T>::wait(const ACE_Time_Value *abstime) method (ace/Condtion_T.h):

>`int wait (const ACE_Time_Value *abstime);`
 Block on condition, or until absolute time-of-day has passed.  If @a abstime == 0 use "blocking" <wait> semantics.  Else, if @a abstime != 0 and the call times out before the condition is signaled wait() returns -1 and sets errno to ETIME.

If native condition variables are enabled for Windows (via ACE_HAS_WTHREADS_CONDITION_VARIABLE) the Windows call ::SleepConditionVariableCS(....) is used. If the wait times out, this function will set a Windows-specific error, ERROR_TIMEOUT and not ETIME.

jwillemsen:
Be aware that the support for ACE_HAS_WTHREADS_CONDITION_VARIABLE was never completed, it is in the code, but was not enabled because we didn't finish the work, had several failing unit tests, see http://bugzilla.dre.vanderbilt.edu/show_bug.cgi?id=4146. I would not recommend you to use this unless you are willing to finish the effort.

mathersm:
Do you have any record of the work you believe to be completed or a list of tests within the suite of unit tests you believe would be affected? I'm guessing no, but some information like that would help others to quickly understand the scope of the effort involved. My lack of familiarity with the complete regression suite hampers me a little I'm afraid. Perhaps something else that may help would be a link to a scoreboard for the current master branch such that I can prune away chasing other known, yet unrelated issues.

jwillemsen:
I don't recall what failed but just that some tests failed and because this was an unfunded effort it was put on hold because I ran out of time. See http://www.dre.vanderbilt.edu/scoreboard/ for the current scoreboard.

mathersm:
Thanks for the link Johnny, that helps especially noting the current state of tests on Windows. I understand the difficulty of only having so much 'free' time.

jwillemsen:
There are so many things we would like to work on (http://bugzilla.dre.vanderbilt.edu/ has a lot wishes from the past) but lack sponsoring to do so. Just running all builds takes a lot of resources on a regular basis.

jwillemsen:
@mathersm Just send you an email

PR #364, opened 2017-02-08
Test and Potential Fix for Issue #363 Native Windows Condition Variables do not set errno to ETIME

mathersm:
- Added a really simple test for the return value and errno combination expected when a wait on a condition variable times out. Project name is Native_Condition_Variable_Test.
- Added new test onto the list for automatic testing via run_test.lst
- Added an additional case to the ACE_FAIL_RETURN macro to mutate ERROR_TIMEOUT into ETIME.
- Added an additional check at the end of the implementation of ACE_Condition<T>::wait(const ACE_Time_Value*) to mutate WIN32 errors into errno (as expected).

It might be worth considering revisiting the MSVC configuration files to enable ACE_HAS_WTHREADS_CONDITION_VARIABLE (with appropriate include statements) for Windows Vista/Server2008 and beyond.

### DIFF
--- a/ACE/tests/Native_Condition_Variable_Test.cpp
+++ b/ACE/tests/Native_Condition_Variable_Test.cpp
@@ -1,0 +1,47 @@
+
+//=============================================================================
+/**
+ *  @file    Native_Condition_Variable_Test.cpp
+ *
+ *    This program tests the support of Native Condition Variables on
+ *    Microsoft Windows. It checks that errno is correctly mutated to
+ *    ensure return and errno values are consistent across implementations
+ *    using ACE_Condition.
+ *
+ *  @author Michael Mathers <michael.p.mathers@live.com>
+ */
+//=============================================================================
+
+
+#include "test_config.h"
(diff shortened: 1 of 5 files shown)


jwillemsen:
FYI, because ACE_HAS_WTHREADS_CONDITION_VARIABLE is not defined by any of the CI or regular scoreboard builds none will test the core changes and the support for this

ACE/tests/Native_Condition_Variable_Test.cpp
```diff
@@ -0,0 +1,47 @@
+
+//=============================================================================
+/**
+ *  @file    Native_Condition_Variable_Test.cpp
```
jwillemsen: I would propose to remove Native_ from the name of this test, it just tests the condition variable behavior on any platform

mathersm: Agreed, I will probably change it to something like Condition_Variable_Timeout_Test.

PR #382, opened 2017-03-27
Fix INET_Addr reset IPV6 port problem 

huangminghuang:

PR #384, opened 2017-03-30
Fix ability to request IPv4 address from ACE_INET_Addr::set

shuston:
Previously ignored a caller's specifying "tcp"

### DIFF
--- a/ACE/ace/INET_Addr.cpp
+++ b/ACE/ace/INET_Addr.cpp
@@ -511,6 +511,9 @@ ACE_INET_Addr::set (const char port_name[],
     }
 
   int address_family = PF_UNSPEC;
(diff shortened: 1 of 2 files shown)


jwillemsen:
Steve, PR #382 seems to have introduced some test failures, would recommend to wait with merging this PR until #382 has been resolved completely given that this is in the same file

shuston:
#382, 385, 287 are working in a different part of the source file - I don't want to let this go so long I have to rebase the change, so I'm going to merge it now.

jwillemsen:
Looks we have a new warning, see http://scoreboard.ociweb.com/doc_master_ant_linux-sb_gcc_d0o1/2017_04_01_07_07_Brief.html#section_5

PR #390, opened 2017-04-04
Fix compiler issues on macOS with ACE_FACE_SAFETY* defined

huangminghuang:
The fix allows ACE to compile on macOS when ACE_FACE_SAFETY_BASE or ACE_FACE_SAFETY_EXTENDED is defined.

### DIFF
--- a/ACE/ace/Shared_Memory_Pool.cpp
+++ b/ACE/ace/Shared_Memory_Pool.cpp
@@ -123,7 +123,7 @@ ACE_Shared_Memory_Pool::commit_backing_store_name (size_t rounded_bytes,
                       "exceeded max number of segments = %d, base = %u, offset = %u\n",
                        counter,
                        this->base_addr_,
-                       offset),
+                       static_cast<unsigned int>(offset)),
                       -1);
   else
     {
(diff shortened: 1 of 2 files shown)


ACE/ace/Shared_Memory_Pool.cpp
```diff
@@ -123,7 +123,7 @@ ACE_Shared_Memory_Pool::commit_backing_store_name (size_t rounded_bytes,
                       "exceeded max number of segments = %d, base = %u, offset = %u\n",
                        counter,
                        this->base_addr_,
-                       offset),
+                       static_cast<unsigned int>(offset)),
                       -1);
```
jwillemsen: If there is a conversion warning shouldn't also a macro be used for the format specifier for offset instead of casting it to a 32bit value?

huangminghuang: The variable is of type off_t, I cannot find the corresponding format specifier macro in ACE for the type. It may be equivalent to ssize_t. However, I couldn't find any documentation to guarantee that. At least, using cast won't break any code on the platform I haven't tried yet.

PR #394, opened 2017-04-07
Fix getaddrinfo() inconsistent behavior on windows machines

huangminghuang:


jwillemsen:
This seems to cause compile errors on several builds, see http://www.dre.vanderbilt.edu/scoreboard/integrated.html. Any background information on what is inconsistent would help people to assess this change

jwillemsen:
I have reverted the change, that should get the compile green again, please open a new PR when you have sorted out the compile failures, for example http://buildlogs.remedy.nl/win_mingw64_64_acetao/index.html

PR #406, opened 2017-04-21
Inet addr fix

huangminghuang:
A new attempt to fix inet_addr::set() with workaround for AIX.

### DIFF
--- a/ACE/ace/INET_Addr.cpp
+++ b/ACE/ace/INET_Addr.cpp
@@ -371,8 +371,11 @@ ACE_INET_Addr::set (u_short port_number,
 #endif /* ACE_HAS_IPV6 && ACE_USES_IPV4_IPV6_MIGRATION */
 
 #ifdef ACE_HAS_IPV6
+
+#ifndef AIX
   if (address_family == AF_UNSPEC && ACE::ipv6_enabled ())
     address_family = AF_INET6;
+#endif /* AIX */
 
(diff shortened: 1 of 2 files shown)


ACE/ace/INET_Addr.cpp
```diff
@@ -371,8 +371,11 @@ ACE_INET_Addr::set (u_short port_number,
 #endif /* ACE_HAS_IPV6 && ACE_USES_IPV4_IPV6_MIGRATION */
 
 #ifdef ACE_HAS_IPV6
+
+#ifndef AIX
```
shuston: OS-specific ifdefs in the middle of code are generally a bad idea. Why is this specific to AIX?

huangminghuang: I don't access to any AIX machines, the only thing I know is that AIX getaddrinfo() seems to behave differently based on the scoreboard results in the past. I just need to see if the ifdef can help on AIX.

shuston: In case I wasn't clear, this change was ill-conceived. It should be reverted.

huangminghuang: Could you give any suggestion on how to actually fix the problem instead of reverting all the changes and let the bug persists?

jwillemsen: Tests still fail on AIX

PR #461, opened 2017-08-04
Fixing build for Oracle DeveloperStudio12.5 (C++ 5.14)

avl42:
These changes were necessary to get a complete successful build with the Oracle's developerstudio12.5 on Solaris 11.3 including examples and tests. 

I'm not sure whether the tests are also run, or merely compiled with a plain "make" within each of ACE_ROOT and TAO_ROOT.

I can't test it on any other platform, but I am positive that these changes would fix more than break.

Having someone test it e.g. with DeveloperStudio12.5 on Linux (just to be safe) wouldn't hurt, though.

### DIFF
--- a/ACE/ace/config-sunos5.10.h
+++ b/ACE/ace/config-sunos5.10.h
@@ -59,4 +59,9 @@
 
 #define ACE_HAS_SOLARIS_ATOMIC_LIB
 
+// Solaris Studio 12.4 implements symbol lookup correctly.
+#if defined (__SUNPRO_CC) && (__SUNPRO_CC >= 0x5130)
+#define ACE_ANY_OPS_USE_NAMESPACE
+#endif
+
 #endif /* ACE_CONFIG_H */
--- a/TAO/orbsvcs/tests/CosEvent/Basic/Random.cpp
+++ b/TAO/orbsvcs/tests/CosEvent/Basic/Random.cpp
@@ -344,12 +344,6 @@ RND_Consumer::push (const CORBA::Any &event)
 }
 
 void
-RND_Timer::push (const CORBA::Any &event)
-{
-   RND_Consumer::push(event);
-}
-
-void
 RND_Consumer::disconnect_push_consumer (void)
 {
 }
--- a/TAO/orbsvcs/tests/CosEvent/Basic/Random.cpp
+++ b/TAO/orbsvcs/tests/CosEvent/Basic/Random.cpp
@@ -344,6 +344,12 @@ RND_Consumer::push (const CORBA::Any &event)
 }
 
 void
+RND_Timer::push (const CORBA::Any &event)
+{
+   RND_Consumer::push(event);
+}
+
+void
 RND_Consumer::disconnect_push_consumer (void)
 {
 }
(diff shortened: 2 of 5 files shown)


ACE/ace/config-sunos5.10.h
```diff
@@ -59,4 +59,9 @@
 
 #define ACE_HAS_SOLARIS_ATOMIC_LIB
 
+// Solaris Studio 12.4 implements symbol lookup correctly.
+#if defined (__SUNPRO_CC) && (__SUNPRO_CC >= 0x5130)
+#define ACE_ANY_OPS_USE_NAMESPACE
+#endif
```
jwillemsen: Wouldn't it be better to include config-suncc-common.h in the base config-sunos file at the moment suncc is used? That way we only have this knowledge once

TAO/orbsvcs/tests/CosEvent/Basic/Random.cpp
```diff
@@ -344,6 +344,12 @@ RND_Consumer::push (const CORBA::Any &event)
 }
 
 void
+RND_Timer::push (const CORBA::Any &event)
```
jwillemsen: Why isn't this one in the header? This operation shouldn't be needed, maybe report a bug to Oracle?

ACE/ace/config-sunos5.10.h
avl42: a legit question... I pondered it for a while, but then I noticed, that most of the defines currently in config-suncc-common.h were already in some of the sunos* headers.    Getting each define into where it really belongs may be another year's work ;-)

TAO/orbsvcs/tests/CosEvent/Basic/Random.cpp
avl42: The question is rather:  if this method isn't ever called, then why is it declared as an override in RND_Timer class at all?

It's a very old and apt heuristic to generate certain class-data into whichever module defines the first "non-inline non-pure-virtual" method,  i.o.w.: the first method that is expected to end up in a .o file.

removing the push from RND_Timer completely would be fine with me, as well.

PS: and keep in mind to never let  operator<<=(unsigned char) in CORBA::Any move to that particular place.

jwillemsen: Just remove it from rnd_timer

ACE/ace/config-sunos5.10.h
avl42: Adding the symbol (conditionally) in sunos5.10 should be fine, because the latest compilers for which it matters, 12.4 and 12.5, don't even offer installation on any older sunos than 5.10.

config-suncc-common seems to be only included by config-linux.

TAO/orbsvcs/tests/CosEvent/Basic/Random.cpp
avl42: will do... hang on...

avl42: tried a test compile, but it seems like I've spoiled my ACE build tree with a "wrong-env" make. might take a bit longer than expected, but once the compile passes the stage where Random.o gets linked I'll prepare the commit & PR.

avl42: Done. (obviously it doesn't matter, if a class has _only_ inline methods/ctors)

PR #494, opened 2017-08-29
Add support for OpenSSL 1.1

jwillemsen:


mitza-oci:
Looks like the names of the lib files have changed
1>LINK : fatal error LNK1104: cannot open file 'ssleay32.lib'
OpenSSL installed libssl.lib and libcrypto.lib

jwillemsen:
Which platform? On Fedora 26 I don't have problems linking.

mitza-oci:
tests/SSL/Bug_2912_Regression_Test.cpp:174 error C2027: use of undefined type 'dh_st' (and similar errors)

mitza-oci:
openssl.mpb (in MPC) lists the different library names that in the past had been used by the OpenSSL project on different platforms, looks like they are consistent now

jwillemsen:
I attempted to fix Bug_2912_Regression_Test from Linux, please try again. Any proposal to handle OpenSSL v1.1 and newer on Windows? We will probably have 1.0 users for a long time

ACE/tests/SSL/Bug_2912_Regression_Test.cpp
```diff
@@ -171,10 +171,23 @@ get_dh1024 ()
   DH *dh;
 
   if ((dh=DH_new()) == 0) return(0);
-  dh->p=BN_bin2bn(dh1024_p,sizeof(dh1024_p),0);
-  dh->g=BN_bin2bn(dh1024_g,sizeof(dh1024_g),0);
-  if ((dh->p == 0) || (dh->g == 0))
-    { DH_free(dh); return(0); }
+
+  BIGNUM* p = BN_bin2bn(dh1024_p,sizeof(dh1024_p),0)
```
mitza-oci: Needs a ; (next line too)

jwillemsen:
Created a separate MPC issue for the OpenSSL library names

jwillemsen:
ACE/TAO now compiles on fedora 26 with OpenSSL 1.1 enabled. We have some failing unit tests but for example the SSLIO MT_SSLIOP runs when I use the certificates and key from the SSLIOP cert directory, probably some of the certificates we have in our tests are too old and need replacement.

PR #521, opened 2017-09-25
Fix gcc7.1 compile errors

jamespthomas:
This commit also fixes gcc 7.1 compile errors in ACE/protocols/ace/INet/Sock_IOStream.cpp:

### DIFF
--- a/ACE/protocols/ace/INet/Sock_IOStream.cpp
+++ b/ACE/protocols/ace/INet/Sock_IOStream.cpp
@@ -81,7 +81,7 @@ namespace ACE
     void Sock_OStreamBase<ACE_SYNCH_USE>::set_interceptor (
           typename buffer_type::interceptor_type& interceptor)
       {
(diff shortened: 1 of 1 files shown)


jwillemsen:
Have you used the recent x.4.5 release? As far as I know this has already been fixed in that release in a slightly different way

jamespthomas:
Hello, sadly I'm using a much older release (and it'll be difficult to bump it), so this patch was needed there. But if you are refering to 6a7b62cae7e61948f85d98e265b4caca9a8dbd88 that only appeared to fix in SSLSock_IOStream.cpp ?

jwillemsen:
No, see commit 0ad218ded544094e63ef45df03143e6e258b5665

jamespthomas:
Ah, great, missed that one! Thanks

jwillemsen:
We fixed lot of warnings and also other errors with GCC 7.1 and newer (also related to C++11/C++14/C++17) in x.4.5, could be that you need to backport more

PR #547, opened 2017-12-05
Add missing ESF header files

lockhart:
Add missing ESF header files to allow extending features in application code.


jwillemsen:
Thanks for the PR!

jwillemsen:
Applied changes manually to the rpm spec files used for OpenSuSE Build Service, RPMs delivered through there will have these files added later today

PR #551, opened 2017-12-11
Added support for cross compiling using MinGW on a Linux host.

mitza-oci:


ACE/ace/config-win32-mingw64.h
```diff
@@ -127,6 +127,8 @@
 #define ACE_HAS_NONCONST_WCSDUP
 #define ACE_ISCTYPE_EQUIVALENT ::_isctype
 
+#define ACE_HAS_PTHREAD_SIGMASK_PROTOTYPE
```
jwillemsen: Does this work with MinGW-64 on native windows?

mitza-oci: I thought the header "/usr/share/mingw-w64/include/pthread_signal.h" would be the same as in the MinGW install on windows but I haven't checked yet.

mitza-oci: I looked at the headers, I don't see any reason why it wouldn't work.

jwillemsen: There is a build on the scoreboard, that will validate this

PR #553, opened 2017-12-19
New ACE test: check if ACE_Singleton works after ACE::fini, ACE::init

mitza-oci:
Some lock types work, others don't, due to different logic per lock type
in ACE_Object_Manager::get_singleton_lock and how ACE_Singleton uses it.

### DIFF
--- a/ACE/tests/.gitignore
+++ b/ACE/tests/.gitignore
@@ -217,6 +217,7 @@
 /Sigset_Ops_Test
 /Simple_Message_Block_Test
 /Singleton_Test
+/Singleton_Restart_Test
 /SOCK_Connector_Test
 /SOCK_Dgram_Bcast_Test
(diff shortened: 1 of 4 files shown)


jwillemsen:
Do we have an issue for the problem this test reproduced? @mitza-oci are you working on a fix?

mitza-oci:
I wanted to get the test committed first to see that it compiles on all platforms.

I know of two possible fixes, either one should be sufficient and just changes a few lines.  I want to do some more research in to how the bug got there.

jwillemsen:
ok, thanks for the info

jwillemsen:
MinGW fails, see http://buildlogs.remedy.nl/win_mingw64_64_acetao, probably the mpc file changes

mitza-oci:
Right, I will fix it in the upcoming pull request.

PR #557, opened 2018-01-05
Match MPC for Wireshark 2 Support

iguessthislldo:
Added wireshark_any and wireshark2 to global.features for [Wireshark2 Support Windows to match the global.features file in MPC](https://github.com/DOCGroup/MPC/pull/47).


mitza-oci:
Merging this before the MPC change since it's just adding new settings to
global.features.  We want this change to be present when MPC is updated.

mitza-oci:
See https://github.com/DOCGroup/MPC/pull/47

PR #559, opened 2018-01-11
clang: set ACE_HAS_CPP11 based on predefined macros

mitza-oci:


jwillemsen:
I was waiting on a clang fix before enablig cpp11 for clang

mitza-oci:
The comment in Compiler_Features_32_Test.cpp says clang fixed it in 2014, but clang 5.0 still fails.

Nothing in ACE depends on this compiler feature, so the test itself can have an #ifdef for clang.  Why should this prevent users from having actual c++11 features available?

jwillemsen:
Yes, we should now patch the test so that everything else can use cpp11 with clang5, with clang4 there are other tests that also don’t compile

mitza-oci:
It works on any clang version 3.8-5, I don't think anyone would be using an older one.

jwillemsen:
Looks with clang < 5 we have a failure on travis, in clang5 a bug was fixed for this construct, maybe only enable ACE_HAS_CPP11 with clang5 or newer?

```
/Users/travis/build/DOCGroup/ACE_TAO/ACE/tests/Compiler_Features_24_Test.cpp:78:25: error: no matching function for call to 'make_f'
  FOO::o_r<FOO::A> l2 = FOO::make_f<FOO::A>();
                        ^~~~~~~~~~~~~~~~~~~
/Users/travis/build/DOCGroup/ACE_TAO/ACE/tests/Compiler_Features_24_Test.cpp:35:22: note: candidate template ignored: couldn't infer template argument ''
    friend o_r<_Tp1> make_f(Args&& ...args);
```

mitza-oci:
I didn't see this in local testing on macOS since we have a newer clang installed.  We could update the travis config to xcode 9.1 (for example OpenDDS has this).

I'm not opposed to changing config-g++-common.h if we want to support the older clang versions with -std=c++11 and not ACE_HAS_CPP11.  However I think that combination is confusing as users may write code like:
```
#ifdef ACE_HAS_CPP11
noexcept
#endif
```

jwillemsen:
Maybe then just exclude this unit test with clang 4?

PR #565, opened 2018-01-16
Bug 3940: Patch to make tao_idl ignore IDL4 annotations

okellogg:
[Bug 3940](http://bugzilla.dre.vanderbilt.edu/show_bug.cgi?id=3940) [attachment 1524](http://bugzilla.dre.vanderbilt.edu/attachment.cgi?id=1524&action=diff): Patch to make tao_idl ignore IDL4 annotations

`TAO/TAO_IDL/fe/idl.ll`
- Add rule `@[A-Za-z][A-Za-z0-9_]*[ \t]*(\([^)]+\))?` with action `break`.

`TAO/TAO_IDL/fe/idl.tab.cpp`
`TAO/TAO_IDL/fe/idl.tab.hpp`
`TAO/TAO_IDL/fe/idl.yy.cpp`
- In directory TAO/TAO_IDL execute script regen.sh using flex 2.6.0 and bison 3.0.2.

`TAO/TAO_IDL/fe/idl.ypp`
- Fix typo in comment.


jwillemsen:
Please extend the TAO IDL test to validate that annotations are ignored correctly. Better it would be to parse them so that they could be used in a backend

mitza-oci:
Lexing any legal "application of annotation" (annotation_appl, rule 225) is going to be difficult.
The regex covers essentially two cases:
```
@identifier
@identifier(anything_not_close_paren) /* with optional whitespace after identifier */
```
But also legal are:
```
@scope::identifier
@identifier /* some comment with a ) in it */ (ARG)
@identifier((1+1))
```
Note that in the latter case the lexer can't check for balanced parens, that's up to the parser.
I didn't list the case of `@identifier(name1=val1, name2=val2)` since I think it's just another instance of the same thing.
I don't think it would be difficult to have the lexer make the a token by itself and then build up the actual annotation_appl in the parser.  Then we would need to determine where it's legal to use annotation_appl, probably starting with struct and field and expanding from there.

okellogg:
@jwillemsen Okay, I'll add a test.

@mitza-oci Right now my focus is on the common usages.
Certainly the more elaborate usages you describe will require more thought.
Also, the current patch does not handle definition of new annotations (`@annotation`).

mitza-oci:
See my comment on "Should this be controlled by a command line argument?"  I think that's the most relevant of the whole bunch.  Our decision on that may influence how it's implemented.

jwillemsen:
What would be the added value of a commandline argument?

TAO/tests/Bug_3940_Regression/README
```diff
@@ -0,0 +1,16 @@
+
+
+This is a compile only test. If it compiles, it's passed.
```
mitza-oci: Since it's a compile-only test, it shouldn't need a run_test.pl or a test.cpp (build a library instead of an executable).  Or even use `custom_only=1` and just run tao_idl without compiling C++ code.

mitza-oci:
If a user types an `@` in an IDL file by mistake, the current tao_idl will reject it.  With this patch it will be silently ignored.

jwillemsen:
I would be ok with just ignoring it

okellogg:
Actually, the regex
```
@[A-Za-z][A-Za-z0-9_]*[ \t]*(\([^)]+\))?
```
prohibits this by the first `[A-Za-z]` after the `@`.

And in fact, adding a stray *@*:
```
diff --git a/TAO/tests/Bug_3940_Regression/test.idl b/TAO/tests/Bug_3940_Regression/test.idl
index d63dfff..cf15eb3 100644
--- a/TAO/tests/Bug_3940_Regression/test.idl
+++ b/TAO/tests/Bug_3940_Regression/test.idl
@@ -8,7 +8,7 @@ module test {
 
    struct structure {
       @key unsigned long key;
-      angle_t degrees;
+@     angle_t degrees;
       temperature_t celsius;
    };
```
will make tao_idl say:
```
syntax error
Error - tao_idl: "test.idl", line 11: Missing ',' between member decls of same type(?)
```

TAO/tests/Bug_3940_Regression/README
okellogg: Thanks, wilco.

jwillemsen:
The C++20 operator dot is a possible way to implement the min/max, was on the C++17 schedule but wasn't finished in time.

Just checked with TAOX11, that parses your example IDL without problems

mitza-oci:
The kind of stray `@` I was thinking of is:
```
      double w@eight;
```
With this patch the error won't occur until some C++ code attempts to use `weight`.

Overall I'm not opposed to integrating this.

jwillemsen:
We have some new conversion warnings which should be fixed, see http://buildlogs.remedy.nl/win_msvc15_opendds_debug/index.html

mitza-oci:
> fe\idl.yy.cpp(1383): warning C4267: 'argument': conversion from 'size_t' to 'int'

Based on the `#line` line, this is in the expansion of the macro `YY_DO_BEFORE_ACTION`.  In the previous version, tao_yyleng was an int.  I don't remember if this was generated by the previous version of flex or if it was hand-edited.

jwillemsen:
Or maybe the ypp needs an update @okellogg can you check?

mitza-oci:
The ypp is the other one:

idl.ypp -> yacc/bison -> idl.tab.*

idl.ll -> lex/flex -> idl.yy.*

okellogg:
Regenerating using flex-2.6.4, I get this diff:
[TAO_IDL-fe-fe_lookup.cpp-regen-using-flex-2.6.4.diff.txt](https://github.com/DOCGroup/ACE_TAO/files/1658260/TAO_IDL-fe-fe_lookup.cpp-regen-using-flex-2.6.4.diff.txt)
I believe the conversion warnings are addressed by the newer flex version.
Shall I make a new pull request for this?

jwillemsen:
Please make a PR

PR #567, opened 2018-01-24
Followup to PR #565: Address new conversion warning in TAO_IDL/fe/idl.yy.cpp

okellogg:
http://buildlogs.remedy.nl/win_msvc15_opendds_debug/index.html
https://github.com/DOCGroup/ACE_TAO/pull/565#issuecomment-359923220 https://github.com/DOCGroup/ACE_TAO/pull/565#issuecomment-359933897
> fe\idl.yy.cpp(1383): warning C4267: 'argument': conversion from 'size_t' to 'int'

`TAO/TAO_IDL/tao_idl_fe.mpc` rule lex.yy
- In the sed regexes, replace the fread related substitution by an
  #include <errno.h> substitution.
  As a marginal cosmetic improvement, adjust the tab substitution
  to produce 4 spaces per tab.
- In the error message of the else-part, increase the minimum required
  flex version to 2.6.4.

`TAO/TAO_IDL/fe/idl.yy.cpp`
- Regenerate using regen.sh with flex 2.6.4.


okellogg:
Hmm.. that is ugly.
I hesitate to propose this because it muddles up the annotation syntax:
```
diff --git a/TAO/TAO_IDL/fe/idl.ll b/TAO/TAO_IDL/fe/idl.ll
index 34b42a3..a6d58c3 100644
--- a/TAO/TAO_IDL/fe/idl.ll
+++ b/TAO/TAO_IDL/fe/idl.ll
@@ -403,8 +403,11 @@ L"'"\\u([0-9a-fA-F]{1,4})"'" {
                     }
                   break;
                 }
-@[A-Za-z][A-Za-z0-9_]*[ \t]*(\([^)]+\))?
-                break;
+@ ?[A-Za-z][A-Za-z0-9_]*[ \t]*(\([^)]+\))? {
+                  /* The optional space after the @ is a workaround for the
+                     SunCC preprocessor which adds it.  */
+                  break;
+                }
 [ \t]*          break;
 {NL}            {
                   idl_global->set_lineno (idl_global->lineno () + 1);
```

mitza-oci:
No matter how ugly it is, the IDL spec makes this legal.  The "@" and <scoped_name> in <annotation_appl> are two distinct tokens.  There could be an amount of whitespace (or a comment) between them.

okellogg:
On the other hand, [7.4.15.4.1](http://fiware-middleware.github.io/KIARA/specification/Middleware_IDL_Specification.html#annotation-header) contains

`(220) <annotation_header> ::= "@annotation" <identifier>`

`<annotation_dcl>` treats `@annotation` as a single token thus the problem with SunCC resurfaces.

mitza-oci:
Since we're not implementing that part of the grammar, I'm fine with deferring that until later.  It also seems to contradict the rules in 7.2 so perhaps it will get changed in IDL 4.3?

okellogg:
I created a proposal to align `(225)` with `(220)`, it's currently in the [Inbox](http://issues.omg.org/browse/INBOX-623).

okellogg:
How shall we go forward?
I can offer to prepare a pull request for the aforementioned [patch](https://github.com/DOCGroup/ACE_TAO/pull/567#issuecomment-360604132) (?)

mitza-oci:
Yes, I think we should have that patch.  I don't think the comment about SunCC is necessary since having a space there is legal according to the spec.

okellogg:
Okay.
I would prefer that the current PR be merged before I make the second one.

mitza-oci:
That's what the AppVeyor build does on the pull request.  It completed successfully and you should be able to view the full logs.

Since these changes cause a full rebuild of everything that uses tao_idl, I'd rather have it in a single PR.  I think the chances are very low of us finding a problem that can't be easily identified as caused by the warnings/flex version change or the whitepsace change.

okellogg:
> That's what the AppVeyor build does on the pull request. [...]
Ah, I see.
Grepping for `idl.yy.cpp` in
https://ci.appveyor.com/project/jwillemsen/ace-tao/build/1.0.766
shows no warning; I guess I got lucky then.

> [...] I'd rather have it in a single PR.

Wilco - assuming that this is okay with @jwillemsen  (?)

jwillemsen:
Ok

PR #588, opened 2018-03-06
Bug 1220 fix applied to SHMIOP (fixed for IIOP long ago).

mitza-oci:
Also added a SHMIOP test, IPv6 compatibility, and fixed wrong indent in Synch_Invocation.cpp.


jwillemsen:
Looks FC6 has an issue with some of the new tests, could you have a look @mitza-oci http://scoreboard.ociweb.com/doc_master_mosquito_linux_gcc_d0o1i0c1/index.html

mitza-oci:
Should be just a matter of excluding the test for CORBA/e Compact.

PR #603, opened 2018-03-28
Fix Unicode issue on load registry data

reicesar:
Hi,

I have a issue on installing the ImRActivator on Windows 2012 Server. My settings was not read by service when it was started. Because ImR Locator worked fine, I've compared Locator_Options.cpp and Activator_Options.cpp and saw some differences at the code that read "ORBInitOptions" key data, for example. I've copied the same solution to Activator_Options.cpp and the service is working fine now.

Kind Regards,
Reinaldo Guedes


jwillemsen:
This looks more unicode changes and not 64bit changes

reicesar:
Yes, I´ve changed the title.

jwillemsen:
Thanks for your contribution!

PR #608, opened 2018-04-05
No need to list ace_wchar.inl twice in the MPC file.

mitza-oci:


mitza-oci:
Relates to Issue DOCGroup/MPC#50

jwillemsen:
@mitza-oci Can we merge this?

mitza-oci:
It's not going to make much of a difference until the MPC issue is fixed, so let's keep it open in case there are other related changes.

PR #609, opened 2018-04-08
* Fix ACE_Handle_Set::is_set to support platforms with both …

dskyle:
ACE_HAS_BIG_FD_SET and ACE_HAS_NONCONST_FD_ISSET.
* Add ACE_HAS_NONCONST_FD_ISSET to compile for recent android NDKs.

### DIFF
--- a/ACE/ace/config-android.h
+++ b/ACE/ace/config-android.h
@@ -357,6 +357,10 @@
 # define ACE_LACKS_CONDATTR 1
 #endif
 
+#if defined(__BIONIC_CAST)
+# define ACE_HAS_NONCONST_FD_ISSET
+#endif
+
 #if !defined ACE_DEFAULT_TEMP_DIR
 # define ACE_DEFAULT_TEMP_DIR "/data/tmp"
 #endif
--- a/ACE/ace/Handle_Set.inl
+++ b/ACE/ace/Handle_Set.inl
@@ -65,18 +65,16 @@ ACE_INLINE int
 ACE_Handle_Set::is_set (ACE_HANDLE handle) const
 {
   ACE_TRACE ("ACE_Handle_Set::is_set");
-#if defined (ACE_HAS_NONCONST_FD_ISSET)
+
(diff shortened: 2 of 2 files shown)


ACE/ace/config-android.h
```diff
@@ -357,6 +357,10 @@
 # define ACE_LACKS_CONDATTR 1
 #endif
 
+#if defined(__BIONIC_CAST)
```
jwillemsen: Any version number checked required?

jwillemsen:
Looks appveyor has a problem compiling the changed code on Windows, can you fix that?

dskyle:
Looks like it's fixed.

ACE/ace/config-android.h
mitza-oci: There's a reasonable argument to be made (not sure if it's 100% correct), that POSIX only guarantees FD_ISSET can work on a pointer to non-const fd_set.  Would it be easier just to make it a mutable member of ACE_Handle_Set?

dskyle: Yeah, that's a reasonable point. But I'd want to avoid making it mutable. That should be a last resort. Alternatively, we could make a local variable copy of the fd_set first, and call FD_ISSET on that. Reasonable compilers should be able to optimize the copy away, at least in release mode.

jwillemsen: Maybe we should go with the const_cast always, that would make this code a little bit easier to maintain.

dskyle: Yeah, the only risk of undefined behavior would be if someone
1. creates a `const ACE_Handle_Set`, which I wouldn't expect to be common usage, *and*
2. calls an `FD_ISSET` implementation with it which actually modifies the `fd_set` given, which doesn't make sense, and as far as I know, doesn't exist anywhere.

So I'd think always `const_cast` is fine. But if you want absolute conformance with POSIX, I think the local variable copy is the best solution.

shuston:
Looks good - thanks!

PR #614, opened 2018-04-10
Riverace case2088, additional fix.

shuston:


mitza-oci:
For those (like me) who don't know what case2088 is, see #583 for the previous change including ACE/NEWS entry.

PR #647, opened 2018-06-26
Timer wheel coverity fixes

shuston:


jwillemsen:
Any unit test extensions for the broken next-count logic?

shuston:
No – it was motivated by coverity saying it would fail. Indeed, it would have. I tried to write a test that triggered the problem and could not without running out of memory on the machine.

jwillemsen:
@shuston looks we have new warnings, can you fix those, see for example http://buildlogs.remedy.nl/rhel70_atcd_release_ndds523rev0/index.html

PR #650, opened 2018-06-29
Removed not used variable

jwillemsen:
* ACE/ace/Timer_Wheel_T.cpp:


jwillemsen:
Fixes warning introduced by #647

PR #652, opened 2018-07-09
Removed duplicated project files (MPC issue 50)

mitza-oci:


mitza-oci:
Link to https://github.com/DOCGroup/MPC/issues/50

PR #669, opened 2018-08-10
For per client activated servers store the server reference within the AAM

jwillemsen:
For a per client activate server we need to store the state within the AAM because at the moment multiple clients at the same moment invoke an operation for the same server multiple instances of the server are spawned and they share the Shared_Info state between them. Storing the started server details within the Shared_Info will result in the situation that only the first server started is reported back to the client, the other clients hang


jwillemsen:
Most initial logging enhancements are part of a separate PR which I will first merge making the real fix only the last commit

Issue #677, opened 2018-08-22
Stack around the variable was corrupted.


KOMMYHAP:
### Version
ACE-6.0.1
TAO-2.0.1

### Host machine and operating system
Windows 7

### Target machine and operating system (if different from host)
Windows 7

### Compiler name and version (including patch level)
Visual Studio 14

### The $ACE_ROOT/ace/config.h file
https://pastebin.com/jH2kAvuD

### Contents of $ACE_ROOT/bin/MakeProjectCreator/config/default.features
Not found.
Used by MPC when you generate your own makefiles

### AREA/CLASS/EXAMPLE AFFECTED:
HTIOP.dll: 
HTIOP_Connection_Handler.cpp: 
int TAO::HTIOP::Connection_Handler::open (void*)

### The problem effects:
Execution failed.

### Synopsis
Array index is out of bounds in attempt to copy wide string to buffer of wide char.

### Description
Look at line 123.
Local variable 'remote_addr' contains htid and tries to copy htid to wide buffer 'client' by method 'addr_to_string'. Something happened wrong into its implementation and at the end of function 'open'
VS throws "Stack around the variable 'client' was corrupted."

### Sample fix/ workaround
Note that second parameter of 'addr_to_string' is number of bytes (not characters).
But if you look into implementation this parameter forwarded to 'ACE_OS::strncpy' like number of characters.
Also we must consider htid_.length() returns number of character, so code below is not correct.
// int ACE::HTBP::Addr::addr_to_string (ACE_TCHAR buffer[], size_t size, int ipaddr_format) const
if (size < htid_.length()) { /*...*/ }
I propose the following code:
https://pastebin.com/k0hJUMv5

jwillemsen:
This version is ancient, can you retest with TAO 2.5.1 and propose a PR based on master?

KOMMYHAP:
Sorry, but I cannot retest with newer version.
Cursory analysis of above-described files on master branch shows bug also exists into ones and this bug does not depend on TAO version.

PR #678, opened 2018-08-22
Proposal for issue #677

KOMMYHAP:
Adapted for ACE_TCHAR


jwillemsen:
Seems you are putting an older version of the file back, there is now an old Id tag back, please make sure the diff only shows the real fix, nothing else

PR #682, opened 2018-08-30
Use RAND_poll instead of RAND_screen that is deprecated in OpenSSL 1.1

likema:
when compiling ACE by MinGW, it reports

``` 
SSL_Context.cpp:175:22: warning: ‘void RAND_screen()’ is deprecated (declared at openssl/rand.h:66) [-Wdeprecated-declarations]
```
See also: https://www.openssl.org/docs/man1.1.0/crypto/RAND_screen.html


jwillemsen:
Thanks!

PR #683, opened 2018-08-30
SSLIOP ECDH Addition

lewisdo:
Modified the SSLIOP::Protocol_Factory::init to process an "-SSLEcName" argument to set the ECDH curve name.


TAO/orbsvcs/orbsvcs/SSLIOP/SSLIOP_Factory.cpp
```diff
@@ -547,6 +549,15 @@ TAO::SSLIOP::Protocol_Factory::init (int argc, ACE_TCHAR* argv[])
         {
           this->check_host_ = true;
         }
+      else if (ACE_OS::strcasecmp(argv[curarg],
+                                  ACE_TEXT("-SSLEcName")) == 0)
+      {
+          curarg++;
+          if (curarg < argc)
+          {
+              ec_name = static_cast<const char *>(ACE_TEXT_ALWAYS_CHAR(argv[curarg]));
```
jwillemsen: This cast looks suspicious, it is not used for any other ssl argument

mitza-oci: It is required to use the copying assignment instead of the owning assignment of String_var.

jwillemsen: In other places a string_dup is used, why not here, that makes very clear a copy is made

mitza-oci: Using `const char*` we can interact with String_var/String_Manager and all the normal C++ behaviors for copy/assign/destruct remain in effect (without explicit memory management).  Unfortunately the same isn't true for `char*`.  It would be helpful if `init()` functions that shouldn't be changing the data used a type like `const char* argv[]` so that const could be maintained throughout, but when interfacing with older APIs we may need to add it explicitly.

jwillemsen:
@mitza-oci @lewisdo Seems we have some issues on older RHEL versions, can you have a look, see http://tao.microfocus.com/scoreboard/rhel5_x86_inline_gcc44/index.html

```
[Details] /home/oftao/doc_group/rhel5_x86_inline_gcc44/ATCD/TAO/orbsvcs/orbsvcs/SSLIOP/SSLIOP_Factory.cpp: In member function ‘virtual int TAO::SSLIOP::Protocol_Factory::init(int, ACE_TCHAR**)’:
[Details] /home/oftao/doc_group/rhel5_x86_inline_gcc44/ATCD/TAO/orbsvcs/orbsvcs/SSLIOP/SSLIOP_Factory.cpp:757: error: ‘EC_KEY’ was not declared in this scope
[Details] /home/oftao/doc_group/rhel5_x86_inline_gcc44/ATCD/TAO/orbsvcs/orbsvcs/SSLIOP/SSLIOP_Factory.cpp:757: error: ‘ecdh’ was not declared in this scope
[Details] /home/oftao/doc_group/rhel5_x86_inline_gcc44/ATCD/TAO/orbsvcs/orbsvcs/SSLIOP/SSLIOP_Factory.cpp:757: error: ‘EC_KEY_new_by_curve_name’ was not declared in this scope

```

lewisdo:
I have looked into the compile errors and I believe I know what the issue
is, working on the best way to correct the issue.

Doug

mitza-oci:
@simonmcqueen please add `<command name="print_openssl_version"/>` to the build config

mitza-oci:
Red Hat's notes: https://access.redhat.com/articles/1462353
In our testing OpenSSL 0.9.8e from openssl.org works, but perhaps Red Hat introduces breaking changes in their package.

mitza-oci:
> 
> 
> > @simonmcqueen please add `<command name="print_openssl_version"/>` to the build config
> 
> @mitza-oci This might be setting a world record for slowest reply - I'm very sorry - my public github notifications were not watched as I wasn't actively committing anywhere. I just found this now by sheer accident. Very much my bad.

Not a problem.  Since the change was merged about 3 years ago and nobody complained we're probably OK.


> I just tried but don't have permissions to the autobuild repo so I've forked it and created a pull request.

You should now have access, or at least an invitation to access.

PR #697, opened 2018-09-07
Qt5 support

mitza-oci:


jwillemsen:
Can you list this addition also in the NEWS file?

ACE/bin/MakeProjectCreator/config/ace_qtreactor.mpb
```diff
@@ -2,3 +2,6 @@ feature(qt4): ace_qt4reactor {
 }
 feature(qt5): ace_qt5reactor {
 }
+feature(!qt4, !qt4) {
```
jwillemsen: Shouldn't the second be qt5?

mitza-oci:
I'm seeing that on our builds too.  We'll look into it.  Maybe the dummy_label part will have to define a no-op MOC_Files custom type.  I wasn't sure if that was resolved before or after checking `requires`.
We may need to remove `qt4=1` from `gnuace.features` since this isn't going to work well with the feature base project.

mitza-oci:
See #700, I left `gnuace.features` alone for backwards-compatibility but users may need to set qt4=0 explicitly in default.features or on the command line.

jwillemsen:
Maybe just delete the setting from gnuace.features, let the user explicitly enable it

PR #698, opened 2018-09-08
Removed qt settings to 0 for debian packaging, these are disabled by default

jwillemsen:
* ACE/debian/default.features:
    * ACE/debian/platform_macros.GNU:


jwillemsen:
#697

PR #703, opened 2018-09-12
Fix AIX and Solaris linking rpath errors.

likema:
In AIX and Solaris, linking rpath is `-Wl,-R` rather than `-Wl,--rpath`.

### DIFF
--- a/ACE/include/makeinclude/platform_aix_g++.GNU
+++ b/ACE/include/makeinclude/platform_aix_g++.GNU
@@ -83,6 +83,7 @@ endif
 LIBS           += -lxti -ldl
 ARFLAGS        += cruv
 RANLIB          = ranlib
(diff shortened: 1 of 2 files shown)


jwillemsen:
Would prefer to keep gnu.mpd as os independent and put platform differences in the specific platform files, the ?= only sets a value when it has not been set yet, can't you use the approach as HPUX used?

likema:
I am not sure whether only AIX and Solaris GCC use -Wl,-R

jwillemsen:
I don't know how it should work for other compilers. Would recommend you only set the necessary flag for AIX/Solaris for the compilers you know they should have the flag, not try to guess what other compilers should have

jwillemsen:
Thanks

PR #575, opened 2018-02-18
Address #571 TAO_IDL annotations: lexing/parsing separation of responsibilities

okellogg:
Handling of annotations is moved from lexer to parser. No actions are implemented yet.

PR #723, opened 2018-10-12
Annotation Support in the tao_idl Frontend

iguessthislldo:
Added Support for Annotations beyond what was added in #565, #567, and #575:

- Annotations are now a type of node in the AST and can now be dumped with the rest of the AST and read by a backend:
  - Added syntax and AST support for annotation declarations, which can be used by the backend to define builtin annotations.
  - AST Support for Annotations on:
    - Modules
    - Constants
    - Structs and Struct Members
    - Unions, Union Discriminators, and Union Members
    - Enums and Enumerators
    - "`typedef`ed" Types
    - Array and Sequence Base Types
  - Any other use of annotations will either cause a syntax error or cause a warning message saying that annotations on a element are not supported.
  - `tao_idl` will error if encountering an annotation in the default IDL 3 mode
  - Support for passing Enums and Constants defined in the annotation declaration as annotation parameters.
- Attempts to Improve various aspects of Bison Generation
- Bison wrapper script is now a perl script via @mitza-oci 
- Additional `tao_idl` command line arguments:
  - `-h` and `--help` as aliases for `-u`
  - `--version` as an alias for `-V`
  - `--syntax-only` exits after parsing the IDL files
  - `--bison-trace` enables Bison state tracing during parsing
  - `--idl-version` sets the version of IDL to use
  - `--default-idl-version` prints the default version of IDL to use
  - `--list-idl-versions` lists the supported versions of IDL
  - `--dump` as an alias for `-d` and now implies`--syntax-only`
  - `--dump-builtins` implies `--dump` and also dumps IDL defined using `idl_global->eval()`
  - `--just-dump-builtins` just dumpes IDL defined using `idl_global->eval()` and exits before parsing any files.
  - `--unknown-annotations` controls the reaction when encountering undefined annotations. The valid arguments to this option are `warn-once` (The Default), `warn-all`, `error`, and `ignore`.
- `tao_idl` will now return an error status when encountering an invalid argument. Other front ends will need to support this behavior explicitly or they will have a mixture of the current and error status behavior.
- Improvements to AST dump
- `idl_global->eval(const char*)` was created to allow a Backend to define builtin annotations by inserting IDL into the AST as though an IDL file passed to the compiler. It could be used for defining other builtin complex types though.
- New warning and error types for invalid IDL by version and unsupported IDL features.
- `ACE_TAO/TAO/tests/IDLv4`: Tests for IDL version 4 Features
  - `annotations`: Instance of `tao_idl` that tests the parsing of annotations just using the internal API.

# TODO

* [x]  `ACE_for_TAO on Traivs: libTAO_IDL_FE.so: undefined reference to ACE_DLList_Node::ACE_DLList_Node`
* [x] Fuzz: TAO/TAO_IDL/util/idl_version.cpp:53: found trailing whitespace
* [x] Fuzz: TAO/TAO_IDL/util/idl_version.cpp:7: include ace/OS.h used
* [x] Make Use of `UTL_Error`
* [x] `typedef` Annotations
* [x] Fix `union` Discriminator Annotations
* [x] Change Annotations to be more like structs instead of just a sequence of pairs of parameter name and const values.
* [x] Fix typedef sequence type annotations.
* [x] Warn once per unknown annotation
* [x] Document how to enable and read annotations from a Back End
* [x] Fix linking error for `yy_parse_string()` on Windows.
* [x] Fix `tao_idl` errors and seg faults when building TAO.
* [x] Add options that control warnings and errors relating to annotations.
* [x] IDL compiler dedicated to testing annotations (Started, just need to finish moving my existing test IDL into builtin IDL using the `eval()` function and verify the state of the AST in each case).
* [x] Annotate Array Base Type. Ex: `typedef long TheArray @something [100000]`


mitza-oci:
Consolidated list of things that the CI checks are complaining about:
- [ ] ACE_for_TAO on Traivs: libTAO_IDL_FE.so: undefined reference to ACE_DLList_Node::ACE_DLList_Node
- [ ] Fuzz: TAO/TAO_IDL/util/idl_version.cpp:53: found trailing whitespace
- [ ] Fuzz: TAO/TAO_IDL/util/idl_version.cpp:7: include ace/OS.h used
- [ ] CodeFactor: some style issues that I'm not sure we care about (where do these rules come from)?

TAO/TAO_IDL/fe/idl.ypp
```diff
@@ -360,181 +379,194 @@ AST_Expression::ExprType t_param_const_type = AST_Expression::EV_none;
...
 definitions
-        : definitions annotations definition
-        | /* EMPTY */
+        : at_least_one_definition
+        | %empty
+        ;
+
+at_least_one_definition
+        /*
+         * This is broken up into 2 rules to avoid a grammer abiguity caused
```
mitza-oci:
```suggestion
         * This is broken up into 2 rules to avoid a grammar ambiguity introduced
```

TAO/TAO_IDL/fe/idl.ypp
```diff
@@ -360,181 +379,194 @@ AST_Expression::ExprType t_param_const_type = AST_Expression::EV_none;
...
+        : definitions at_least_one_annotation definition
+        {
+          AST_Decl *d = $3;
+          if (d)
+            {
+              d->annotation_appls ($2);
+            }
+          else
+            {
+              ACE_DEBUG ((LM_WARNING,
```
mitza-oci: Should this use UTL_Error's warning support for more uniform output?

TAO/TAO_IDL/fe/idl.ypp
```diff
@@ -2343,69 +2254,186 @@ positive_int_expr :
         }
         ;
 
-/* Annotations
-   TODO: actions
-   TODO: annotation_dcl and friends
- */
+annotation_dcl
+        : IDL_ANNOTATION_DECL IDENTIFIER '{' annotation_body '}' {
+          if (idl_global->idl_version_ >= IDL_VERSION_4)
```
mitza-oci: Similar to my note above, should this delegate error/warning output to UTL_Error?  (I'll stop making the same comment now...)

TAO/TAO_IDL/fe/idl.ypp
```diff
@@ -2343,69 +2254,186 @@ positive_int_expr :
...
+        | const_dcl
+        | type_dcl
+        ;
+
+annotation_member_dcl
+        : annotation_member_type simple_declarator
+        ;
+
+annotation_member_type
+        : const_type /* Also covers scope_name wanted by the spec grammer */
```
mitza-oci:
```suggestion
        : const_type /* Also covers scoped_name wanted by the spec grammar */
```

TAO/TAO_IDL/include/ast_decl.h
```diff
@@ -169,6 +173,13 @@ class TAO_IDL_FE_Export AST_Decl : public virtual COMMON_Base
       , NT_mirror_port              // Denotes a mirror port
       , NT_connector                // Denotes a CCM connector
       , NT_param_holder             // Denotes a template param placeholder
+      , NT_annotation_decl          // The declaration of an annotation
+      , NT_annotation_appl          // Application of an annotation to an IDL element
+      /**
```
mitza-oci: This could use rephrasing (and format the comment like the other enumerators), but I'm not sure exactly how to rephrase it.

TAO/TAO_IDL/include/utl_idlist.h
```diff
@@ -113,6 +113,17 @@ class TAO_IDL_FE_Export UTL_IdList : public UTL_List
 
   int compare (UTL_IdList *other);
   // Compares each component for equality.
+
+  /**
+   * Return a copy of the contatins as a string
```
mitza-oci:
```suggestion
   * Return a copy of the contents as a string
```

mitza-oci:
@okellogg does this branch work with the way you're using annotations?

okellogg:
> @okellogg does this branch work with the way you're using annotations?

@iguessthislldo wrote:
> AST Support for Annotations on:
>
>    Modules
>    Constants
>    Structs and Struct Members
>    Unions, Union Discriminators, and Union Members
>    Enums and Enumerators

We mainly used the \@range annotation on Typedefs to generate Ada range constrained types and to mimick this feature in other languages (C++, Java).

Example:
`@range(min = 1, max = 10)  typedef short small_positive;`
Generated Ada:
`type small_positive is new Interfaces.Integer_16 range 1 .. 10;`

Ada will insert range checking to ensure that variables of type small_positive are within the defined range; an exception is raised on range violation.
We emulated this feature in C++ and Java by generating the range check in the body of setters (for Java, we used a valuetype-like mapping of IDL `struct` where members are private and only accessible via setters and getters).

iguessthislldo:
Annotating typedefs is still supported gramatically (It will just say they are not supported). I will just have to figure out how far away the actual object that represents the custom type is from the point of annotation in the code and close that distance.

jwillemsen:
> > * [ ]   CodeFactor: some style issues that I'm not sure we care about (where do these rules come from)?
> 
> I would agree with that sentiment.

CodeFactor does a static code check and has reported some interesting issues in the past. I see the empty lines just as suggestion, could be removed, or just left there if it clarified the code

iguessthislldo:
@jwillemsen Windows on Azure can't download strawberry perl, it appears the site is down.

jwillemsen:
Strawberry perl is back again, just push a small change to trigger the CI integrations

TAO/TAO_IDL/ast/ast_annotation_appl.cpp
```diff
@@ -0,0 +1,182 @@
...
+          if (!iter.done ())
+            {
+              dump_i (o, ", ");
+            }
+        }
+      dump_i (o, ")");
+    }
+}
+
+int AST_Annotation_Appl::ast_accept (ast_visitor *visitor)
```
jwillemsen: Just comment visitor out as argument, the ACE_UNUSED_ARG can be removed when you do that

TAO/TAO_IDL/ast/ast_annotation_appl.cpp
```diff
@@ -0,0 +1,182 @@
...
+AST_Annotation_Appl::Param *
+AST_Annotation_Appl::find_param (const char *name)
+{
+  Params *parameters = params ();
+  if (parameters)
+    {
+      // Check for single nameless parameter
+      if (parameters->size () == 1)
+        {
+          Param *top;
```
jwillemsen: Initialise with 0

TAO/TAO_IDL/ast/ast_annotation_appl.cpp
```diff
@@ -0,0 +1,182 @@
...
+          parameters->top (top);
+          if (top && !top->id && top->expr)
+            {
+              // Don't reuse it if used
+              return top->used ? 0 : top;
+            }
+        }
+      Param **param;
```
jwillemsen: Initialise 0

TAO/TAO_IDL/ast/ast_annotation_member.cpp
```diff
@@ -0,0 +1,112 @@
...
+    AST_Field (name, dynamic_cast<AST_Field*>(other)),
+    expr_type_ (other->expr_type ()),
+    value_ (other->value () ?
+      new AST_Expression (other->value (), other->expr_type ()) : 0)
+{
+}
+
+AST_Annotation_Member::~AST_Annotation_Member ()
+{
+  if (value_)
```
jwillemsen: No need to check before calling delete

TAO/TAO_IDL/include/ast_expression.h
```diff
@@ -335,12 +344,14 @@ class TAO_IDL_FE_Export AST_Expression
   AST_Param_Holder *param_holder_;
   // Non-zero if we were created from a reference template param.
 
-private:
```
jwillemsen: Why remove several privates in multiple headers? Should all be public?

TAO/TAO_IDL/include/ast_structure.h
```diff
@@ -163,16 +173,13 @@ class TAO_IDL_FE_Export AST_Structure
...
-  virtual AST_Field *fe_add_field (AST_Field *f);
-
   virtual AST_Enum *fe_add_enum (AST_Enum *e);
 
   virtual AST_EnumVal *fe_add_enum_val (AST_EnumVal *v);
 
   virtual int compute_size_type (void);
   // Compute the size type if it is unknown.
 
-protected:
```
jwillemsen: Should this be removed?

iguessthislldo: This one was redundant

TAO/TAO_IDL/include/ast_expression.h
iguessthislldo: I made a few things public or protected because I needed them, but the most should have been because they were redundant. This one might have been a mistake though, I'll have to check.

iguessthislldo: Nevermind, I wasn't reading Github right, this is redundant as well.

Issue #676, opened 2018-08-21
Support Clang on Android


pepijndevos:
Sorry for deleting all the standard questions. It's pretty simple: the Android NDK deprecates GCC, and using it gives various deprecation warnings and later compilation errors due to a broken mix of deprecated GCC stuff and new Clang stuff. These issues are marked as wontfix upstream, and GCC will be completely removed form the next release.

This library only supports GCC on Android, so Clang support is needed for compatibility with newer NDKs.

See https://github.com/DOCGroup/ACE_TAO/blob/3f3878a5639c4954bad48b036ca68cc57eb0cac2/ACE/include/makeinclude/platform_android.GNU#L235

mitza-oci:
@iguessthislldo please update this issue when you have a change ready

jwillemsen:
Support for clang on Android has been added by PR #726

PR #726, opened 2018-10-18
Support for Clang based Android NDKs

iguessthislldo:
This PR addresses #676 by allowing ACE and TAO to compile against Android NDK r18+, which uses Clang instead of GCC. This is still a work in progress as support for older NDKs is spotty. In addition these changes intentionally break builds using NDKs older than r16 because these are missing  `ndk_version.h` which is needed to determine what macros for ACE are needed. This can to fixed by supplying the correct values `__NDK_MAJOR__` and `__NDK_MINOR__` in `platform_macros.GNU`. In exchange for this, assuming `config-android.h` is written correctly, ACE should be backwards compatible with NDKs from now on.

### DIFF
--- a/ACE/ace/os_include/sys/os_sem.h
+++ b/ACE/ace/os_include/sys/os_sem.h
@@ -27,6 +27,10 @@
 
 #if !defined (ACE_LACKS_SYS_SEM_H)
 #  include /**/ <sys/sem.h>
+#else
+#  ifdef ACE_HAS_LINUX_SEM_H
+#    include <linux/sem.h>
+#  endif
 #endif /* !ACE_LACKS_SYS_SEM_H */
 
 // Place all additions (especially function declarations) within extern "C" {}
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,8 @@ ipch/
 /setenv.cmd
 /setenv.sh
 /build
+/ACE/build
+/TAO/build
 
 *.o
 *.res
--- a/ACE/ACE-INSTALL.html
+++ b/ACE/ACE-INSTALL.html
@@ -2,9 +2,33 @@
 <html><head><!--  -->
   <meta content="text/html; charset=ISO-8859-1" http-equiv="content-type"><title>Building and Installing ACE and Its Auxiliary Libraries and Services</title>
 
-  <link rev="made" href="mailto:d.schmidt@vanderbilt.edu"></head>
+  <link rev="made" href="mailto:d.schmidt@vanderbilt.edu">
 
-<body bgcolor="#ffffff" link="#000fff" text="#000000" vlink="#ff0f0f">
+<style>
+body {
+  background-color: #ffffff;
+  color: #000000;
+}
+table, th, td {
+  border: 1px solid black;
+}
+.indent {
+  margin-left: 2em;
+}
+.boxed {
+  border: 1px solid black;
+  margin: .5em;
+  padding: .5em;
+}
+pre, code {
+  background-color: #e3e3e3;
+  padding: .1em;
+  margin: .1em;
+}
+</style>
+</head>
+
+<body>
 
 <hr>
 <h1>Building and Installing ACE and Its Auxiliary Libraries and Services</h1>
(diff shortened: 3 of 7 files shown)


iguessthislldo:
Support for NDK r17 seems to have been fixed by forcing clang by default. This also doesn't break older versions (At least 12 onwards) like I feared, so I'm going to keep that change. I've set up my travis build tonight to check 26 combinations of NDK and API versions for build issues. If that goes alright and I finish editing `ACE-INSTALL.html` I would say I don't have anything major to add this PR.

The only major issue I can't address is the name collision of ACE's `semun` and `linux/sem.h`'s `semun` when the NDK is younger than 15 with API set to 21 or greater. For now I just use `#error` to say that's not supported.

In case anyone's interested, these are the combinations I'm checking.
```
  ("r18b", {"arm": ["16", "21", "22", "23", "24", "25", "26", "27", "28",], "arm64": ["28",], "x86_64": ["28",], }),
  ("r17c", {"arm": ["16", "21", "24",                   "26", "27", "28",],                                      }),
  ("r16b", {"arm": ["16", "21", "24",                                    ],                                      }),
  ("r15c", {"arm": ["16", "21", "24",                                    ],                                      }),
  ("r14b", {"arm": ["16",                                                ],                                      }),
  ("r13b", {"arm": ["16",                                                ],                                      }),
  ("r12b", {"arm": ["16",                                                ],                                      }),
```

jwillemsen:
Can you also share your travis configuration to test with the NDK? Maybe extend the configuration for ACE with one NDK release (latest maybe) and add some comments in the travis file how to test various combinations?

iguessthislldo:
These are the updated combinations I've been checking:

```
  ("r18b", {"arm": ["16", "17", "18", "19",      "21", "22", "23", "24",       "26", "27", "28",], "arm64": ["28",], "x86_64": ["28",], }),
  ("r17c", {"arm": ["16",                        "21",             "24",       "26", "27", "28",],                                      }),
  ("r16b", {"arm": ["16",                        "21",             "24",                        ],                                      }),
  ("r15c", {"arm": ["16",                        "21",             "24",                        ],                                      }),
  ("r14b", {"arm": ["16",                                                                       ],                                      }),
  ("r13b", {"arm": ["16",                                                                       ],                                      }),
  ("r12b", {"arm": ["16",                                                                       ],                                      }),
```
I just fixed something for r15 so I have to wait for them to be finished to be absolutely sure, but I'm pretty sure that ACE can be built across all of these combinations of Android NDK revision, API levels, and CPUs. I should also clarify that all I have been doing is building everything necessary for OpenDDS on Android.

`ACE-INSTALL.html` is updated, this PR is ready to reviewed.

jwillemsen:
It looks clang is now always used, also with NDK < 18+, is that intentional or can we keep the default for gcc with NDK < 18+?

ACE/ace/os_include/sys/os_sem.h
```diff
@@ -27,6 +27,10 @@
 
 #if !defined (ACE_LACKS_SYS_SEM_H)
 #  include /**/ <sys/sem.h>
+#else
+#  ifdef ACE_HAS_LINUX_SEM_H
```
jwillemsen: Looks ACE_HAS_LINUX_SEM_H is not set in any config file, is this addition required?

iguessthislldo: No, it was leftover from an attempt to remedy to the problem with `semun` and it can be removed.

jwillemsen: Ok, removed with #745

iguessthislldo:
> It looks clang is now always used, also with NDK < 18+, is that intentional or can we keep the default for gcc with NDK < 18+?

It was a relatively last minute change, but yes, it's intentional. Without `android_force_clang`, it seems to work everywhere by just using g++, except for r17, where something changed about the standard library and that causes linking errors with things like `stderr` when using g++. I didn't see a good way to get the version number in make so that it uses g++ for <17 and clang >=17, so I decided the default should favor the newer NDKs since these changes break the older ones that don't have clang would be affected anyway over the NDK version.

iguessthislldo:
There is only so much ease of use that can be built into makefiles verses a more general scripting language. However 13b should use clang just fine. In my travis builds I tested 12 and 13 with clang and I'm pretty sure it would work with even older NDKs as well, although I haven't tested them to be sure.

iguessthislldo:
> In my travis builds I tested 12 and 13 with clang

This was incorrect, looking back at [the configure script I used](https://github.com/iguessthislldo/OpenDDS-Android/blob/55d0aeb7b4d77b6d2fd2c8059b2c4a39e6cc2f18/configure.sh#L22) I'm letting it use `g++` if NDK is 14 or before.

PR #741, opened 2018-11-07
Only define ACE_HAS_SEMUN when we have API >= 12 and at least NDK 15

jwillemsen:
Giving an error is not user friendly, we just not set ACE_HAS_SEMUN when it is not usable

This is a change related to PR #726

PR #744, opened 2018-11-07
Android API or NDK>=15 do have a senum so don't let ACE define its ve…

jwillemsen:
…rsion

    * ACE/ace/config-android.h:


jwillemsen:
#726 #741

PR #777, opened 2018-12-21
Fix end() on ACE_Vector

iguessthislldo:
It was based on the the base array, not the actual elements in the array.

This is an example program that would fail:
```C++
#include <ace/Vector_T.h>
typedef ACE_Vector<int> Vector;
int main() {
  Vector v;
  for (Vector::iterator i = v.begin(); i != v.end(); ++i) {
    return 1;
  }
  return 0;
}
```

### DIFF
--- a/ACE/ace/Vector_T.h
+++ b/ACE/ace/Vector_T.h
@@ -188,6 +188,16 @@ public:
 
   void swap (ACE_Vector &rhs);
 
+  /*
+   * Implement our own end functions because Array_Base's end functions use the
+   * current capacity, not the Vector's actual element count!
+   */
+  /// C++ Standard End Iterator
+  ///{
+  typename ACE_Array_Base<T>::iterator end ();
+  typename ACE_Array_Base<T>::const_iterator end () const;
+  ///}
+
 protected:
 
   /**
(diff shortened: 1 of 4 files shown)


iguessthislldo:
@jwillemsen I tried to run the whole test suite, but it fails on Bug_2610_Regression_Test and hangs on Bug_2772_Regression_Test, long before getting to the vector test. I'm running `ACE/tests/run_test.pl`, is this different from the ace tests normally ran?

iguessthislldo:
So after a several minutes (371 seconds to be precise) Bug_2772_Regression_Test timed out.

jwillemsen:
Some tests are known to fail, they are tagged with !FIXED_BUGS_ONLY in run_test.lst, Bug_2610_Regression and Bug_2772_Regression are both known failures.

jwillemsen:
Are your changes complete for this PR?

iguessthislldo:
Compatibility with the standard library probably needs to be improved, but I don't have anything else to add to this PR.

jwillemsen:
I would prefer to use more std and deprecate all ACE containers but that is a huge effort, in new code we can use more std step by step

jwillemsen:
Thanks!

iguessthislldo:
> I would prefer to use more std and deprecate all ACE containers but that is a huge effort, in new code we can use more std step by step

I don't think I fully processed that at the time, but for future reference does that mean I can directly use the standard C++03 library in TAO_IDL? Because there was none I was under the impression that I couldn't use it.

jwillemsen:
We are already using several std containers in ACE, the ones we use there should be safe to use in TAO_IDL

PR #781, opened 2019-01-07
IDL4 fixes

mitza-oci:


TAO/tests/IDLv4/annotations/be_init.cpp
```diff
@@ -522,7 +522,7 @@ BE_post_init (char *[], long)
       "};\n"
     ).assert_annotation_decl ("::@test_annotation_1");
     t.assert_annotation_member_count (test_annotation_1, 0);
-  } catch (Failed &f) {}
+  } catch (Failed &) {}
```
jwillemsen: Should catch by const&, else some compilers will give a warning

PR #786, opened 2019-01-09
NEWS for TAO_IDL changes

mitza-oci:


okellogg:
@iguessthislldo, @mitza-oci  Thanks for the awesome work on annotations.

PR #787, opened 2019-01-10
tao_idl: Fixes for Coverity caused by #723

iguessthislldo:


iguessthislldo:
I think fixed everything caused by #723 except for the resource leaks.

Issue #784, opened 2019-01-08
TAO_IDL: valuetype state member parsing


mitza-oci:
The TAO_IDL parser changes in #723 accidentally broke parsing valuetype members that have `public|private` before the type.  This shows up as a failure in CIAO's IDL3_to_XMI but may break TAO valuetypes as well.

We have discussed the design for a fix, which we expect to have ready soon.

PR #788, opened 2019-01-10
Fix for invalid valuetype member state visibility

iguessthislldo:
Fix for #784 revealed after #723
In tao_idl, do not pass member visibility via the Bison stack indirectly.


jwillemsen:
Can you extend one of our tests (for example TAO/tests/IDL_Test) with an example that failed before this fix? Would be good to have such a test at the TAO level

TAO/TAO_IDL/include/ast_field.h
```diff
@@ -113,7 +113,12 @@ class TAO_IDL_FE_Export AST_Field : public virtual AST_Decl
 
   // Data Accessors.
   AST_Type *field_type (void) const;
-  Visibility visibility (void) const;
+
+  /// Get and Set Visibility
+  ///{
+  Visibility visibility () const;
```
mitza-oci: Where is this method defined?

iguessthislldo: [Here](https://github.com/DOCGroup/ACE_TAO/blob/3507f3c21ac55d6ffe1806a1b46fcbc7113bba55/TAO/TAO_IDL/ast/ast_field.cpp#L219), it's not new, I just moved the declaration to be with its counterpart.

iguessthislldo:
> Can you extend one of our tests (for example TAO/tests/IDL_Test) with an example that failed before this fix? Would be good to have such a test at the TAO level

I'm not sure it's possible, unless I'm looking in the wrong place, the failures were in normal looking structs.

PR #791, opened 2019-01-14
Support IDL output in places other than the IDL file location

iguessthislldo:
Using `output_follows_input = 0`, set IDL output location to be the current directory when using `IDL_Files` in MPC.


mitza-oci:
https://github.com/objectcomputing/OpenDDS/pull/1036 will fix that issue

Issue #792, opened 2019-01-16
getconf failure can lead to strange compilation failure


KrisThielemans:
### Version

current master
### Host machine and operating system

linux

### The $ACE_ROOT/include/makeinclude/platform_macros.GNU file

`platform_linux_common.GNU`

### AREA/CLASS/EXAMPLE AFFECTED:

building

### The problem effects:

compilation failure

### Synopsis

`getconf` error (when asking for Large File Support) will lead to an incomprehensible error message

### Description
We had a strange error building ACE on OSX
```
c++ -fvisibility=hidden -fvisibility-inlines-hidden -Wnon-virtual-dtor -O3 -ggdb -pthread -fno-strict-aliasing -Wall -W -Wpointer-arith -pipe -DACE_LACKS_LINUX_NPTL Unknown -D_GNU_SOURCE   -I/Users/rich/Documents/Code/SIRF-Superbuild/Build/builds/ACE/build/source/ -D__ACE_INLINE__ -I.. -DACE_BUILD_DLL  -c -fPIC -o .shobj/ACE.o /Users/rich/Documents/Code/SIRF-SuperBuild/Build/builds/ACE/build/source/ace/ACE.cpp
clang: error: no such file or directory: 'Unknown'
```
It seems that this error was caused by using the wrong platform file (more detail at https://github.com/CCPPETMR/SIRF-SuperBuild/issues/186). However, looking at the [code](https://github.com/DOCGroup/ACE_TAO/blob/95c147c08eaad4f651347ccf898c88b3a5fca705/ACE/include/makeinclude/platform_linux_common.GNU#L110) that causes this:
```make
PLATFORM_LARGEFILE_CFLAGS := $(shell getconf LFS_CFLAGS 2> /dev/null || echo Unknown)
ifdef PLATFORM_LARGEFILE_CFLAGS
  # Large file support
  CPPFLAGS += $(PLATFORM_LARGEFILE_CFLAGS)
endif #largefile
```
it appears that if `getconf` fails for some reason, this error will appear. There's 2 possible solutions I guess:
- let `make` fail with an understandable error
- proceed assuming that nothing needs to be added to `CPPFLAGS`

This is maybe a corner case that is not worth your time of course, but I thought I'd file it.

mitza-oci:
This seems to contradict itself:
> We had a strange error building ACE on OSX

vs.

> Host machine and operating system: linux
> platform_linux_common.GNU

jwillemsen:
[...]

PR #794, opened 2019-01-17
When getconf fails don't set the value to unknown, that gives a stran…

jwillemsen:
…ge error to our users, issue #792

    * ACE/include/makeinclude/platform_linux_common.GNU:


jwillemsen:
Proposed fix for issue #792

mitza-oci:
Somewhat off-topic but it seems like AppVeyor always fails due to the 1-hour limit.  Is that check redundant now that Azure Pipelines is working?

jwillemsen:
> Somewhat off-topic but it seems like AppVeyor always fails due to the 1-hour limit. Is that check redundant now that Azure Pipelines is working?

Let us see whether we can bring AppVeyor within an hour, maybe just let it compile ACE as quick test, if that doesn't work we can remove it

PR #796, opened 2019-01-20
Fixes for ACE Tests on Android

iguessthislldo:
I finally got the ACE tests compiling on [iguessthislldo/OpenDDS-Android](https://github.com/iguessthislldo/OpenDDS-Android) but [they don't build apparently](https://travis-ci.org/iguessthislldo/OpenDDS-Android/builds/481900404). These are fixes for that.

Also NDK r19 got released so I will add that to the matrix and see how that goes.


iguessthislldo:
So I ran [a travis build](https://travis-ci.org/iguessthislldo/OpenDDS-Android/builds/481958053) with these changes and so far everything is alright as far as these changes go, but NDK r19 is failing on Compiler_Features_32_Test:
```
x86_64-linux-android-clang++ -fdelayed-template-parsing -ggdb -W -Wall -Wpointer-arith -W -Wall -pipe -Wno-deprecated -Wno-deprecated-declarations -D_REENTRANT  -D_GNU_SOURCE -DACE_HAS_CUSTOM_EXPORT_MACROS=0 -DACE_ANDROID_NDK_HAS_NDK_VERSION_H   -I/home/travis/build/iguessthislldo/OpenDDS-Android/ACE_TAO/ACE/build/target -DACE_NO_INLINE -I..  -c -o .obj/Compiler_Features_32_Test.o /home/travis/build/iguessthislldo/OpenDDS-Android/ACE_TAO/ACE/build/target/tests/Compiler_Features_32_Test.cpp
/home/travis/build/iguessthislldo/OpenDDS-Android/ACE_TAO/ACE/build/target/tests/Compiler_Features_32_Test.cpp:55:41: error: expected the class name after '~' to name a destructor
  this->u_.string_member_.std::string::~string ();
                                        ^
/home/travis/build/iguessthislldo/OpenDDS-Android/ACE_TAO/ACE/build/target/tests/Compiler_Features_32_Test.cpp:67:24: error: expected the class name after '~' to name a destructor
    u_.m.std::string::~string ();
                       ^
2 errors generated.
```

The nature of the test is that Clang had a problem with this before, so I'm guessing that the problem or a similar one has appeared in the version of clang being used in r19.

jwillemsen:
> The nature of the test is that Clang had a problem with this before, so I'm guessing that the problem or a similar one has appeared in the version of clang being used in r19.

This is a known bug in the clang compiler which has not been fixed yet in clang. I would expect that CLANG_WORKAROUND gets defined on Android with clang given that they are using clang 7. Any idea why the workaround is not used?

jwillemsen:
Thanks for the fixes!

iguessthislldo:
> This is a known bug in the clang compiler which has not been fixed yet in clang. I would expect that CLANG_WORKAROUND gets defined on Android with clang given that they are using clang 7. Any idea why the workaround is not used?

The Clang version in r19 is 8.0.2, so the workaround doesn't get applied. I'm a little confused though, because in the test source it says this was fixed:

```
 * This is fixed May 2014 in clang, see
 * http://llvm.org/viewvc/llvm-project?view=revision&revision=209319
```

If this isn't actually the case or it has regressed shouldn't we submit a bug report to clang and/or have the workaround always apply to clang?

PR #809, opened 2019-01-27
Fixed memory leak of annotation_appls_

jwillemsen:
* TAO/TAO_IDL/ast/ast_decl.cpp:


iguessthislldo:
I don't think this can't be done like this because the annotations will be shared in this situation:

```
struct sample {
  @key
  int a, b, c;
};
```

I think this was on my mental todo list but I never got around to fixing it. It would nice to be able to use `std::shared_ptr` here...

jwillemsen:
Reverted, can you make an issue and work on a good fix? Running taoidl with address sanitizer now gives reported memory leaks with each invocation

iguessthislldo:
Yes, I will do that this week.

PR #812, opened 2019-01-29
Fix ACE_Mem_Addr::get_host_addr by getting external address.

likema:
The comment of ACE_Mem_Addr::get_host_addr definition:

https://github.com/DOCGroup/ACE_TAO/blob/master/ACE/ace/MEM_Addr.h#L108

### DIFF
--- a/ACE/ace/MEM_Addr.inl
+++ b/ACE/ace/MEM_Addr.inl
@@ -4,8 +4,7 @@
 
 ACE_BEGIN_VERSIONED_NAMESPACE_DECL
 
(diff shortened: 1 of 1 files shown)


jwillemsen:
Can you extend one of our unit tests as reproducer?

likema:
I can't find any related test in ACE/tests/MEM_Stream_Test.cpp

I just fixed it by the definition comment.

jwillemsen:
Yes, as far as I can tell from the other methods the external address should be used

PR #814, opened 2019-01-29
Add thread safe ACE_MEM_Addr::get_host_addr.

likema:


likema:
ACE_INET_Addr has a  thread safe `get_host_addr`, but ACE_MEM_Addr does not has one.

jwillemsen:
Maybe you can take the docu of ACE_INET_Addr and use that also here, now the documentation in ACE_MEM_Addr doesn't tell why there are two operations with the same name but different arguments

likema:
OK

jwillemsen:
Thanks!

PR #815, opened 2019-01-29
Fix ACE_DLL_Manager error info

likema:
Fix ACE_DLL_Manager without error info when loading a shared object with undefined symbol.

### DIFF
--- a/ACE/ace/DLL_Manager.cpp
+++ b/ACE/ace/DLL_Manager.cpp
@@ -148,7 +148,7 @@ ACE_DLL_Handle::open (const ACE_TCHAR *dll_name,
               // mask it.
               // @TODO: If we've found our DLL _and_ it's
               // broken, should we continue at all?
-              if ((errno != 0) && (errno != ENOENT) && (errors || ACE::debug ()))
+              if ((errno != ENOENT) && (errors || ACE::debug ()))
                 {
                   ACE_TString errtmp;
                   if (errors)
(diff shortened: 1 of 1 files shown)


ACE/ace/DLL_Manager.cpp
```diff
@@ -241,15 +241,15 @@ ACE_DLL_Handle::open (const ACE_TCHAR *dll_name,
               if (errors || ACE::debug ())
                 {
                   ACE_TString errtmp;
-                  if (errors)
+                  this->error (errtmp);
+
+                  if (errors && errors->is_empty())
```
jwillemsen: Why an added check for is_empty here?

likema: Can these lines be removed? It looks like duplicated.

ACE/ace/DLL_Manager.cpp
```diff
@@ -241,15 +241,15 @@ ACE_DLL_Handle::open (const ACE_TCHAR *dll_name,
...
                     {
-                      errors->push (this->error (errtmp));
+                      errors->push (errtmp);
```
jwillemsen: Can the return value of the error method now be removed?

ACE/ace/DLL_Manager.cpp
jwillemsen: Can it be that error==0 but debug==1?

ACE/ace/DLL_Manager.cpp
jwillemsen: Can the return value of the error method now be removed?

ACE/ace/DLL_Manager.cpp
likema: No matter whether `errors` is 0, `errtmp` is set by `this->error` before.

But after reading it again just now, I found line from 239 to 262 can be removed because all errors are pushd or logging inside while loop before.

Am I right?

jwillemsen: Not sure, would keep them for the moment

likema: OK. Given keeping them, 

If `is_empty` is true, no error messages are **pushed** into `errors` and **logged**.

Otherwise error messages are **pushed** and **logged** so that it is not necessary to push and log it again.

PR #820, opened 2019-01-31
Fix ACE_DLL_Manager potential buffer overflow.

likema:


ACE/ace/DLL_Manager.cpp
```diff
@@ -11,7 +11,7 @@
 #include "ace/Recursive_Thread_Mutex.h"
 #include "ace/Guard_T.h"
 #include "ace/OS_NS_dlfcn.h"
-#include "ace/OS_NS_string.h"
+#include "ace/OS_NS_stdio.h"
```
jwillemsen: When this include is only for AIX, maybe guard it also with a ifdef check?

ACE/ace/DLL_Manager.cpp
```diff
@@ -177,11 +177,8 @@ ACE_DLL_Handle::open (const ACE_TCHAR *dll_name,
               if (ACE_TString::npos != name->strstr (ACE_TEXT (".a")))
                 {
                   ACE_TCHAR aix_pathname[MAXPATHLEN + 1];
-                  ACE_OS::strncpy (aix_pathname,
-                                   name->c_str (),
-                                   name->length ());
-                  aix_pathname[name->length ()] = '\0';
-                  ACE_OS::strcat (aix_pathname, ACE_TEXT ("(shr.o)"));
+                  ACE_OS::snprintf (aix_pathname, sizeof (aix_pathname),
+                                    ACE_TEXT ("%s(shr.o)"), name->c_str ());
```
jwillemsen: Shouldn't the return value of ACE_OS::snprintf be checked?

jwillemsen:
Be aware that there are no AIX builds anymore on the [scoreboard](http://www.dre.vanderbilt.edu/scoreboard/integrated.html). It can happen that AIX breaks now without notice.

ACE/ace/DLL_Manager.cpp
shuston: Good idea. I did not check the rest of the file for a need for OS_NS_stdio.h.

ACE/ace/DLL_Manager.cpp
shuston: Strictly speaking, yes. However, the code is better than it was before. Alternately the combined length could be checked before attempting ACE_OS:snprintf() because the lengths are fixed and known.

ACE/ace/DLL_Manager.cpp
```diff
@@ -177,11 +178,19 @@ ACE_DLL_Handle::open (const ACE_TCHAR *dll_name,
...
+                                        ACE_TEXT ("%s(shr.o)"),
+                                        name->c_str ()) < sizeof (aix_pathname))
+                    {
+
+                      if (ACE::debug ())
+                        ACELIB_ERROR ((LM_ERROR,
+                              ACE_TEXT ("ACE (%P|%t) DLL_Handle::open: ")
+                              ACE_TEXT ("('%s(shr.o)') is too long\n"),
+                              name->c_str()));
+
```
jwillemsen: Shouldn't we put an error text also to `errors` so that in case the caller uses `errors` he can log this error case at a higher level?

likema:
@jwillemsen It is too complex to check snprintf error:

* `< 0` denotes errno mesage. It is also complex to strerror or strerror_r  for compatibities.  https://github.com/DOCGroup/ACE_TAO/blob/master/ACE/ace/Log_Msg.cpp#L1321
* `>= sizeof(aix_pathname)` denotes path is too long.

so I use strcpy and strcat again.

Issue #811, opened 2019-01-27
Fix tao idl memory leaks related to annotations


jwillemsen:
### Version

The version of ACE and/or TAO you are using

### The $ACE_ROOT/ace/config.h file

If you use a link to a platform-specific file, simply state which one

### The $ACE_ROOT/include/makeinclude/platform_macros.GNU file

if you use a link to a platform-specific file, simply state which one (unless this isn't used in this case, e.g., with Microsoft Visual C++)

### Contents of $ACE_ROOT/bin/MakeProjectCreator/config/default.features

Used by MPC when you generate your own makefiles

### AREA/CLASS/EXAMPLE AFFECTED:

What example failed?  What module failed to compile?

### The problem effects:

Does it affect compilation, linking, or execution. Please indicate whether ACE/TAO, your application, or both are affected.

### Synopsis

Brief description of the problem

### Description

Detailed description of problem.  Don't just say "<blah> doesn't work, here's a fix," explain what your program does to get to the <blah> state.

### Repeat by

What you did to get the error; include test program or session transcript if at all possible.

### Sample fix/ workaround

If available

jwillemsen:
See 809

jwillemsen:
Also add some docu to the members that are shared

mitza-oci:
ACE_Strong_Bound_Ptr could be useful here.  Or adding another node to the AST to represent the individual "reference" (binding an Annotation Application to its target Decl).

Also, as a general note... I think last I checked tao_idl wasn't asan-clean before these changes.

jwillemsen:
We have a build with address-sanitizer=1 in platform_macros.GNU, before the TAO_IDL changes for annotations there where no reported leaks on tao_idl, right now we have the ones as mentioned in the first comment

iguessthislldo:
@jwillemsen 

It's possible that the address sanitizer said it was clean but memory management in tao_idl is a  mess in general. The usage of dynamically allocated memory is ridiculously high, which might be understandable, but on top of it there's the completely strange pattern of using a `destroy` method on everything.

@mitza-oci

I'm not sure I want to add a new ast node for this, but `ACE_Strong_Bound_Ptr` is a god send.

PR #822, opened 2019-01-31
 TAO_IDL: Fix Memory Leaks Caused By Annotations

iguessthislldo:
Fix for #811 and #809. Annotations were added in #723.

The majority of fixes were many missing destroys and deletes. The biggest change is `AST_Annotation_Appl`s are now stored in a class named `AST_Annotation_Appl` and managed using `ACE_Bounded_Strong_Ptr` because they can be shared between AST_Decls.

With these changes `valgrind --leak-check=full --show-leak-kinds=all annotest_idl` show no lost memory (There is about 1KiB still on the heap but it's not lost and not related to annotations).

There is one small change in the API as described in `annotations.md`, but it only affects what I consider to be advanced usage.


iguessthislldo:
After talking to @mitza-oci, I moved the helper function for finding annotations on nodes into the new class I created and moved that class into into its own file. The former breaks the annotation API I defined in #723, but he said that would be okay.

PR #834, opened 2019-02-12
Use Logcat on Android

iguessthislldo:
On Android `stdout` and `stderr` are discarded by default. It's possible to
workaround this but it's not pretty. The best solution is to use
Android's logging system, Logcat. I've added an `ACE_Log_Msg_Backend`,
`ACE_Log_Msg_Android_Logcat`, and set it to be the default output on
Android.


iguessthislldo:
I actually had to test this to make sure it works...

jwillemsen:
Do the ACE unit tests under tests still write a log file to disk? For our android testing we write to the console, will have to check tomorrow whether your changes don’t break that, maybe we should keep the default as is and let the user set a flag to enable this?

iguessthislldo:
In that case they don't have to be mutually exclusive, STDERR and SYSLOG can both be enabled. I just don't want the user to have to miss an error because they didn't know about this.

jwillemsen:
I checked our android test build, no special settings, so please make sure logging keeps being send to stdout/stderr by default so that when we run unit tests we get the output on the commandline. Maybe let the user enable logcat in his main or through some ACE define?

ACE/ace/Log_Msg.cpp
```diff
@@ -80,11 +82,11 @@ class ACE_Msg_Log_Cleanup: public ACE_Cleanup_Adapter<ACE_Log_Msg>
 
 #if defined (ACE_WIN32) && !defined (ACE_HAS_WINCE) && !defined (ACE_HAS_PHARLAP)
 #  define ACE_LOG_MSG_SYSLOG_BACKEND ACE_Log_Msg_NT_Event_Log
+#elif defined (ACE_ANDROID)
+#  define ACE_LOG_MSG_SYSLOG_BACKEND ACE_Log_Msg_Android_Logcat
 #elif !defined (ACE_LACKS_UNIX_SYSLOG) && !defined (ACE_HAS_WINCE)
 #  define ACE_LOG_MSG_SYSLOG_BACKEND ACE_Log_Msg_UNIX_Syslog
-#else
```
jwillemsen: Is the removal of the else part intended?

iguessthislldo: It was because it's not used. `ACE_LOG_MSG_SYSLOG_BACKEND` was only used if the conditions above were true and `SYSLOG` was enabled, other wise it created `ACE_Log_Msg_IPC` but calls it by that name, not `ACE_LOG_MSG_SYSLOG_BACKEND`. I mainly did it because it simplifies the check below and it only has to check if `ACE_LOG_MSG_SYSLOG_BACKEND` is defined.

iguessthislldo:
I will definitely enable stderr. I will also disable the logcat by default do that if you insist on it, but I don't see a reason to hide it from the user, especially since these messages aren't hidden by default if developing a normal cli ace application, so could you explain your reasoning for that?

jwillemsen:
Just make sure stderr is enabled by default. The only disadvantage for logcat would be performance, if there is a way the user can disable it I am ok with enabling it by default.

iguessthislldo:
It looks like debug level messages are ignored in release builds if that helps. I was going to see if it was possible to check if the application was release or not from Android's standpoint and disable the logging backend based on that [but it doesn't seem that's possible from the NDK alone](https://stackoverflow.com/questions/6406382/ndk-detect-if-application-is-debuggable).

ACE/ace/Log_Msg.h
```diff
@@ -161,6 +161,12 @@
 # undef THREAD
 #endif /* THREAD */
 
+#ifdef ACE_ANDROID
```
jwillemsen: Maybe add a ifndef ACE_DEFAULT_LOG_FLAGS around here so that the user can also set in his config.h file?

ACE/ace/Log_Msg.cpp
```diff
@@ -136,7 +138,13 @@ class ACE_Log_Msg_Manager
 ACE_Log_Msg_Backend *ACE_Log_Msg_Manager::log_backend_ = 0;
 ACE_Log_Msg_Backend *ACE_Log_Msg_Manager::custom_backend_ = 0;
 
-u_long ACE_Log_Msg_Manager::log_backend_flags_ = 0;
+#ifdef ACE_ANDROID
+#  define DEFAULT_LOG_BACKEND_FLAGS ACE_Log_Msg::SYSLOG
```
jwillemsen: ACE_ before and a ifndef check so that the user can set it in his config.h?

iguessthislldo: Yes, I can

ACE/ace/Log_Msg.h
iguessthislldo: Yes, I can

PR #838, opened 2019-02-19
Fix AIX dlopen debug log.

likema:


ACE/ace/DLL_Manager.cpp
```diff
@@ -203,20 +203,6 @@ ACE_DLL_Handle::open (const ACE_TCHAR *dll_name,
...
 
-                  if (ACE::debug ())
-                    {
-                      ACE_TString err;
-                      ACELIB_DEBUG ((LM_DEBUG,
-                                  ACE_TEXT ("ACE (%P|%t) DLL_Handle::open ")
-                                  ACE_TEXT ("(\"%s\", 0x%x) -> %s: %s\n"),
-                                  aix_pathname,
-                                  open_mode,
-                                  (this->handle_ != ACE_SHLIB_INVALID_HANDLE
```
likema: handle_ is always ACE_SHLIB_INVALID_HANDLE

jwillemsen: This probably should move below the ACE_OS::dlopen which was on line 220, it prints the result of the dlopen

likema: Yes, your are right.

likema: Fixed

PR #840, opened 2019-02-21
IDLv4 Explicitly-named Integer Types

iguessthislldo:
Implement part of IDL4.2 Section 7.4.13 "Building Block Extended Data-Types" by implementing:

- 7.4.13.4.4 "Integers restricted to holding 8-bits of information":  `int8` and `uint8`
  - New integer types that have to be distinguished from the existing `octet` and `char` types.
- 7.4.13.4.5 "Explicitly-named Integer Types":  `int16`, `uint16`, `int32`, `uint32`, `int64`, and `uint64`
  - Just aliases of the existing integer types


jwillemsen:
Any updates on this PR, do we need to keep this open?

iguessthislldo:
> Any updates on this PR, do we need to keep this open?

I don't see a reason to close it as I plan to come back to it someday and it also keeps the work visible to someone who might be interested. If it's an issue of time there are older PRs.

jwillemsen:
New test should be in TAO/bin/tao_orb_tests.lst, msvc doesn't compile and has some new warnings, will review when the checks are all ok

iguessthislldo:
```

**Comment by iguessthislldo** [comment]
263>CustomBuild:
         D:\a\ACE_TAO\ACE_TAO/ACE\bin\tao_idl: Unable to rename temporary file "C:\Users\RUNNER~1\AppData\Local\Temp\tao-idlf_N2p6U3" to "C:\Users\RUNNER~1\AppData\Local\Temp\tao-idlf_N2p6U3.cpp": Input/output error
   263>C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Microsoft\VC\v160\Microsoft.CppCommon.targets(241,5): error MSB8066: Custom build for 'AnySeq.pidl;Bounds.pidl;Dynamic.pidl;Dynamic_Parameter.pidl;ValueModifier.pidl;Visibility.pidl' exited with code 1. [D:\a\ACE_TAO\ACE_TAO\TAO\tao\AnyTypeCode\AnyTypeCode_Idl.vcxproj]
   263>Done Building Project "D:\a\ACE_TAO\ACE_TAO\TAO\tao\AnyTypeCode\AnyTypeCode_Idl.vcxproj" (default targets) -- FAILED.

**Comment by iguessthislldo** [comment]
> msvc doesn't compile and has some new warnings,

I can fix the warnings in `AST_Expression.cpp`, but for the error this says it's an I/O error:

```
    263>CustomBuild:
         D:\a\ACE_TAO\ACE_TAO/ACE\bin\tao_idl: Unable to rename temporary file "C:\Users\RUNNER~1\AppData\Local\Temp\tao-idlf_N2p6U3" to "C:\Users\RUNNER~1\AppData\Local\Temp\tao-idlf_N2p6U3.cpp": Input/output error
   263>C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Microsoft\VC\v160\Microsoft.CppCommon.targets(241,5): error MSB8066: Custom build for 'AnySeq.pidl;Bounds.pidl;Dynamic.pidl;Dynamic_Parameter.pidl;ValueModifier.pidl;Visibility.pidl' exited with code 1. [D:\a\ACE_TAO\ACE_TAO\TAO\tao\AnyTypeCode\AnyTypeCode_Idl.vcxproj]
   263>Done Building Project "D:\a\ACE_TAO\ACE_TAO\TAO\tao\AnyTypeCode\AnyTypeCode_Idl.vcxproj" (default targets) -- FAILED.
```

I don't know the cause of this particular error, but if there is chance for the temporary file to problematic on Windows, it would be fixed by avoiding using a temporary file in #1357.

File: ACE/ace/CDR_Base.h [hunk 1]
```diff
@@ -207,6 +207,14 @@ class ACE_Export ACE_CDR
   typedef ACE_UINT32 ULong;
   typedef ACE_INT64 LongLong;
   typedef ACE_UINT64 ULongLong;
+  typedef ACE_INT8 Int8;
+  typedef ACE_UINT8 Uint8;
```
jwillemsen, line 211: `Uint` or `UInt`, looking at the other typedefs I think we should go for `UInt`

TAO/TAO_IDL/be/be_predefined_type.cpp
```diff
@@ -222,9 +225,17 @@ be_predefined_type::compute_tc_name ()
                  Identifier (tcname));
         break;
       }
-    default:
-      ACE_ERROR ((LM_WARNING, "Unknown or invalid predefined type"));
+    case AST_PredefinedType::PT_uint8:
+      ACE_NEW (id, Identifier ("_tc_Uint8"));
```
jwillemsen, line 229: `_tc_uint8`, all others don't have a capital U

TAO/tests/IDLv4/explicit_ints/main.cpp
```diff
@@ -0,0 +1,79 @@
...
+      any_failed = true;
+    }
+}
+
+int
+ACE_TMAIN (int, ACE_TCHAR *[])
+{
+  bool any_failed = false;
+
+  expect_equals<CORBA::Uint8>(any_failed, "u8_max", u8_max, 255);
```
jwillemsen, line 37: `Uint` or `UInt`?

TAO/TAO_IDL/be/be_predefined_type.cpp
iguessthislldo: I don't really know much about typecode and the goal here just is basic support, so that depends on if that's trivial or not.

ACE/ace/CDR_Base.h
iguessthislldo: I didn't really notice the difference. I would say `Uint` if this was by itself, but yeah it should probably be `UInt` to match the others.

TAO/TAO_IDL/be/be_predefined_type.cpp
mitza-oci: Wouldn't it be better to avoid TypeCode completely until there is a spec for it?

TAO/TAO_IDL/fe/fe_lookup.cpp
```diff
@@ -176,7 +176,7 @@ TAO_IDL_CPP_Keyword_Table::lookup (const char *str, unsigned int len)
 
   if (len <= MAX_WORD_LENGTH && len >= MIN_WORD_LENGTH)
     {
-      unsigned int key = hash (str, len);
+      unsigned int const key = hash (str, len);
```
mitza-oci, line 179: this is a generated file

TAO/TAO_IDL/ast/ast_expression.cpp
```diff
@@ -1725,6 +1811,32 @@ eval_kind (AST_Expression::AST_ExprValue *ev, AST_Expression::EvalKind ek)
...
+      result = a + b;
+      break;
+    case AST_Expression::EC_minus:
+      result = a - b;
+      break;
+    case AST_Expression::EC_mul:
+      result = a * b;
+      break;
+    case AST_Expression::EC_div:
+      if (!b) return true;
```
mitza-oci, line 1830: Does this apply to both integer and floating point?  Should it return false if it fails?

TAO/TAO_IDL/include/ast_expression.h
```diff
@@ -279,10 +287,11 @@ class TAO_IDL_FE_Export AST_Expression
   // Evaluate then store value inside this AST_Expression.
   void evaluate (EvalKind ek);
 
-  // Compare two AST_Expressions.
-
+  /// Compare two AST_Expressions.
+  ///{
   bool operator== (AST_Expression *vc);
-  long compare (AST_Expression *vc);
+  bool compare (AST_Expression *vc);
```
mitza-oci, line 293: Was this previously a `strcmp` style return value (negative, 0, positive)?
If it's going to return `bool` would a different name be better?

TAO/tao/CDR.h
```diff
@@ -499,6 +499,8 @@ TAO_Export CORBA::Boolean operator<< (TAO_OutputCDR &os,
 TAO_Export CORBA::Boolean operator<< (TAO_OutputCDR &os,
                                       ACE_OutputCDR::from_std_wstring x);
 #endif /* ACE_LACKS_STD_WSTRING */
+TAO_Export CORBA::Boolean operator<< (TAO_OutputCDR &os, CORBA::Int8 x);
+TAO_Export CORBA::Boolean operator<< (TAO_OutputCDR &os, CORBA::Uint8 x);
```
mitza-oci, line 503: Are these going to result in ambiguous overload resolution?

mitza-oci (changes requested):

TAO/TAO_IDL/fe/fe_lookup.cpp
iguessthislldo: Yes, this is what happened when I generated it.

TAO/TAO_IDL/include/ast_expression.h
iguessthislldo: > Was this previously a strcmp style return value (negative, 0, positive)?

[No](https://github.com/DOCGroup/ACE_TAO/blob/943dd53bd48bf4a16c02f98d217c59bbb4b78ea7/TAO/TAO_IDL/ast/ast_expression.cpp#L2891)

> If it's going to return bool would a different name be better?

I think I added a `operator==` for it.

TAO/TAO_IDL/ast/ast_expression.cpp
iguessthislldo: > Does this apply to both integer and floating point?

Yes, I think so.

> Should it return false if it fails?

`true` is failure in this case.

mitza-oci: > 
> 
> > Does this apply to both integer and floating point?
> 
> Yes, I think so.

Then should floating point allow dividing by zero resulting in NaN or +/-Inf?

> 
> > Should it return false if it fails?
> 
> `true` is failure in this case.

yikes... how? why?

TAO/tao/CDR.h
iguessthislldo: As opposed to `char`? They probably will in ACE6 depending on the backing type, but with `int8_t` and `uint8_t` they seem to be fine. I was thinking of ifdefing them out in ACE6 if the types weren't stdint-backed.

TAO/TAO_IDL/ast/ast_expression.cpp
iguessthislldo: > Then should floating point allow dividing by zero resulting in NaN or +/-Inf?

[It didn't do that before](https://github.com/DOCGroup/ACE_TAO/blob/943dd53bd48bf4a16c02f98d217c59bbb4b78ea7/TAO/TAO_IDL/ast/ast_expression.cpp#L1882-L1887), but yes I could skip the check if it's a floating point.

> yikes... how? why?

I wrote like that because there has to be an action taken if there's a failure, but not for a success.

mitza-oci (changes requested):

TAO/TAO_IDL/include/ast_expression.h
iguessthislldo: Maybe, but I left it there for some level of compatibility.

TAO/TAO_IDL/be/be_helper.cpp
```diff
@@ -597,7 +612,7 @@ TAO_OutStream::print (AST_Expression *expr)
       this->TAO_OutStream::print ("L'%lc'", ev->u.wcval);
       break;
     case AST_Expression::EV_octet:
-      this->TAO_OutStream::print ("%d", ev->u.oval);
+      this->TAO_OutStream::print ("0x%x", ev->u.oval);
```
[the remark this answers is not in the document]
iguessthislldo: Could you tell me how would it be breaking?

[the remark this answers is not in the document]
mitza-oci: I'm not sure where this is used, so that's what I was asking the question.  If it's generating literals for the C++ output then I don't think it's going to matter.

TAO/TAO_IDL/ast/ast_expression.cpp
```diff
@@ -3049,6 +2868,7 @@ dump_expr_val (ACE_OSTREAM_TYPE &o, AST_Expression::AST_ExprValue *ev)
           o << "(null string)";
         }
       break;
+    case AST_Expression::EV_enum:
```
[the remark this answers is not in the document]
iguessthislldo: No. As far as TAO_IDL is concerned it's always a`uint32` like normally. I don't know why I put it here. Probably was a mistake.

TAO/TAO_IDL/ast/ast_expression.cpp
```diff
@@ -3003,6 +2800,27 @@ dump_unary_expr (ACE_OSTREAM_TYPE &o,
...
+    o << 'B';
+  } else if (value == 0xC) {
+    o << 'C';
+  } else if (value == 0xD) {
+    o << 'D';
+  } else if (value == 0xE) {
+    o << 'E';
+  } else if (value == 0xF) {
+    o << 'F';
+  }
```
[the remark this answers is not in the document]
iguessthislldo: [My version in OpenDDS](https://github.com/objectcomputing/OpenDDS/blob/3b1aa7d2d535a49349afa0e1d629ecb80960a7bc/dds/DCPS/SafetyProfileStreams.cpp#L116-L120) way is more compact. If using the standard library I'd have to use something like `std::hex` or `sprintf`.

Also I just noticed that I have the calls backwards where I'm using this function.

TAO/TAO_IDL/be/be_helper.cpp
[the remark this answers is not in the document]
iguessthislldo: This should only be used for C++ output.

ACE/ace/CDR_Base.h
```diff
@@ -207,6 +207,14 @@ class ACE_Export ACE_CDR
   typedef ACE_UINT32 ULong;
   typedef ACE_INT64 LongLong;
   typedef ACE_UINT64 ULongLong;
+  typedef ACE_INT8 Int8;
+  typedef ACE_UINT8 UInt8;
+  typedef Short Int16;
+  typedef UShort Uint16;
+  typedef Long Int32;
+  typedef ULong Uint32;
```
jwillemsen, line 215: `UInt32` and `UInt64` to keep it all consistent

iguessthislldo: Thought I did that already, but I guess it was just the int8 ones.

TAO/tests/IDLv4/explicit_ints/main.cpp
```diff
@@ -0,0 +1,79 @@
...
+  expect_equals<CORBA::Int8> (any_failed, "i8_e2", i8_e2, 4);
+  expect_equals<CORBA::Int8> (any_failed, "i8_e3", i8_e3, 12);
+  expect_equals<CORBA::Int8> (any_failed, "i8_e4", i8_e4, 3);
+  expect_equals<CORBA::Int8> (any_failed, "i8_e5", i8_e5, 7);
+  expect_equals<CORBA::Int8> (any_failed, "i8_e6", i8_e6, 1);
+  expect_equals<CORBA::Int8> (any_failed, "i8_e7", i8_e7, 1);
+  expect_equals<CORBA::Int8> (any_failed, "i8_e8", i8_e8, 16);
+  expect_equals<CORBA::Int8> (any_failed, "i8_e9", i8_e9, 8);
+
+  return any_failed ? 1 : 0;
```
mitza-oci, line 78: I've been using the `EXIT_SUCCESS` and `EXIT_FAILURE` constants from the standard library for this.

TAO/TAO_IDL/ast/ast_expression.cpp
```diff
@@ -3034,7 +2832,11 @@ dump_expr_val (ACE_OSTREAM_TYPE &o, AST_Expression::AST_ExprValue *ev)
       o << ev->u.wcval;
       break;
     case AST_Expression::EV_octet:
-      o << static_cast<int> (ev->u.oval);
+      {
+        char buffer[5] = {0};
+        ACE_OS::sprintf (buffer, "0x%02x", ev->u.oval);
+        o << &buffer[0];
```
mitza-oci, line 2838: If this is going to an `ostream` would it be cleaner to use its own formatting capabilities instead of `sprintf`?

mitza-oci (changes requested):

TAO/TAO_IDL/ast/ast_expression.cpp
iguessthislldo: I had to look everything up because I don't remember the last time I tried to use ostream formatting.

```c++
        std::ios saved(nullptr);
        saved.copyfmt(o);
        o << "0x" << std::hex << std::setw(2) << std::setfill('0') << unsigned(ev->u.oval);
        o.copyfmt(saved);
```

I'm not sure I'd call it cleaner. I'd definitely say much more verbose.

mitza-oci: Yes, it is more verbose... but it's using the object that's already involved in this operation instead of managing a separate buffer.

TAO/tests/IDLv4/explicit_ints/main.cpp
```diff
@@ -75,5 +76,5 @@ ACE_TMAIN (int, ACE_TCHAR *[])
   expect_equals<CORBA::Int8> (any_failed, "i8_e8", i8_e8, 16);
   expect_equals<CORBA::Int8> (any_failed, "i8_e9", i8_e9, 8);
 
-  return any_failed ? 1 : 0;
+  return any_failed ? EXIT_SUCCESS : EXIT_FAILURE;
```
mitza-oci, line 79: strike that and reverse it

iguessthislldo: Good catch

TAO/tao/CDR.h
mitza-oci: int8_t is a typedef/alias of some other type -- since there's already `from_octet` and `from_char` in `ACE_OutputCDR` shouldn't we have the same for u/int8?

iguessthislldo: Yea I wasn't thinking about this right. I will redo the serialization to make it like `octet`/`char`.

ACE/ace/Basic_Types.h
```diff
@@ -562,18 +562,21 @@ ACE_END_VERSIONED_NAMESPACE_DECL
 # endif /* ACE_SIZEOF_LONG_DOUBLE */
 
 // Max and min sizes for the ACE integer types.
-#define ACE_CHAR_MAX 0x7F
-#define ACE_CHAR_MIN -(ACE_CHAR_MAX)-1
-#define ACE_OCTET_MAX 0xFF
+#define ACE_INT8_MAX 0x7F
+#define ACE_INT8_MIN -(ACE_INT8_MAX) - 1
+#define ACE_UINT8_MAX 0xFF
```
jwillemsen, line 567: `ACE_UINT8_MIN`?

ACE/ace/Basic_Types.h
```diff
@@ -562,18 +562,21 @@ ACE_END_VERSIONED_NAMESPACE_DECL
 # endif /* ACE_SIZEOF_LONG_DOUBLE */
 
 // Max and min sizes for the ACE integer types.
-#define ACE_CHAR_MAX 0x7F
-#define ACE_CHAR_MIN -(ACE_CHAR_MAX)-1
-#define ACE_OCTET_MAX 0xFF
+#define ACE_INT8_MAX 0x7F
```
jwillemsen, line 565: We should convert these to `constexpr` at some point

TAO/TAO_IDL/ast/ast_expression.cpp
```diff
@@ -892,6 +948,15 @@ coerce_value (AST_Expression::AST_ExprValue *ev,
           ev->u.ulval = (unsigned long) ev->u.oval;
           ev->et = AST_Expression::EV_ulong;
           return ev;
+        case AST_Expression::EV_int8:
+          if (ev->u.int8val < 0) return 0;
```
jwillemsen, line 952: return nullptr? More places in this file

TAO/tao/Basic_Types_IDLv4.h
```diff
@@ -0,0 +1,58 @@
+#ifndef TAO_BASIC_TYPES_IDLV4_H
```
jwillemsen, line 1: Why not put these in `Basic_Types.h` and in the `CORBA` namespace, I think we should support `CORBA::Int8` without the requirement that the user includes this file, makes it much easier to use this construct which is also much clear in application code. For TAOX11 we could use this also without a special include and a namespace using.

jwillemsen (changes requested):
We shouldn't introduce a new `CORBA::IDLv4` namespace

TAO/tao/Basic_Types_IDLv4.h
mitza-oci: int8 is not a CORBA type (it has no Typecode, Any, IfR, DII, DSI, shouldn't be used for remote calls...).
For this whole group of type names, we wanted to make it explicit that these are different - tao_idl with default arguments will not accept them, they're not "peers" of the other names.

jwillemsen: Ok, we could keep Int8/UInt8 in this file, but the others are just typedefs which could be used without problems. That would enable TAOX11 to easily use these internally without a special include

ACE/ace/Basic_Types.h
iguessthislldo: We should, but I'd like that to be later because I don't want to make this harder to backport than it already is.

ACE/ace/CDR_Stream.h
```diff
@@ -791,6 +807,18 @@ class ACE_Export ACE_InputCDR
     ACE_CDR::Octet &ref_;
   };
 
+  struct ACE_Export to_int8
```
jwillemsen, line 810: Is this really required, isn't int8_t a distinct type itself?

TAO/tao/tao.mpc
```diff
@@ -60,6 +60,12 @@ project(TAO_Core_idl) : tao_versioning_idl_defaults, gen_ostream, install, pidl_
     WStringSeq.pidl >> AnyTypeCode/WStringSeqA.h AnyTypeCode/WStringSeqA.cpp
   }
 
+  IDL_Files {
+    idlflags += -Sci -Gse -Gata --idl-version 4
+    Int8Seq.pidl >> AnyTypeCode/Int8SeqA.h AnyTypeCode/Int8SeqA.cpp
```
jwillemsen, line 65: New Int8SeqA/UInt8SeqA needs to be added to AnyTypeCode

ACE/ace/CDR_Stream.h
mitza-oci: `(u)int*_t` are typedefs, probably at least one is the same as the one used for ACE_CDR's Octet, etc.

TAO/tao/tao.mpc
iguessthislldo: That's not in scope. Can I just remove those parts of the line?

jwillemsen: Are the files generates but empty? Then just add them, we shouldn’t have generated files which we don’t compile

mitza-oci (approved):
Still needs some CI fixes and to resolve @jwillemsen's comment about AnyTypeCode.  But the review itself is good to go.

jwillemsen (approved):

jwillemsen:
@iguessthislldo Several new warnings in the https://buildlogs.remedy.nl/win_vs2019_tao_debug/index.html build, can you have a look?

jwillemsen:
Can you also check the Coverity scan results, we have fresh results, several new issues related to this PR

iguessthislldo:
Yes, I will look into both, but for the warnings I don't think I will be able to tell if there are any warnings I need to fix outside the explicit_ints test.

PR #848, opened 2019-02-26
Fix wchar_t strnew and strdup.

likema:
char `strnew` and `strdup` have code of ACE_HAS_ALLOC_HOOKS, but wchar_t versions haven't.

### DIFF
--- a/ACE/ace/ACE.cpp
+++ b/ACE/ace/ACE.cpp
@@ -3240,11 +3240,18 @@ ACE::strndup (const wchar_t *str, size_t n)
        len++)
     continue;
 
+  size_t const size = (len + 1) * sizeof (wchar_t);
   wchar_t *s;
+#if defined (ACE_HAS_ALLOC_HOOKS)
   ACE_ALLOCATOR_RETURN (s,
-                        static_cast<wchar_t *> (
-            ACE_OS::malloc ((len + 1) * sizeof (wchar_t))),
+                        static_cast<wchar_t*> (
+                          ACE_Allocator::instance ()->malloc (size)),
                         0);
+#else
+  ACE_ALLOCATOR_RETURN (s,
+                        static_cast<wchar_t*> (ACE_OS::malloc (size)),
+                        0);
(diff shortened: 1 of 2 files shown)


ACE/ace/ACE.cpp
```diff
@@ -3241,10 +3241,15 @@ ACE::strndup (const wchar_t *str, size_t n)
...
   wchar_t *s;
+#if defined (ACE_HAS_ALLOC_HOOKS)
+  ACE_ALLOCATOR_RETURN (s,
+                        (wchar_t *) ACE_Allocator::instance()->malloc (len + 1),
+                        0);
+#else
   ACE_ALLOCATOR_RETURN (s,
-                        static_cast<wchar_t *> (
-            ACE_OS::malloc ((len + 1) * sizeof (wchar_t))),
+                        (wchar_t *) ACE_OS::malloc (len + 1),
```
jwillemsen: why sizeof remove?

ACE/ace/ACE.cpp
```diff
@@ -3241,10 +3241,15 @@ ACE::strndup (const wchar_t *str, size_t n)
     continue;
 
   wchar_t *s;
+#if defined (ACE_HAS_ALLOC_HOOKS)
+  ACE_ALLOCATOR_RETURN (s,
+                        (wchar_t *) ACE_Allocator::instance()->malloc (len + 1),
+                        0);
```
jwillemsen: would prefer a static_cast here instead of a c-style cast

likema: Fixed

ACE/ace/ACE.cpp
likema: Fixed

ACE/ace/ACE.cpp
likema: There are other c-style casts in ACE.cpp

Shall I fix them?

jwillemsen: No need to fix them all at this point, just when you work on code

PR #851, opened 2019-02-28
Avoid allocating buffer when printing log.

likema:


jwillemsen:
With small embedded systems couldn't this lead faster to a stack overflow?

likema:
The size of verbose buffer on stack is `VERBOSE_LEN` + 1 = `129`.

Is it too large for embedded systems?

```cpp
/// Size used by verbose mode.
/// 20 (date) + 15 (host_name) + 10 (pid) + 10 (type)
///           + 4 (@) ... + ? (progname)
VERBOSE_LEN = 128,
```

likema:
Maybe I can reduce it to `20 + 15 + 10 + 10 + 4 = 59` without progname by more complex code?

jwillemsen:
I now see that you replaced MAXVERBOSELOGMSGLEN with VERBOSE_LEN which is much smaller, was that the intent?

likema:
Yes. I separate a line of log message into a prefix (`verbose` buffer) and content so that only prefix needs a small buffer on stack.

likema:
I find that 

```
/// Size used by verbose mode.
/// 20 (date) + 15 (host_name) + 10 (pid) + 10 (type)
///           + 4 (@) ... + ? (progname)
VERBOSE_LEN = 128,
```

`progname` is not formatted into `verbose` buffer on stack. So `VERBOSE_LEN` can be reduce to 
20 + 15 + 10 + 10 + 4 = 59

ACE/ace/Log_Record.cpp
```diff
@@ -19,6 +19,14 @@
 
 #include "ace/OS_Memory.h"
 
+// The sprintf format needs to be different for Windows and POSIX
```
mitza-oci: Looks like this part would be useful in ace_wchar.h or OS_NS_stdio.h since other callers of ACE_OS::snprintf may need it.

likema: Which name is better for it?

* `ACE_TEXT_FORMAT_SPECIFIER`
* `ACE_STRING_FORMAT_SPECIFIER`
* `ACE_STR_FORMAT_SPECIFIER`
* `ACE_C_STRING_FORMAT_SPECIFIER`
* `ACE_C_STR_FORMAT_SPECIFIER`

mitza-oci: This was added in #1907

jwillemsen: Looks I remembered correctly that there was a macro already, it was just not merged yet, thanks for linking Adam

PR #854, opened 2019-03-03
Fix CID 1442918.

likema:


likema:
@jwillemsen What about it?

jwillemsen:
See https://www.remedy.nl for our services, reviewing issues takes time and we are looking for sponsoring for that.

PR #858, opened 2019-03-04
Try to Prevent Usage of System SSL on Android

iguessthislldo:
In a regular Android app, the system has preloaded the system SSL library (which can be any version of OpenSSL or BoringSSL) for use in the Android API. Even if OpenSSL is statically linked, the symbols that have already been loaded will take precedence and will be used instead. This change to `ace_openssl` on Android requires that OpenSSL symbols be kept hidden from the global linking, causing the symbols to be fixed to the OpenSSL library being included.


jwillemsen:
Looks the OCI android build is now failing, see http://scoreboard.ociweb.com/doc_master_firefly_android_clang_d1o0/index.html

PR #865, opened 2019-03-08
Remove ACE_HAS_DLSYM_SEGFAULT_ON_INVALID_HANDLE in config-linux.h

likema:
In Linux, it is valid to `dlsym` with `RTLD_DEFAULT` that is equal to
`ACE_SHLIB_INVALID_HANDLE`.


jwillemsen:
Can you extend one of the ACE unit tests to call ACE_OS::dlsym with an invalid handle so that we know it doesn't segfault?

likema:
Added test case into DLL_Test

ACE/tests/DLL_Test.cpp
```diff
@@ -168,6 +168,11 @@ run_main (int, ACE_TCHAR *[])
               ACE_TEXT ("Dynamically Linkable Libraries not supported on this platform\n")));
 #endif /* ACE_HAS_DYNAMIC_LINKING */
 
+#if defined(ACE_LINUX)
+  ACE_TEST_ASSERT (RTLD_DEFAULT == ACE_SHLIB_INVALID_HANDLE);
+  ACE_TEST_ASSERT (ACE_OS::dlsym (ACE_SHLIB_INVALID_HANDLE, "open"));
```
jwillemsen: Why not test this on all platforms?

likema: Fixed

PR #867, opened 2019-03-08
Fix ACE_Thread_Manager::join memory leak and deadlock.

likema:
* Fix `ACE_Thread_Descriptor_Base` object leak when failing to join thread.
* Fix deadlock in `ACE_Thread_Manager::exit` when thread is exiting and being join on the same time.

### DIFF
--- a/ACE/ace/Thread_Manager.cpp
+++ b/ACE/ace/Thread_Manager.cpp
@@ -1496,14 +1496,18 @@ ACE_Thread_Manager::join (ACE_thread_t tid, ACE_THR_FUNC_RETURN *status)
       {
         if (ACE_OS::thr_equal (biter.next ()->thr_id_, tid))
           {
-            ACE_Thread_Descriptor_Base *tdbl = biter.advance_and_remove (false);
+#if defined (ACE_HAS_CPP11)
+            std::unique_ptr<ACE_Thread_Descriptor_Base> tdbl (biter.advance_and_remove (false));
+#else
+            auto_ptr<ACE_Thread_Descriptor_Base> tdbl (biter.advance_and_remove (false));
+#endif /* ACE_HAS_CPP11 */
+            ace_mon.release();
 #ifndef ACE_LACKS_PTHREAD_JOIN
             if (ACE_Thread::join (tdbl->thr_handle_, status) == -1)
               {
                 return -1;
               }
 #endif
-            delete tdbl;
 
             // return immediately if we've found the thread we want to join.
             return 0;


likema:
Deadlock occurs:

https://github.com/DOCGroup/ACE_TAO/blob/master/ACE/ace/Thread_Manager.cpp#L1686-L1704

likema:
https://github.com/DOCGroup/ACE_TAO/blob/fa4af72dade90897e8508bb88b1908339594763d/ACE/ace/Thread_Manager.cpp#L1686-L1704

ACE/ace/Thread_Manager.cpp
```diff
@@ -1496,14 +1496,14 @@ ACE_Thread_Manager::join (ACE_thread_t tid, ACE_THR_FUNC_RETURN *status)
       {
         if (ACE_OS::thr_equal (biter.next ()->thr_id_, tid))
           {
-            ACE_Thread_Descriptor_Base *tdbl = biter.advance_and_remove (false);
+            ACE_Auto_Ptr<ACE_Thread_Descriptor_Base> tdbl (biter.advance_and_remove (false));
+            ace_mon.release();
```
jwillemsen: When we have c++11 use a unique_ptr, see similar constructs in ACE

likema: Do you mean `ACE_Auto_Ptr` is deprecated?

Use auto ptr similar to https://github.com/DOCGroup/ACE_TAO/commit/8f2464a0b24466f850f36a7f7ad5898c883c67a0

jwillemsen: Yes, with cpp11 this gives a warning, with cpp17 an error

likema: Fixed

likema:
Why not define a macro for auto ptr? Just like:

```cpp
#if defined(ACE_HAS_CPP11)
#define AUTO_PTR(TYPE, VAR) std::unique_ptr<TYPE> VAR
#else
#define AUTO_PTR(TYPE, VAR) auto_ptr<TYPE> VAR
#endif
```

jwillemsen:
There are several references on the web that describe that autoptr can’t be replaced with unique_ptr in all cases, using such a macro could cause problems

likema:
I see.

jwillemsen:
@likema My time for free reviews and the resulting merge of a PR is very limited. After merging a PR I also commit myself to check the scoreboard for a few days and coordinate/resolve any breakage from a change. As much as I appreciate all your PRs, when you would like me to look at them with the attention they deserve consider hiring me through @RemedyIT, see https://www.remedy.nl/services/opensource-support.html, than I can really look at all of them and work with you to get them reviewed and merged

jwillemsen (approved):

PR #868, opened 2019-03-09
Fix ACE_UNIX_Addr size.

likema:


jwillemsen:
Any test extension possible to reproduce this and make sure the size is correct?

likema:
I haven't found any test cases about `ACE_UNIX_Addr`.

I will add a new one for it later.

likema:
Added

PR #877, opened 2019-03-18
Fix Android Linking Errors

iguessthislldo:
Fixes linking errors caused by enabling hidden visibility in #858 which I unfortunately must have done without testing. This does build on NDK 19 and I'm [testing it OpenDDS-Android as well](https://travis-ci.org/iguessthislldo/OpenDDS-Android/builds/506532752).

I changed it to match `config-linux.h` below and let `config-g++-common.h` do its job. For me this confirms the fact that there should be a `config-linux-common.h` like there is a `platform_linux_common.GNU`. I would like to do that in the future but I need to sort out what's necessary to keep for Android and what is leftover from the `config-linux.h` when they diverged.

### DIFF
--- a/ACE/ace/config-android.h
+++ b/ACE/ace/config-android.h
@@ -209,13 +209,6 @@
   // this must appear before its #include.
 # define ACE_HAS_STRING_CLASS
 # include "ace/config-g++-common.h"
-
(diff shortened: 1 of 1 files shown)


iguessthislldo:
[The builds were successful](https://travis-ci.org/iguessthislldo/OpenDDS-Android/builds/506532752)

PR #887, opened 2019-04-12
Linux: Refactor config-linux.h and use gettid() for %t for Log_Msg

iguessthislldo:
On Linux systems that support it, use `gettid()` to get a much smaller number for the thread id.

The main motivator for this PR, besides slightly smaller logs for OpenDDS, is the fact that gdb on linux has two identifiers for threads beside the simple thread index; One is LWP (Light Weight Process), which is another name for the TID that `gettid()` returns, and other is the pthread_t value in hex.

Also refactored `config-android.h` and `config-linux.h`, pulling out the common parts into a `config-linux-common.h` so that the check for gettid could go into the common file.

### DIFF
--- a/ACE/NEWS
+++ b/ACE/NEWS
@@ -1,8 +1,8 @@
 USER VISIBLE CHANGES BETWEEN ACE-6.5.5 and ACE-6.5.6
 ====================================================
 
-. Shorten output from `%t` (current thread id) for ACE_Log_Msg on non-Windows
-  systems
+. On Linux, the ACE_Log_Msg format specifier `%t` is now replaced with the
+  system thread id provided by gettid(), instead of the much longer pthread id.
 
 USER VISIBLE CHANGES BETWEEN ACE-6.5.4 and ACE-6.5.5
 ====================================================
--- a/ACE/NEWS
+++ b/ACE/NEWS
@@ -1,6 +1,9 @@
 USER VISIBLE CHANGES BETWEEN ACE-6.5.5 and ACE-6.5.6
 ====================================================
 
+. Shorten output from `%t` (current thread id) for ACE_Log_Msg on non-Windows
+  systems
+
 USER VISIBLE CHANGES BETWEEN ACE-6.5.4 and ACE-6.5.5
 ====================================================
 
--- a/ACE/ace/Log_Msg.cpp
+++ b/ACE/ace/Log_Msg.cpp
@@ -1795,13 +1795,15 @@ ACE_Log_Msg::log (const ACE_TCHAR *format_str,
 
 #  ifdef ACE_HAS_GETTID
 #    define ACE_LOG_MSG_GET_THREAD_ID ACE_OS::thr_gettid
+#    define ACE_LOG_MSG_GET_THREAD_ID_BUFFER_SIZE 8
 #  else
 #    define ACE_LOG_MSG_GET_THREAD_ID ACE_OS::thr_id
+#    define ACE_LOG_MSG_GET_THREAD_ID_BUFFER_SIZE 32
 #  endif
 
 #  if defined ACE_USES_WCHAR
                   {
-                    char tid_buf[32] = {};
+                    char tid_buf[ACE_LOG_MSG_GET_THREAD_ID_BUFFER_SIZE] = {};
                     ACE_LOG_MSG_GET_THREAD_ID (tid_buf, sizeof tid_buf);
                     this_len = ACE_OS::strlen (tid_buf);
                     ACE_OS::strncpy (bp, ACE_TEXT_CHAR_TO_TCHAR (tid_buf),
--- a/ACE/ace/Log_Msg.cpp
+++ b/ACE/ace/Log_Msg.cpp
@@ -1791,16 +1791,25 @@ ACE_Log_Msg::log (const ACE_TCHAR *format_str,
                       ACE_OS::sprintf (bp,
                                        format,
                                        static_cast <unsigned> (ACE_Thread::self ()));
-#elif defined ACE_USES_WCHAR
+#else
+
+#  ifdef ACE_HAS_GETTID
+#    define ACE_LOG_MSG_GET_THREAD_ID ACE_OS::thr_gettid
+#  else
+#    define ACE_LOG_MSG_GET_THREAD_ID ACE_OS::thr_id
+#  endif
+
+#  if defined ACE_USES_WCHAR
                   {
                     char tid_buf[32] = {};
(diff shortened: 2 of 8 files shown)


ACE/ace/OS_NS_Thread.cpp
```diff
@@ -4806,6 +4807,18 @@ ACE_OS::unique_name (const void *object,
 }
 #endif
 
+#ifdef ACE_HAS_GETTID
+pid_t
+ACE_OS::thr_gettid ()
```
mitza-oci: Consider if this should be provided unconditionally, and if it should just fall back to tid on other platforms/versions.

ACE/NEWS
```diff
@@ -1,6 +1,9 @@
 USER VISIBLE CHANGES BETWEEN ACE-6.5.5 and ACE-6.5.6
 ====================================================
 
+. Shorten output from `%t` (current thread id) for ACE_Log_Msg on non-Windows
```
mitza-oci: Could use some more description of which identifier is actually used (on platforms that support it).

ACE/ace/OS_NS_Thread.inl
```diff
@@ -3178,12 +3178,21 @@ ACE_OS::thr_id (char buffer[], size_t buffer_length)
 #else /* ACE_HAS_OPAQUE_PTHREAD_T */
   return ACE_OS::snprintf (buffer,
                            buffer_length,
-                           "%lu",
+                           "%lx",
```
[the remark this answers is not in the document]
iguessthislldo: This was what I originally just wanted to do to have the thread id match what is used in gdb and still want to do for platforms like macOS. I'm fairly certain It's not a breaking change as far as ACE/TAO is concerned. There is the possibility some user code is relying on this being in decimal, but user code could also be relying on `%t` being the posix thread id.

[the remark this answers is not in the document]
iguessthislldo: Thinking about this more I don't know how other platforms thread ids look like, so I don't think I want to change that for now. I will limit this PR to Linux that has gettid.

iguessthislldo:
It passed all checks but Azure's status on one job is stuck. If using `ACE_NOTSUP_RETURN(-1)` on platforms that don't have `gettid` is okay I think this is ready to be merged.

As a side note Linux 2.4.11, the version I'm checking for, was [released in 2001](https://linux.slashdot.org/story/01/10/10/0134219/kernel-2411-released). I know ACE is older than that, but I'm curious if ACE will compile on a system that old today.

iguessthislldo:
@mitza-oci It's been a month, is this good to go?

mitza-oci:
> @mitza-oci It's been a month, is this good to go?

I'm OK with it.  Does anyone else in the maintainer/user community want to review?

ACE/ace/OS_NS_Thread.cpp
```diff
@@ -4806,6 +4807,16 @@ ACE_OS::unique_name (const void *object,
 }
 #endif
 
+pid_t
+ACE_OS::thr_gettid ()
+{
+#if defined (ACE_LINUX) && defined (ACE_HAS_GETTID)
```
jwillemsen: Is a check of just ACE_HAS_GETTID not enough?

ACE/ace/Log_Msg.cpp
```diff
@@ -1791,16 +1791,25 @@ ACE_Log_Msg::log (const ACE_TCHAR *format_str,
...
+
+#  ifdef ACE_HAS_GETTID
+#    define ACE_LOG_MSG_GET_THREAD_ID ACE_OS::thr_gettid
+#  else
+#    define ACE_LOG_MSG_GET_THREAD_ID ACE_OS::thr_id
+#  endif
+
+#  if defined ACE_USES_WCHAR
                   {
                     char tid_buf[32] = {};
```
jwillemsen: When ACE_HAS_GETTID is defined, do we need a buffer of size 32 or can we use a smaller buffer?

ACE/ace/OS_NS_Thread.cpp
iguessthislldo: At the moment it would be. `syscall(SYS_gettid)` is basically the only way to use gettid, but is very specific to Linux. I was thinking of `ACE_HAS_GETTID` also being able to be defined if another platform had gettid that someone wanted to had.

ACE/ace/Log_Msg.cpp
iguessthislldo: tids are drawn from the same pool as pids. It seems the maximum possible limit on that pool for 64bit systems is 2^22 or 4194304, 5 digits. So yes it could be smaller if 24 bytes (assuming we make it 8) is worth shaving off the stack.

iguessthislldo: > 4194304, 5 digits

Apparently I forgot how to count. I meant to say 7.

iguessthislldo:
@jwillemsen I made the changes you requested.

I know this won't help this merged faster, but I wanted to enable `gettid` in Android, since it can be. Instead of copying the check across the config files, I created `config-linux-common.h` from the common parts of `config-linux.h` and `config-android`. There are things in Linux that might be able to be added to Android, but my goal was to keep the resultant includes basically equivalent as before. This has been something I've been wanting to do since I started working on the Android stuff and I saw this as a good point to do it. I will have my Android build matrix build this branch tonight as a extra check.

ACE/ace/OS_NS_Thread.cpp
```diff
@@ -17,6 +17,7 @@
 #include "ace/Thread_Mutex.h"
 #include "ace/Condition_Thread_Mutex.h"
 #include "ace/Guard_T.h"
+#include "ace/OS_NS_sys_resource.h"
```
jwillemsen: Conditional include?

iguessthislldo:
I've been fixing some Android stuff, because of course the things I thought weren't going to impact Android did, especially with the older NDKs which had ancient Android headers. I'm running OpenDDS-Android again and will report here if the remaining failing combinations pass.

iguessthislldo:
[I got it to pass the Android builds](https://travis-ci.org/iguessthislldo/OpenDDS-Android/builds/542351765)

iguessthislldo:
As a postscript to this, I just discovered there once was a `config-linux-common.h` which was removed in 9ec55c56dd40b42d353d9bd8b9448c0fefc078ce, while `config-android.h` was created in 0ef33e023ae1b4f28790fb0c27f90091a7608c11 shortly afterwards, and both of these are for the same version of ACE, 6.3.4

PR #896, opened 2019-04-27
Fix unresolved symbol with MinGW and fix linux warning

jwillemsen:


TAO/orbsvcs/FTRT_Event_Service/Event_Service/FT_EventService.cpp
```diff
@@ -248,7 +248,7 @@ FT_EventService::report_factory(CORBA::ORB_ptr orb,
       ORBSVCS_DEBUG((LM_DEBUG,"Factory connected\n"));
       CORBA::String_var my_ior_string = orb->object_to_string(ec);
 
-      size_t const len = ACE_OS::strlen(my_ior_string.in()) ;
+      int const len = ACE_OS::strlen(my_ior_string.in()) ;
```
iguessthislldo: But `strlen` returns `size_t`... what was the warning about?

iguessthislldo: After looking at this closer I'm guessing it's warning about comparing it with the resulting `ssize_t` from `send_n`? Regardless I don't like that send_n is taking in a `size_t` and returning a correlating `ssize_t` just for the case of error handling.

jwillemsen: Yes, comparison size_t and ssize_t, decided to first get all builds green, hopefully at some point we get some sponsoring to fix all 64bit windows warnings

PR #897, opened 2019-04-27
TAO/TAO_IDL/docs/annotations.md : typo fixes

okellogg:
On compiling the `union Grade` example, tao_idl complains about the missing keyword *switch*.
Further, tao_idl complains `identifier spellings differ only in case: "grade" and "Grade"` but I didn't fix that one, assuming use of the tao_idl switch *-Cw* to soften the error into a warning.


iguessthislldo:
Yea, I definitely tested the full example. but I don't think I ever actually tried to compile the special case examples unfortunately, thanks for catching these.

jwillemsen:
Thanks!

mitza-oci:
> identifier spellings differ only in case: "grade" and "Grade"

We should fix this so that the documentation demonstrates the recommended usage.

PR #923, opened 2019-06-12
Allow for anonymous types as per IDLv4

ceneblock:
Specific section is 7.4.14 in IDL4.2

### DIFF
--- a/TAO/TAO_IDL/util/utl_err.cpp
+++ b/TAO/TAO_IDL/util/utl_err.cpp
@@ -1508,19 +1508,21 @@ UTL_Error::anonymous_type_diagnostic (void)
     {
       return;
     }
-
-  if (idl_global->anon_warning ())
+  if (idl_global->idl_version_ < IDL_VERSION_4)
     {
-      if (idl_global->print_warnings ())
+      if (idl_global->anon_warning ())
         {
-          idl_warning_header (EIDL_ANONYMOUS_WARNING);
-          ACE_ERROR ((LM_WARNING, "\n"));
+          if (idl_global->print_warnings ())
+            {
+              idl_warning_header (EIDL_ANONYMOUS_WARNING);
+              ACE_ERROR ((LM_WARNING, "\n"));
+            }
         }
-    }
-  else
-    {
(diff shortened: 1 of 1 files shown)


jwillemsen:
Anonymous types are not part of the CORBA building block set so we should at least give a warning

mitza-oci:
> Anonymous types are not part of the CORBA building block set so we should at least give a warning

Good point, we didn't get to discuss the building blocks yet.  In terms of TAO_IDL I think we could have the backend determine which building blocks are supported.  I don't know if need to explicitly declare which building blocks or just have different switches that can be toggled by the backend initialization code.

iguessthislldo:
I'm a little confused. IDL reallowed anonymous sequence and arrays, is CORBA contradicting that? Even if that is the case, IDL4 is not the default, so why should we warn about something they've already implicitly agreed to?

jwillemsen:
Personally I was not in favor of allowing anonymous types in IDL4, from the DDS side it was pushed to allow it. Some language mappings don’t support them, for example IDL to C++11 doesn’t allow anonymous types at all. Using anonymous types as user is asking for problems at some point in the future.

mitza-oci:
> I'm a little confused. IDL reallowed anonymous sequence and arrays, is CORBA contradicting that? Even if that is the case, IDL4 is not the default, so why should we warn about something they've already implicitly agreed to?

Building blocks allow different implementations to essentially pick and choose features from IDL.  A future version of TAO could use IDL4 without anonymous types.

iguessthislldo:
> Some language mappings don’t support them

> Building blocks allow different implementations to essentially pick and choose features from IDL.

Alright, these two points together makes sense.Yes, we could have a interface that gets called in `BE_init` or equivalent that controls what building blocks where enabled. It would be a pain to make every building block in IDL4 controllable though.

iguessthislldo:
I meant to update this earlier Adam, you were right earlier, tao_idl already the mechanism we need: `idl_global->anon_type_diagnostic` for setting accept, warn, or error with corresponding command line options, so I think this PR can be closed.

PR #929, opened 2019-06-26
Increase the ACE_LOG_MSG_GET_THREAD_ID_BUFFER_SIZE from 8 to 12 as an…

ocielliottc:
… int is used in the implementation of ACE_OS::thr_gettid and could require 11 bytes of text, including a sign, and a byte to null terminate the string.


iguessthislldo:
It really shouldn't since the [possible value from gettid is limited on Linux](https://github.com/DOCGroup/ACE_TAO/pull/887#discussion_r290418807), but I'm not against increasing the size just in case.

ocielliottc:
I'm only basing this on the number of digits that could be used in representing an int in a text format.  I don't know what the actual limit is or what it may be in the future.

ocielliottc:
So, yes, while right now 8 bytes is probably sufficient, future implementations of ACE_OS::thr_gettid() could return values using the full range of an integer.

PR #939, opened 2019-08-14
config-linux-common.h: fix glibc build failure

saininav:
Recent glibc v2.30 dropped XSI STREAMS declarations,
which causing below build failure.

poky/build/tmp/work/corei7-64-poky-linux/ace/6.5.6-r0/ACE_wrappers/ace/os_include/os_stropts.h:56:17: fatal error: stropts.h: No such file or directory
   56 | #  include /**/ <stropts.h>
      |                 ^~~~~~~~~~~
compilation terminated.

Added GLIBC checks for affected versions.

For more information about glibc v2.30 change, please check:
https://sourceware.org/git/?p=glibc.git;a=commit;h=a0a0dc83173ce11ff45105fd32e5d14356cdfb9c

Signed-off-by: Naveen Saini <naveen.kumar.saini@intel.com>

### DIFF
--- a/ACE/ace/config-linux-common.h
+++ b/ACE/ace/config-linux-common.h
@@ -173,7 +173,7 @@
 
 // Starting with FC9 rawhide this file is not available anymore but
 // this define is set
-#if defined _XOPEN_STREAMS && _XOPEN_STREAMS == -1
+#if !defined(_XOPEN_STREAMS) || (defined _XOPEN_STREAMS && _XOPEN_STREAMS == -1)
 #  define ACE_LACKS_STROPTS_H
 #  define ACE_LACKS_STRRECVFD
 #endif


ACE/ace/config-linux-common.h
```diff
@@ -173,7 +173,7 @@
 
 // Starting with FC9 rawhide this file is not available anymore but
 // this define is set
-#if defined _XOPEN_STREAMS && _XOPEN_STREAMS == -1
+#if (defined(__GLIBC__) && (__GLIBC__ == 2 && __GLIBC_MINOR__ >= 30)) || (defined _XOPEN_STREAMS && _XOPEN_STREAMS == -1)
```
jwillemsen: What when glibc 3.x gets released?

jwillemsen: Is _XOPEN_STREAMS defined with glibc v2.30? If not, maybe check !_XOPEN_STREAMS?

iguessthislldo: I like that solution much better than glibc specific check. Regardless, @saininav do you know why glibc dropped this macro? If we could pass macros to glibc to get it back, it would be an even better solution.

jwillemsen:
```suggestion
#if !defined(_XOPEN_STREAMS) || (defined _XOPEN_STREAMS && _XOPEN_STREAMS == -1)
```

iguessthislldo: From Glibc 2.30 News File:

> * The obsolete and never-implemented XSI STREAMS header files <stropts.h> and <sys/stropts.h> have been removed.

Nevermind about getting it back I guess. However if it was never implemented does that mean this never actually worked with glibc?

saininav: I also think so, as _XOPEN_STREAMS == -1 was true in previous versions.

saininav: Agreed with suggested changes, as _XOPEN_STREAMS in not defined in newer release. If everyone is agree, I can send v2 patch.

jwillemsen: You should be able to commit my suggestion which automatically updates this PR

jwillemsen:
Thanks for your contribution

PR #945, opened 2019-08-21
Optimization: avoid nested iteration in ACE_Service_Repository::find_i

mgrojo:
This method was performing two iterations: the first using an index, and
the second in the call to this->service_array_.find, which uses an
iterator over the Array_Map looking for the element with that index.

This simplifies and optimizes this method by iterating for the element
once using directly the map iterator.


mgrojo:
The callgrind reports in my application showed a suspicious amount of cost in ACE_Array_Map<...>::find and ACE_Service_Repository::find_i, which is reduced with this change.

jwillemsen:
Thanks!

PR #949, opened 2019-08-21
Optimization: avoid invoking gettimeofday in each loop cycle

mgrojo:
Constants can be moved outside the loop so delay is calculated just one, and
gettimeofday is only called once for each invocation to cleanup (it is
assumed that the difference in time for each loop cycle is irrelevant for
the cleanup strategy).

Callgrind reports of my application showed an unexpected high cost in this function. This alleviates the problem.


jwillemsen:
Thanks!

PR #950, opened 2019-08-26
Support for TLS 1.2

sonnin:
TLS version 1.2 has existed since openSSL 1.0.1 (2012). This is preferred over the old/unsafe SSL v2 and v3.


jwillemsen:
Looks you are reverting a different PR, see https://github.com/DOCGroup/ACE_TAO/pull/156 for background info why this got removed some years ago

sonnin:
Aha, thanks for pointing that out.
I was in the progress of upgrading from ACE 6.1 which did not have this kind of support. I will adapt our code to the new filter_versions function. It sound like this can restrict usage of older SSL versions.

PR #952, opened 2019-08-26
Added SSL read/write queues.

sonnin:
Comments in the code states below:
// only one read operation is allowed now
// later it will be possible to make a queue

Somehow our usage of ACE 6.1.7 hits this limitation in some situations. Resulting in read/write errors.

Adding read/write queues resolves this limitation.

### DIFF
--- a/ACE/ace/SSL/SSL_Asynch_Stream.cpp
+++ b/ACE/ace/SSL/SSL_Asynch_Stream.cpp
@@ -95,8 +95,6 @@ ACE_SSL_Asynch_Stream::ACE_SSL_Asynch_Stream (
   : type_         (s_type),
     proactor_     (0),
     ext_handler_  (0),
-    ext_read_result_ (0),
-    ext_write_result_(0),
     flags_        (0),
     ssl_          (0),
     handshake_complete_(false),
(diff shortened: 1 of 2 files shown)


jwillemsen:
Some fuzz errors
```
Running NULL usage check
Error: ./ACE/ace/SSL/SSL_Asynch_Stream.cpp:342: NULL found
Error: ./ACE/ace/SSL/SSL_Asynch_Stream.cpp:382: NULL found
```

ACE/ace/SSL/SSL_Asynch_Stream.cpp
```diff
@@ -339,7 +339,7 @@ ACE_SSL_Asynch_Stream::read (ACE_Message_Block & message_block,
   if (this->flags_ & SF_REQ_SHUTDOWN)
     return -1;
 
-  ACE_SSL_Asynch_Read_Stream_Result* result = NULL;
+  ACE_SSL_Asynch_Read_Stream_Result* result = nullptr;
```
jwillemsen, line 342: nullptr is C++11 only, please use 0

jwillemsen:
Do you have an extension or new ACE unit test to validate this new code?

ACE/ace/SSL/SSL_Asynch_Stream.cpp
sonnin: Ok, I have changed this

sonnin:
> Do you have an extension or new ACE unit test to validate this new code?

No.

The existing test should continue to work. But will only test with one message in the queue. Hence the FIFO logic is not validated by test cases.

jwillemsen:
Can you extend the exiting test to also test the FIFO logic?

PR #954, opened 2019-08-29
Improved support on Windows platforms with differing IPv4/IPv6 support

wkbrd:
Improved support on Windows platforms with differing IPv4/IPv6 support, including over network segments.  It was found that on Windows sometimes IPv4 traffic would not connect due to the results of name lookups.  

This change corrects this behavior and we have been utilizing the same patched files on both Windows and Linux.

### DIFF
--- a/ACE/ace/INET_Addr.cpp
+++ b/ACE/ace/INET_Addr.cpp
@@ -408,7 +408,11 @@ ACE_INET_Addr::set (u_short port_number,
 
   addrinfo hints;
   ACE_OS::memset (&hints, 0, sizeof hints);
+#if defined (ACE_WIN32)
+  hints.ai_family = AF_UNSPEC;
+#else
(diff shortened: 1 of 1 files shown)


wkbrd:
Hi @jwillemsen, what is the process for having this pull request be considered?  Does the code complexity (CodeFactor) need to be addressed or can it be accepted by a project team member's review?

iguessthislldo:
The complex method complaint is normal and personally I think it's being over zealous in this case since the function is only about 120 lines long, although it's probably counting the possible code paths. As for acceptance, it's not directly based on what checks are saying, someone who with free time will have to review it.

jwillemsen:
Acceptance required a detailed review which depens on available time. Also the person who merges a PR implicitly commits to check all builds to watch for any regression. All together it takes several hours for one PR.

PR #962, opened 2019-09-24
Workaround is still necessary with clang9

jwillemsen:
* ACE/tests/Compiler_Features_32_Test.cpp:


jwillemsen:
@mitza-oci Fixes an error in the OCI clang build which seems to use clang9 now

PR #963, opened 2019-09-26
Use ACE_OS::stat for ACE_OS::filesize.

likema:
The original `ACE_OS::filesize` opens file with `O_RDONLY`, which would
fail to open CACHE.DATA of [InterSystems Caché ](https://en.wikipedia.org/wiki/InterSystems_Cach%C3%A9) in AIX.

To open it successfully, it requires `O_CIO` additionally which is not
portable even for AIX.


likema:
@jwillemse What about this PR?

jwillemsen:
Need to review this more closely but lack time for free reviews at this moment.

PR #967, opened 2019-10-05
TAO IDL: Interface Annotations

iguessthislldo:
Updated `tao_idl` to support annotations on interfaces, operations, and attributes.

Also Updated annotations.md With:
- a list of what can be annotated
- how to extend annotation support in the future
- other minor fixes

# TODO

- [x] Build TAO and OpenDDS with this to make sure I didn't break interfaces
- [x] Run Annotation Test with Memory Checks
- [x] Update TAO NEWS File


iguessthislldo:
So the fuzz has a check for `#include <...>`. I want to know if I can disable this check, because I don't think this convention is right since it forces the compiler to check the current directory for the header is definitely not there. `fuzz.pl` just says:

> This check is suggested by Don Hinton to force user to use # " " instead of <> to avoid confict with Doxygen.

What kind of conflict was it and will it still happen? Regardless this sounds like Doxygen was/is being configured incorrectly to me.

TAO/TAO_IDL/docs/annotations.md
```diff
@@ -345,11 +375,11 @@ these lines would also need to be added:
   if (document)
     {
       const char *comment =
-        AST_Annotation_Member::narrow_from_decl ((*document)["comment"])->
+        dynamic_cast<AST_Annotation_Member *> ((*document)["comment"])->
```
knutpett: Out of curiosity, why do you use dynamic_cast instead of narrow_from_decl?

TAO/TAO_IDL/include/utl_err.h
```diff
@@ -191,6 +191,20 @@ class TAO_IDL_FE_Export UTL_Error
                  AST_Decl *t2,
                  AST_Decl *t3);
 
+  /**
+   * Report an error that can be out of context of IDL parsing.
+   */
+  void direct_error (
+    const char *reason, ACE_CString filename, long lineno,
```
jwillemsen: pass filename as const&?

TAO/TAO_IDL/docs/annotations.md
iguessthislldo: `narrow_from_decl` is defined in `include/idl_narrow.h` where it's just a `dyanmic_cast`. It's a left over from an old narrowing system from the original pre C++98 compiler when they didn't apparently have `dynamic_cast`.

iguessthislldo: I guess I should also say that we should prefer standard C++ over pre-standard stuff whenever it's possible to.

jwillemsen: Maybe make a todo to make this a global change in TAO_IDL to simplify the code overall

TAO/TAO_IDL/util/utl_global.cpp
```diff
@@ -1942,8 +1942,11 @@ IDL_GlobalData::eval (const char *string)
   idl_global->set_lineno (1);
   UTL_String *old_idl_src_file = idl_src_file ();
 
-  // Name this pseudo-file "builtin"
-  UTL_String utl_string ("builtin", true);
+  // Name this pseudo-file "builtin-N"
+  static char buffer[64];
+  static unsigned n = 1;
+  ACE_OS::sprintf (&buffer[0], "builtin-%d", n++);
```
mitza-oci: prefer snprintf

TAO/TAO_IDL/docs/annotations.md
```diff
@@ -74,16 +78,42 @@ struct Report {
...
-though is that other than before normal declarations, like before structs,
-typedefs, and constants, in TAO\_IDL as of writing, annotations can also be
-applied in the following cases. See ["Reading Special Cases of
-Annotations"](#reading-special-cases-of-annotations) for how to have the
-compiler use these kinds of annotations.
+IDL 4.2 [(Limitation #1)](#limitation1). Because of the lack of standardization
+of what can be annotated and how, annotations for specific IDL elements have to
+be added on a as-needed basis.
+
+As of writing you can put annotations before the following things in IDL and a
```
mitza-oci:
```suggestion
You can put annotations before the following things in IDL and a
```

TAO/TAO_IDL/docs/annotations.md
```diff
@@ -74,16 +78,42 @@ struct Report {
...
+backend using TAO IDL should be able to read them:
```
mitza-oci:
```suggestion
backend using TAO IDL's front-end library can use them:
```

TAO/TAO_IDL/fe/idl.ypp
```diff
@@ -912,41 +914,47 @@ template_module_inst
 
 interface_def
         : interface
+        {
+          $$ = $1;
```
mitza-oci: Is this needed?  IIRC Bison does this by default.

TAO/TAO_IDL/fe/idl.ypp
```diff
@@ -912,41 +914,47 @@ template_module_inst
...
-                                        $1->inherits (),
-                                        $1->n_inherits (),
-                                        $1->inherits_flat (),
-                                        $1->n_inherits_flat (),
-                                        $1->is_local (),
-                                        $1->is_abstract ()
-                                      );
-              AST_Interface::fwd_redefinition_helper (i,
-                                                      s);
+          if (s != 0 && header)
```
mitza-oci:
```suggestion
          if (s && header)
```
or the other way round, for consistency

TAO/TAO_IDL/fe/idl.ypp
iguessthislldo: I thought it did, but I was having trouble with the annotation working and I put these in to make sure it was passing the value.

TAO/TAO_IDL/docs/annotations.md
iguessthislldo: Between these two changes, this sentence is a lot more confident then mine.

TAO/TAO_IDL/docs/annotations.md
```diff
@@ -74,16 +78,42 @@ struct Report {
...
-typedefs, and constants, in TAO\_IDL as of writing, annotations can also be
-applied in the following cases. See ["Reading Special Cases of
-Annotations"](#reading-special-cases-of-annotations) for how to have the
-compiler use these kinds of annotations.
+IDL 4.2 [(Limitation #1)](#limitation1). Because of the lack of standardization
+of what can be annotated and how, annotations for specific IDL elements have to
+be added on a as-needed basis.
+
+You can put annotations before the following things in IDL and a backend using
+TAO IDL's fround-end library can read them:
```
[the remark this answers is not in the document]
iguessthislldo: Darn

iguessthislldo:
Sure, it's something I thought about doing when I originally created it, but never quite go around to doing.

PR #970, opened 2019-10-10
tao_idl: robustness in run_test.pl

knutpett:
Fail the test if process could not be spawned.


jwillemsen:
Looks this script is doing way too much, see https://github.com/DOCGroup/ACE_TAO/blob/eb60837204ad2b40e73af2086713d90370cec218/TAO/tests/Portable_Interceptors/Recursive_ORBInitializer/run_test.pl for a simple run_test.pl that has 1 executable.

knutpett:
> Please take the other run_test.pl and make it like the others

Ok, new attempt coming up.

PR #971, opened 2019-10-10
Added NEWS for PR936

mitza-oci:


mitza-oci:
Merge at the same time as #967 to avoid starting CI builds twice

PR #972, opened 2019-10-17
fuzz.pl: Disable check_include by Default

iguessthislldo:
I mentioned this in [#967](https://github.com/DOCGroup/ACE_TAO/pull/967#issuecomment-539148218):

> So the fuzz has a check for #include <...>. I want to know if I can disable this check, because I don't think this convention is right since it forces the compiler to check the current directory for the header is definitely not there. fuzz.pl just says:
>
> > This check is suggested by Don Hinton to force user to use # " " instead of <> to avoid confict with Doxygen.
>
> What kind of conflict was it and will it still happen? Regardless this sounds like Doxygen was/is being configured incorrectly to me.

I know this check isn't necessarily harmful, but it's a restriction that seems to be arbitrary to me, especially since Doxygen should be just be documenting what the code, not dictating how we write it.

Finally I wanted to point out that this rule is not documented in `ACE-guidlines.html`.

### DIFF
--- a/ACE/bin/fuzz.pl
+++ b/ACE/bin/fuzz.pl
@@ -2447,7 +2447,7 @@ if (!getopts ('cdx:hl:t:s:mv') || $opt_h) {
            check_for_bad_ace_trace
            check_for_changelog_errors
            check_for_ptr_arith_t
-           check_for_include
+           check_for_include (disabled by default)
            check_for_non_bool_operators
            check_for_long_file_names
            check_for_refcountservantbase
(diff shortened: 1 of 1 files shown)


jwillemsen:
Have you tried what happens when you change the include? I think Doxygen ignores includes with <> for its include graph but not 100% sure. Given the fact that we have this fuzz for a long time already I think we should add a note to ACE_guidelines.html and just keep the fuzz unless there is a blocking problem with it

iguessthislldo:
I just tested this with a small Doxygen example I already had and it works fine:

Screenshot from 2019-10-19 20-33-26

All that seems to be necessary is that the include paths be in the `INPUT` list. It's only ignored If the file can not be found like `stdio.h` in my example.

Next week, I can try to change change some random files in ACE/TAO and see what Doxygen does.

PR #974, opened 2019-10-23
Expose Monotonic Time Support Status Macros

iguessthislldo:
Expose support status of `ACE_Monotonic_Time_Policy` using `ACE_HAS_MONOTONIC_TIME_POLICY` and support status of monotonic conditions using `ACE_HAS_MONOTONIC_CONDITIONS`.

Also refactored macros to try to simplify them while retaining compatibility and moved POSIX checks into `config-posix.h`.


ACE/ace/config-hpux-11.00.h
```diff
@@ -162,7 +162,6 @@
 #define ACE_HAS_CLOCK_SETTIME
 
 #define ACE_LACKS_CLOCK_MONOTONIC
-#define ACE_LACKS_MONOTONIC_TIME
```
iguessthislldo: I couldn't find a reference to this macro anywhere other than here and in the monictonic code. Under my changes, `ACE_LACKS_CLOCK_MONOTONIC` should have the same effect. Regardless I kept a check for it on the chance a user passes it.

jwillemsen: Why not just remove it and mention this in the NEWS file?

ACE/ace/Condition_Attributes.h
```diff
@@ -22,6 +22,18 @@
 
 #include "ace/OS_NS_Thread.h"
 
+#if defined (ACE_HAS_CONDATTR_SETCLOCK) && defined (ACE_HAS_CLOCK_GETTIME_MONOTONIC)
+#  if !defined (ACE_HAS_POSIX_MONOTONIC_CONDITIONS)
+#    define ACE_HAS_POSIX_MONOTONIC_CONDITIONS
```
jwillemsen: Shouldn't this be in config-posix.h? And the define below in a config-win32 file?

ACE/ace/config-all.h
```diff
@@ -82,6 +82,24 @@
...
+#  endif
+#  if defined (ACE_HAS_CLOCK_GETTIME_MONOTONIC)
+#    undef ACE_HAS_CLOCK_GETTIME_MONOTONIC
+#  endif
+#endif
+
+#if defined (ACE_LACKS_CONDATTR_SETCLOCK) || defined (ACE_LACKS_CONDATTR)
+#  if defined (ACE_HAS_CONDATTR_SETCLOCK)
+#    undef ACE_HAS_CONDATTR_SETCLOCK
+#  endif
```
jwillemsen: Is this not just an user error, why add an undef for that?

ACE/ace/Monotonic_Time_Policy.h
```diff
@@ -16,6 +16,12 @@
 # pragma once
 #endif /* ACE_LACKS_PRAGMA_ONCE */
 
+#if defined (ACE_WIN32) || defined (ACE_HAS_CLOCK_GETTIME_MONOTONIC)
```
jwillemsen: Shouldn't the part for ACE_WIN32 be in the config-win32-common.h and the other part in another config file?

ACE/ace/OS_NS_Thread.inl
```diff
@@ -218,31 +218,18 @@ ACE_OS::condattr_synctype (ACE_condattr_t &attributes, int& type)
 ACE_INLINE int
 ACE_OS::condattr_setclock (ACE_condattr_t &attributes, clockid_t clock_id)
 {
-# if defined (ACE_HAS_THREADS)
```
jwillemsen: Why remove the check for ACE_HAS_THREADS? We can build ACE without thread support

ACE/ace/config-android.h
```diff
@@ -118,6 +118,15 @@
 #  define ACE_LACKS_WCSTOLL
 #  define ACE_LACKS_WCSTOULL
 #  define ACE_LACKS_CONDATTR_SETCLOCK
+#  if (defined ACE_HAS_CONDATTR_SETCLOCK)
```
jwillemsen: Why undef them here and in some other config files, how can it be that these are already defined?

iguessthislldo: Because in all these cases these are after `config-posix.h`, which could be defining them.

ACE/ace/OS_NS_Thread.inl
iguessthislldo: I know that, this implementation is only active if `ACE_CONDATTR_SETCLOCK` is defined in `config-posix.h`, which in turn is ultimately defined if `ACE_HAS_THREADS` is defined.

ACE/ace/Monotonic_Time_Policy.h
iguessthislldo: I'm not sure about doing that since this is specific to `Monotonic_Time_Policy`. I can you insist that's right.

ACE/ace/config-android.h
jwillemsen: Why would config-posix.h define these? I would assume the underlying defines are not set, or are they set incorrectly by the os headers?

iguessthislldo: https://github.com/DOCGroup/ACE_TAO/blob/dcbb7778f1c86c562efbf4e87f404a11592f3a48/ACE/ace/config-linux.h#L147-L155

jwillemsen: That comment/code is specifically for some linux kernel revisions, I am not sure if this is valid for android/hpux, I wouldn't expect that those set the incorrect posix macros and whether we need to do that for those also.

ACE/ace/Monotonic_Time_Policy.h
mcorino: This block could be moved to config-macros.h (preferably) or alternatively to Monotonic_Time_Policy.inl.

ACE/ace/config-android.h
mcorino: Again the common block could be moved to config-macros.h with an "#if defined (ACE_LACKS_CONDATTR_SETCLOCK)"/"#endif" bracket.

iguessthislldo: I originally had the common code to handle this in `config-all.h`, but `config-macros.h` looks much more appropriate for this.

ACE/ace/Monotonic_Time_Policy.h
iguessthislldo: Putting it in `Monotonic_Time_Policy.inl` is the opposite of what I want because the entire point of this is to expose the support status. However since you both seem to agree it shouldn't go here, I will break it up like @jwillemsen  suggested and put 

```
#define ACE_HAS_MONOTONIC_TIME_POLICY
```

in `config-win32-common.h` and

```
#if  ACE_HAS_CLOCK_GETTIME_MONOTONIC
#  define ACE_HAS_MONOTONIC_TIME_POLICY
#endif
```

in `config-macros.h` like @mcorino suggested.

iguessthislldo:
@jwillemsen So azure looks like it has the changes from #987, but it's still using 16.04 xenial by the looks of the `apt update` it's doing, so it's not going to have clang 7 and clang 8.

jwillemsen:
Maybe try to merge upstream master into your branch and see whether azure takes 18.04

iguessthislldo:
> Maybe try to merge upstream master into your branch and see whether azure takes 18.04

It did the same thing. It's starting to sound like a problem on azure's end...

jwillemsen:
> It did the same thing. It's starting to sound like a problem on azure's end...

Yes, looks a problem at the azure end, very strange why it uses xenial

PR #988, opened 2019-11-06
Add access control argument for ACE_Configuration_Win32Registry

likema:


ACE/ace/Configuration.h
```diff
@@ -520,7 +521,8 @@ class ACE_Export ACE_Configuration_Win32Registry : public ACE_Configuration
    */
   static HKEY resolve_key (HKEY hKey,
                            const ACE_TCHAR* path,
-                           int create = 1);
+                           int create = 1,
```
jwillemsen: Can you maybe at the same time change create to a bool? From very very long ago we didn't use bool because not all compilers did support it, but now we should use it

ACE/ace/Configuration.h
```diff
@@ -533,6 +535,8 @@ class ACE_Export ACE_Configuration_Win32Registry : public ACE_Configuration
   ACE_Configuration_Win32Registry (void);
   ACE_Configuration_Win32Registry (const ACE_Configuration_Win32Registry& rhs);
   ACE_Configuration_Win32Registry& operator= (const ACE_Configuration_Win32Registry& rhs);
+
+  u_long security_access_;
```
jwillemsen: Can this member be const?

likema: Done

ACE/ace/Configuration.h
likema: Only this one or all `create` arg in these two files?

jwillemsen: Please update all create arguments in these two files

likema: Updated

PR #991, opened 2019-11-11
Use additional repo for clang7/8

jwillemsen:
* azure-pipelines.yml:


iguessthislldo:
What if we tried specifically saying `ubuntu-18.04` instead of `ubuntu-latest`?

PR #995, opened 2019-11-13
Use unspecified_bool_type instead of bool.

likema:
For better compatibility, use unspecified_bool_type instead of bool in `ACE_Refcounted_Auto_Ptr`, which is inspired by `ACE_Intrusive_Auto_Ptr`.


jwillemsen:
Please extend one of the ACE unit tests to demonstrate and test this usage

likema:
It is already there https://github.com/DOCGroup/ACE_TAO/blob/master/ACE/tests/Refcounted_Auto_Ptr_Test.cpp#L405

jwillemsen:
Created PR #1127, the change doesn't compile with vc7.1, reverting this for the moment.

jwillemsen:
Looks vc71 had a problem in the past, see https://github.com/DOCGroup/ACE_TAO/commit/5de88ed8cb5f4ebbb4bc689e230e587664f19798

jwillemsen:
When you make a new PR please add a comment on both places that the inline code is required for vc71, that way we know that for the future and we can cleanup that code at the moment we deprecate 7.1

PR #998, opened 2019-11-18
Use POSIX-style Macros for Monotonic Support

iguessthislldo:
Tweak to #974 that I'm hoping to get this into the release.

After starting to work on https://github.com/objectcomputing/OpenDDS/pull/1363, I realized it would be better for forward compatibility and cross compatibility with OCI ACE/TAO if the monotonic time support macros were like [POSIX macros](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/unistd.h.html). These macros would always be defined and the value of the macro says if the feature is supported. This way if the macros are not defined, the application can definitively know that ACE doesn't have them at all and fall back to an alternative method for determining if monotonic time is supported.

Also took @mitza-oci's suggestion to get rid of the "HAS" part of the macros.


jwillemsen:
I am also not in favor of adopting a different style for these macros, would propose to leave it as is

iguessthislldo:
> It changes behavior from the way most ACE feature macros function

That is the point. I don't like the way ACE does macros, it's too chaotic, especially the fact that there are both "LACKS" and "HAS" macros.

> it also definitely does not simplify the feature tests (to say the least).

That's true. I could actually get rid of the define checks for the tests since it would always have them so something like this:

```
#if ACE_MONOTONIC_TIME_POLICY >= 0 && ACE_MONOTONIC_CONDITIONS >= 0
```

shuston:
I agree with Martin and Johnny. Although the proposed macro style is "POSIX-ish", it is inconsistent with the rest of the ACE feature-test macros. IMO, consistency wins unless something is really broken.

iguessthislldo:
I had no idea this would be so controversial, especially since some macros like `ACE_HAS_THREADS` have similar behavior. I can see that I'm being overruled though. Closing...

PR #1001, opened 2019-11-20
Prevent crash

knutpett:
Discovered that if there are no scopes, then bottom() will return nullptr. This was discovered in my own tests where the AST was constructed programatically and not by the frontend.


jwillemsen:
Thanks for your review @iguessthislldo and @knutpett for the proposed PR

PR #1004, opened 2019-12-06
Fixed typo in config-posix.h

knutpett:
I suspect this was a typo.

### DIFF
--- a/ACE/ace/config-posix.h
+++ b/ACE/ace/config-posix.h
@@ -58,7 +58,7 @@
 #       define ACE_HAS_PTHREADS
 #     endif /* ACE_HAS_PTHREADS */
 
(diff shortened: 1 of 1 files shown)


mitza-oci:
> I'm wondering if there is a flag for warning when evaluating a macro that is undefined. It might be useful given how many macros ACE deals with.

-Wundef
https://gcc.gnu.org/onlinedocs/gcc-9.2.0/cpp/Invocation.html#Invocation

knutpett:
> > I'm wondering if there is a flag for warning when evaluating a macro that is undefined. It might be useful given how many macros ACE deals with.
> 
> -Wundef
> https://gcc.gnu.org/onlinedocs/gcc-9.2.0/cpp/Invocation.html#Invocation

That's how I found it. (Building all of ACE+TAO a lot more is reported)

PR #1013, opened 2019-12-29
Add ACE_GCC_FORMAT_ATTRIBUTE to ACE_Process_Options

likema:
Decorate `ACE_Process_Options::command_line` and `ACE_Process_Options::setenv` by `ACE_GCC_FORMAT_ATTRIBUTE` for GCC warning incorrect `printf` format.

Without `ACE_GCC_FORMAT_ATTRIBUTE`, the following snippet would neither be effective in run time nor warning in compile time.

```cpp
unsigned v = 1;
...
opts.set_env("KEY", "%s", v);
```

### DIFF
--- a/ACE/ace/Process.h
+++ b/ACE/ace/Process.h
@@ -118,7 +118,11 @@ public:
   /// @param format must be of the form "VARIABLE=VALUE".  There can not be
   /// any spaces between VARIABLE and the equal sign.
   int setenv (const ACE_TCHAR *format,
-              ...);
+              ...)
+#if !defined (ACE_USES_WCHAR)
+    ACE_GCC_FORMAT_ATTRIBUTE (printf, 2, 3)
+#endif /* !ACE_USES_WCHAR */
+    ;
 
   /**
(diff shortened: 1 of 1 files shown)


iguessthislldo:
This caused [an issue with this wchar build](https://travis-ci.com/DOCGroup/ACE_TAO/jobs/271043225). It looks like it's caused by [a long lived bug with how GCC treats the format attribute on different string types](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=64862). Changing `ACE_GCC_FORMAT_ATTRIBUTE` to be a NOP on wchar builds may be a possible solution.

likema:
Fixed it.

> This caused [an issue with this wchar build](https://travis-ci.com/DOCGroup/ACE_TAO/jobs/271043225). It looks like it's caused by [a long lived bug with how GCC treats the format attribute on different string types](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=64862). Changing `ACE_GCC_FORMAT_ATTRIBUTE` to be a NOP on wchar builds may be a possible solution.

PR #1014, opened 2020-01-01
Rewrote auto_run_tests.pl Arg Parsing

iguessthislldo:
I've been wanting to do this since at least #943, because I remember being confused by some of this stuff then.

- Help message:
  - It said test list options, like `-l` or `-a`, "only" do the tests from that option. That isn't true, they include the tests in the run in addition to other ones.
  - Removed reference to non-existent `-c` option which is supposed to be an alias to `-Config`.
  - Removed outdated synopsis and try to give a slightly more detailed explanation of the script.
  - Documented `-Exclude`
  - Moved configuration listing to its own option and made it do it for all test list files that would ran otherwise.
- Argument Parsing:
  - Switched to using `GetOpt::Long` and added long options.
  - Fixed dry run `-z` option which wasn't being parsed.
  - Fixed that when using `-l` to specify multiple files it only uses the last one. Also made it so it takes non-option arguments as test lists.
- Documented what arguments are being parsed by `PerlACE` modules.
- Added 'use strict' and fixes for that.
- Removed seemingly unused `use English;`.

### DIFF
--- a/ACE/bin/auto_run_tests.pl
+++ b/ACE/bin/auto_run_tests.pl
@@ -3,8 +3,10 @@ eval '(exit $?0)' && eval 'exec perl -S $0 ${1+"$@"}'
     if 0;
 
 # -*- perl -*-
-# This file is for running the run_test.pl scripts listed in
-# auto_run_tests.lst.
+# Executes test list files (*.lst), which contain commands with conditions
+# called configurations under which the commands are run.
+
+use strict;
 
 use lib "$ENV{ACE_ROOT}/bin";
 if (defined $ENV{srcdir}) {
@@ -12,8 +14,7 @@ if (defined $ENV{srcdir}) {
 }
 use PerlACE::Run_Test;
 
-use English;
-use Getopt::Std;
+use Getopt::Long;
 use Cwd;
 
 use Env qw(ACE_ROOT PATH TAO_ROOT CIAO_ROOT DANCE_ROOT DDS_ROOT);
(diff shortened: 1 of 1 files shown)


ACE/bin/auto_run_tests.pl
```diff
@@ -31,20 +32,24 @@
...
+      my $coredump = $? & 128;
       my $error_message;
       if ($? == -1) {
         $error_message = "failed to run: $!";
       }
-      elsif ($? & 127) {
-        $error_message = sprintf("exited on signal %d", ($? & 127));
-        $error_message .= " and created coredump" if ($? & 128);
+      elsif ($signal) {
+        $error_message = sprintf ("exited on signal %d", ($signal));
```
mitza-oci:
```suggestion
        $error_message = sprintf("exited on signal %d", $signal);
```
Not sure how strictly we care about Perl's official style (in the perlstyle man page), but ACE-guidelines says we should use it.  (There are a few more like this if you decide to change it.)

jwillemsen:
@iguessthislldo  I think there is a problem with the test framework, very likely triggered by this PR, see https://buildlogs.remedy.nl/win_msvc15_opendds_debug/

iguessthislldo:
I will look into it.

mitza-oci:
> 
> 
> @iguessthislldo I think there is a problem with the test framework, very likely triggered by this PR, see https://buildlogs.remedy.nl/win_msvc15_opendds_debug/

It wasn't related.  See https://github.com/objectcomputing/OpenDDS/pull/1437 and https://github.com/objectcomputing/OpenDDS/pull/1456

PR #1022, opened 2020-01-15
Rename VERSION Files and Add Instructions for Building for Android on Windows

iguessthislldo:
- Added instructions to `ACE-INSTALL.html` for building ACE/TAO for Android on Windows.
- Renamed `VERSION` file to `VERSION.txt` to avoid conflicting with the `version` standard header.


iguessthislldo:
Building on OpenDDS-Android [here](https://travis-ci.org/iguessthislldo/OpenDDS-Android/builds/637287510)

ACE/include/makeinclude/platform_linux_common.GNU
```diff
@@ -1,15 +1,18 @@
...
 
-# We always include config-linux.h on Linux platforms.
+# Include config-linux.h by default on Linux platforms.
 ACE_PLATFORM_CONFIG ?= config-linux.h
 
 debug ?= 1
 optimize ?= 1
 threads ?= 1
 insure ?= 0
+linux_is_cross_compiled ?= 0
```
jwillemsen: Will this have any impact when using gcc with cross compilation to ARM?

iguessthislldo: I just realized I have the checks for this backwards down below, so I will have to fix that, but no I don't think so. I'm only trying to make it act different for Android at the moment.

iguessthislldo: Actually, after playing around with it a little more, I think my changes to this file and `platform_android.GNU` are unnecessary and I will reverse them.

ACE/include/makeinclude/platform_clang_common.GNU
```diff
@@ -76,7 +76,7 @@ ifneq ($(DLD),)
   endif # DLD = CXX_FOR_VERSION_TEST
   # The -E option is GNU ld specific
   ifneq ($(mingw32),1)
-    GNU_LD := $(shell sh -c '$(LD_FOR_VERSION_TEST) -v 2>&1 | grep -c "GNU ld"')
+    GNU_LD := $(shell $(LD_FOR_VERSION_TEST) -v 2>&1 | grep -c 'GNU ld')
```
mitza-oci: At this point should we remove this conditional from the makefile?  The `else` side should work on all platforms, unless we really need the stderr content?

ACE/ACE-INSTALL.html
```diff
@@ -3102,8 +3187,10 @@ <h2>To create an Eclipse project for ACE from a local source distribution:</h2>
 
 </pre>
 
-</p><hr><p>
-</p><h2><a name="power">Building from git</a></h2>
+</p><hr>
+
+<h2><a name="buildfiles">Generating Build Files</a></h2>
```
mitza-oci: Is there a more specific term than "build files" we could use for this?

ACE/ACE-INSTALL.html
```diff
@@ -1623,20 +1626,53 @@ <h3><a href="mailto:Jaffar_Shaikh@Mitel.COM">Jaffar Shaikh's</a>
...
+virtualized Linux environment like Windows Subsystem for Linux (WSL), Docker,
+or a traditional VM, then you can use the Linux version of the NDK and ignore
+rest of this note and all the other Windows specific notes.
+</p>
+
+<p>If that is not the case, you should also pay attention to the notes marked
+with "<b>Windows Users:</b>" in addition to the rest of the instructions. In
+addition to the Windows version of the Android NDK, you will also need MinGW
+for Unix utilities that ACE needs. Install <a href="http://www.mingw.org">
+MinGW</a> and the <code>mingw-developer-toolkit-bin</code> meta package using
```
mitza-oci: It would be good to find out what the minimal subset is, since `mingw-developer-toolkit-bin` brings a lot of dependencies in (autotools?).  I don't think we have to do it in this PR though.

ACE/ACE-INSTALL.html
```diff
@@ -1623,20 +1626,53 @@ <h3><a href="mailto:Jaffar_Shaikh@Mitel.COM">Jaffar Shaikh's</a>
...
+the MinGW package manager.
+
+<p>You will also have to delete the <code>%ACE_ROOT%\VERSION</code> and
```
mitza-oci:
```suggestion
<p>You will also have to rename the <code>%ACE_ROOT%\VERSION</code> and
```

ACE/ACE-INSTALL.html
iguessthislldo: True, but I didn't want to go through them at the time and I didn't know everything that gnuace would want.

mitza-oci: Run procmon while it builds and then use the process tree when it's done

ACE/ACE-INSTALL.html
```diff
@@ -2507,7 +2576,7 @@ <h1><a name="installnotes">Installation Notes</a></h1>
       When building without network support you will need the ace_for_tao
       subsetting functionality enabled. For this add <code>ace_for_tao = 1</code>
       to your <code>bin/MakeProjectCreator/config</code> file and regenerate the
-      GNU makefiles as described <a href="#power">here</a>.</p><p>
+      GNU makefiles as described <a href="#buildfiles">here</a>.</p><p>
```
jwillemsen: Where is the buildfiles section?

jwillemsen:
Maybe we should remove the VERSION files completely, looks only doxygen generation uses it, it could use ace/Version.h as alternative

ACE/ACE-INSTALL.html
mitza-oci:
```suggestion
      GNU makefiles as described <a href="#power">here</a>.</p><p>
```

iguessthislldo: I renamed it but decided to revert it because it could break a link. I just missed this one.

iguessthislldo:
> Maybe we should remove the VERSION files completely, looks only doxygen generation uses it, it could use ace/Version.h as alternative

I would be better if we could eliminate the problem all together and make it so it's not possible to include files like this, but I don't know how to do that without putting an extra directory between the project root and the headers.

PR #1023, opened 2020-01-16
ACE_OS::rwlock_init - fixed bug when RW Lock is emulated and creating a condition variable fails

mitza-oci:
Need to avoid calling cond_destroy on a garbage struct.

Problem was seen on Windows when process doesn't have permissions to create named kernel objects.

Also ACE's emulated condition variables (USYNC_THREAD) shouldn't be creating named kernel objects -- not fixed in this PR.

### DIFF
--- a/ACE/ace/OS_NS_Thread.cpp
+++ b/ACE/ace/OS_NS_Thread.cpp
@@ -3132,79 +3132,99 @@ ACE_OS::lwp_setparams (const ACE_Sched_Params &sched_params)
 #endif /* ! ACE_HAS_STHREADS && ! sun */
 }
 
+#if defined ACE_HAS_THREADS && defined ACE_LACKS_RWLOCK_T
+namespace {
+struct UniqueName {
+  explicit UniqueName (const void *addr)
+  {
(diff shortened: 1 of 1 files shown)


ACE/ace/OS_NS_Thread.cpp
```diff
@@ -3132,79 +3132,99 @@ ACE_OS::lwp_setparams (const ACE_Sched_Params &sched_params)
...
+
+  ~RWLockCleaner ()
+  {
+    ACE_Errno_Guard error (errno);
+    switch (this->state_)
+      {
+      case RWLC_Cond2:
+        ACE_OS::cond_destroy (&this->rw_->waiting_writers_);
+        // FALLTHROUGH
+      case RWLC_Cond1:
```
jwillemsen: Instead of 1/2 maybe using Writer/Reader?

jwillemsen:
New compile error, see http://scoreboard.ociweb.com/doc_master_hornet_win7-x64_vc14_i1d0o1w1/2020_01_18_02_54_Brief.html#section_4

PR #1039, opened 2020-02-04
Configuration changes for iOS builds.

jonesc-oci:
Configuration file changes for ACE for iOS Xcode 11 builds.


ACE/ace/config-macosx-leopard.h
```diff
@@ -204,11 +204,17 @@
...
 #define ACE_LACKS_CONDATTR_SETCLOCK
+
+#ifndef ACE_HAS_IPHONE
 #if __MAC_OS_X_VERSION_MAX_ALLOWED < 101200
 #define ACE_LACKS_CLOCKID_T
 #define ACE_LACKS_CLOCK_MONOTONIC
 #define ACE_LACKS_CLOCK_REALTIME
 #endif
+#else
+#define ACE_LACKS_SYSTEM
```
iguessthislldo: Shouldn't this go in an iOS header if it's iOS specific?

ACE/ace/config-macosx-iOS-hardware.h
```diff
@@ -3,12 +3,12 @@
...
+#define ACE_LACKS_SYSTEM
 
-#include "ace/config-macosx-mavericks.h"
+#include "ace/config-macosx-mojave.h"
 
 #ifdef ACE_HAS_SYSV_IPC
 #undef ACE_HAS_SYSV_IPC
 #endif
 
 #endif /* ACE_CONFIG_MACOSX_IPHONE_HARDWARE_H */
```
mitza-oci: Do "simulator" and "hardware" actually need separate config files?  Seems like a simulator should be able to run the same code as the hardware.  The only differences are `ACE_SIZEOF_LONG_DOUBLE` and `ACE_HAS_SYSV_IPC`.

Issue #879, opened 2019-04-01
During demarshaling CORBA sequence type indirection is not supported


xor2003:
### Version
TAO v2.0a (but applicable to all versions)
### Host machine and operating system
RHEL 6
### Target machine and operating system (if different from host)
?
### Compiler name and version (including patch level)
gcc 4.4.7
### The $ACE_ROOT/ace/config.h file
#define ACE_HAS_IPV6 1
#define ACE_USES_IPV4_IPV6_MIGRATION 1
#include "ace/config-linux.h"

### The $ACE_ROOT/include/makeinclude/platform_macros.GNU file
buildbits=64
threads=1
ssl=1
install_rpath=0
debug=1
optimize=0
versioned_so=0
INSTALL_PREFIX = /home/xor/tao_build/tao_install_linux
include /home/xor/tao_build/tao_build_linux/ACE_wrappers/include/makeinclude/platform_linux.GNU
### The problem effects:
TAO execution
### Synopsis
TAO fails to demarshal OMG event when CORBA Sequence type provided with indirection (-1 code).
### Description
TAO received OMG event. It's remainder_of_body field is Any type.
Which is struct. One of the field (offset 1716) has tc_kind=-1 which is indirection pointing to sequence (offset -416).
From the source code ACE_wrappers_2.0\TAO\tao\AnyTypeCode\TypeCode_CDR_Extraction.cpp:1299
tc_demarshal_indirection()
        // Only struct, union, event and valuetype TypeCodes may be
        // recursive. However non-recursive ALIAS may be indirected.
### Repeat by
Attached
### Sample fix/ workaround
Not available

xor2003:
CORBA standard
"General Inter-ORB Protocol"
Chapter "Indirection: Recursive and Repeated TypeCodes"
does not have such limitation

jwillemsen:
Please create a pull request with a test extension for one of the existing unit test under [TAO/tests/Any](https://github.com/DOCGroup/ACE_TAO/tree/master/TAO/tests/Any) and it would help when you could also propose a fix.

xor2003:
Add pull request with test.
How Jacorb handle it:
FINE     read TypeCode kind 2 at startposition 1700
FINE     return {TypeCode: Kind=2 (tk_short), ID=null, recursive=false} (org.jacorb.orb.TypeCode@398918169)
FINE     read TypeCode kind -1 at startposition 1716
FINE       startPosition: 1716
FINE       negative offset: -416
[...]

PR #1047, opened 2020-02-27
Sequence type indirection support

xor2003:
There is no limitation in CORBA standard for not supporting indirection for sequence type.
Some unknown our customer's ORB sends indirected sequence type.
I have rebased and squashed previous patch for the same.
This code is already working in production.
It is related to issue:
During demarshaling CORBA sequence type indirection is not supported #879


TAO/tao/AnyTypeCode/TypeCode_CDR_Extraction.cpp
```diff
@@ -1399,6 +1400,17 @@ namespace
         return false;
       }
 
+    if (CORBA::tk_sequence == kind || CORBA::tk_array == kind)  // @todo 1. check if recursion is properly handled 2. check need for other complex types
+      {
+        using namespace TAO::TypeCodeFactory;
```
jwillemsen: Don't use `using`

xor2003: Fixed it

PR #1069, opened 2020-04-07
IPv6 and RECVINFO on macOS; Windows if_nametoindex

jrw972:


ACE/ace/config-macosx-mojave.h
```diff
@@ -1,6 +1,8 @@
 #ifndef ACE_CONFIG_MACOSX_MOJAVE_H
 #define ACE_CONFIG_MACOSX_MOJAVE_H
 
+#define __APPLE_USE_RFC_3542
```
mitza-oci, line 4: should this be guarded with an `ifdef` in case its already defined?

ACE/ace/OS_NS_sys_socket.h
```diff
@@ -65,9 +65,7 @@
 #endif
 
 #if defined (ACE_HAS_IPV6)
-#if defined (IPV6_PKTINFO)
-#define ACE_RECVPKTINFO6 IPV6_PKTINFO
-#elif defined (IPV6_RECVPKTINFO)
+#if defined (IPV6_RECVPKTINFO)
```
jwillemsen, line 68: Why remove the lines above?

jrw972: They incorrectly define `ACE_RECVPKTINFO6` to `IPV6_PKTINFO` which is a tag used to identify the packet info as opposed to a socket option.

ACE/ace/config-macosx-mojave.h
jwillemsen: and some docu maybe why this define is set?

ACE/ace/OS_NS_sys_socket.h
```diff
@@ -276,7 +274,7 @@ namespace ACE_OS
   int shutdown (ACE_HANDLE handle,
                 int how);
 
-#if defined (ACE_LINUX) && defined (ACE_HAS_IPV6)
+#if (defined (ACE_LINUX) || defined(ACE_HAS_MAC_OSX)) && defined (ACE_HAS_IPV6)
   ACE_NAMESPACE_INLINE_FUNCTION
   unsigned int if_nametoindex (const char *ifname);
```
jwillemsen, line 279: In the implementation of this operation already ACE_LACKS_IF_NAME_INDEX (which I think we should rename to ACE_LACKS_IF_NAMETOINDEX) is used, so why not only compare ACE_HAS_IPV6? Looks also Windows has this operation in newer windows releases.

ACE/ace/SOCK_Dgram.cpp
```diff
@@ -729,7 +729,7 @@ ACE_SOCK_Dgram::make_multicast_ifaddr6 (ipv6_mreq *ret_mreq,
                   0,
                   sizeof (lmreq));
 
-#if defined(ACE_LINUX)
+#if defined(ACE_LINUX) || defined(ACE_HAS_MAC_OSX)
```
jwillemsen, line 732: Here we could then use the renamed ACE_LACKS_IF_NAMETOINDEX

mitza-oci:
There are build issues on Azure Pipelines CI.  Once those are resolved we are probably ready, unless @jwillemsen has any other comments.

jwillemsen:
Looks Visual Studio 2017 is failing with a real error, not related to CI

jwillemsen:
@mitza-oci by accident I also triggered a new review by you

mitza-oci:
Windows doesn't make these functions available for all versions.  ACE's default is
```
# define _WIN32_WINNT 0x0501
```
(in config-win32-common.h)
If we had at least `0x0600` here, we could use `if_nametoindex` and related functions.  Doing so may require some extra `#include`s, but that's the easy part.

There are a few ways we could address this:
- update config-win32-common to move up to a more recent default version
- make or use ACE_HAS/ACE_LACKS macros to make ACE's use of these functions conditional on Windows version
- other?

jwillemsen:
@mitza-oci Personally I would be in favor of doing both actions, based on _WIN32_WINNT enable usage of if_nametoindex and related functions on windows and when that works update _WIN32_WINNT in ACE to something newer which we document in the NEWS file.

mitza-oci:
There will still be changes to make to ACE to compile with the Windows API headers.
See documentation: https://docs.microsoft.com/en-us/windows/win32/api/netioapi
ACE assumes (if `!ACE_LACKS_IF_NAME_INDEX`) that there are functions `if_nameindex` and `if_freenameindex`.  These aren't available on Windows, which has the functions with `to` in their names.

mitza-oci:
Aside from the issues I posted about before, the only other change I made locally was to `#include` the header for IP helper API:

```
diff --git a/ACE/ace/os_include/net/os_if.h b/ACE/ace/os_include/net/os_if.h
index e3cb6842340..a0d815e2165 100644
--- a/ACE/ace/os_include/net/os_if.h
+++ b/ACE/ace/os_include/net/os_if.h
@@ -40,6 +40,7 @@

 #if defined (ACE_HAS_WINSOCK2) && (ACE_HAS_WINSOCK2 != 0)
 #  include /**/ <ws2tcpip.h>
+#  include /**/ <iphlpapi.h>
 #endif /* ACE_HAS_WINSOCK2 */

 // Place all additions (especially function declarations) within extern "C" {}
```

ACE/NEWS
```diff
@@ -1,6 +1,8 @@
 USER VISIBLE CHANGES BETWEEN ACE-6.5.8 and ACE-6.5.9
 ====================================================
 
+. _WIN32_WINNT updated to 0x0600 from 0x0501.
```
jwillemsen, line 4: We should add some more information about the implications for the user for this

mitza-oci: I reconsidered this one...
Since we are changing this in config-win32-common, this seems like a good time to either:
- not change _WIN32_WINNT at all, this should be up to the user
- not reduce _WIN32_WINNT if the user has already set it to a larger value
- possibly emit a warning if _WIN32_WINNT is too low
In any case, it seems like this is a good time to make ACE play better with other libs/configurations

ACE/ace/config-win32-common.h
```diff
@@ -55,8 +55,8 @@
 #  endif  /* !_FILE_OFFSET_BITS */
 #endif /* _WIN64 || WIN64 */
 
-#if !defined (_WIN32_WINNT)
-# define _WIN32_WINNT 0x0501 // pretend it's at least Windows XP or Win2003
+#if !defined (_WIN32_WINNT) || _WIN32_WINNT < 0x0600
+#  error _WIN32_WINNT must be at least 0x0600
```
mitza-oci, line 59: According to @jwillemsen's comment should this line be changed to
`# define ACE_LACKS_IF_NAME_INDEX`
If were doing something like that, this block would have to move to after the `#include` of windows headers to pick up the default value of _WIN32_WINNT

jwillemsen:
Looks one travis-ci build is failing due to a redefinition

mitza-oci:
> Ok, only the travis-ci build has to be fixed

GitHub Actions now seems have an issue; but the actual bug seen in Travis is fixed

mitza-oci:
@jwillemsen is GitHub Actions running in to space limitations?
```
.obj/Main.o: file not recognized: File truncated
collect2: error: ld returned 1 exit status
make[1]: *** [Aio_Platform_Test] Error 1
```

jwillemsen:
The action worked a few hours ago, did restart it just now

mitza-oci:
> The action worked a few hours ago, did restart it just now

I've done about 15 restarts today

jwillemsen:
Looks one windows build from OCI has an issue, see http://scoreboard.ociweb.com/doc_master_weevil_win2003_vc71_s1d1o0/index.html

```
[Details] ..\ace\OS_NS_Thread.inl(3756) : error C2039: 'TryEnterCriticalSection' : is not a member of 'operator``global namespace'''
[Details] ..\ace\OS_NS_Thread.inl(3756) : error C3861: 'TryEnterCriticalSection': identifier not found, even with argument-dependent lookup
[Details] OS_NS_Thread.cpp(1331) : error C2039: 'SignalObjectAndWait' : is not a member of 'operator``global namespace'''
[Details] OS_NS_Thread.cpp(1331) : error C3861: 'SignalObjectAndWait': identifier not found, even with argument-dependent lookup
[Details] CDR_Base.cpp(813) : warning C4244: 'initializing' : conversion from 'ACE_CDR::ULongLong' to 'const unsigned int', possible loss of data
```

mitza-oci:
> ```
> [Details] ..\ace\OS_NS_Thread.inl(3756) : error C2039: 'TryEnterCriticalSection' : is not a member of 'operator``global namespace'''
> [Details] ..\ace\OS_NS_Thread.inl(3756) : error C3861: 'TryEnterCriticalSection': identifier not found, even with argument-dependent lookup
> [Details] OS_NS_Thread.cpp(1331) : error C2039: 'SignalObjectAndWait' : is not a member of 'operator``global namespace'''
> [Details] OS_NS_Thread.cpp(1331) : error C3861: 'SignalObjectAndWait': identifier not found, even with argument-dependent lookup
> ```

I was just looking at that.  It's an old unsupported (by Microsoft) version of Windows and Visual Studio.  May be easier just to turn it off.

mitza-oci:
> 
> 
> > ```
> > [Details] ..\ace\OS_NS_Thread.inl(3756) : error C2039: 'TryEnterCriticalSection' : is not a member of 'operator``global namespace'''
> > [Details] ..\ace\OS_NS_Thread.inl(3756) : error C3861: 'TryEnterCriticalSection': identifier not found, even with argument-dependent lookup
> > [Details] OS_NS_Thread.cpp(1331) : error C2039: 'SignalObjectAndWait' : is not a member of 'operator``global namespace'''
> > [Details] OS_NS_Thread.cpp(1331) : error C3861: 'SignalObjectAndWait': identifier not found, even with argument-dependent lookup
> > ```
> 
> I was just looking at that. It's an old unsupported (by Microsoft) version of Windows and Visual Studio. May be easier just to turn it off.

See #1079

Issue #1068, opened 2020-04-06
ORBListenEndpoints SSLIOP port span support


wkbrd:
### Version

6.5.5

### Host machine and operating system

Windows and Linux

#### Problem description

Is there a way to specify an ORBListenEndpoints string to have an SSLIOP port span and if so, what is it?  From our review, it does not appear to be supported

When providing an IIOP port and an SSLIOP port as the value of ORBListenEndpoints (e.g., iiop://:10100/ssl_port=10500/portspan=100), it appears that portspan is not being honored by the SSL implementation.  Attempts to connect to a port once the first in the SSL span is already in use fails with CORBA::BAD_PARAM (on Linux; equivalent behavior on Windows with /reuse_addr=0). 

IIOP_Connector uses the port span:
https://github.com/DOCGroup/ACE_TAO/blob/master/TAO/tao/IIOP_Connector.cpp (line 388)

However nothing in the SSLIOP connector classes seem to:
- https://github.com/DOCGroup/ACE_TAO/blob/master/TAO/orbsvcs/orbsvcs/SSLIOP/SSLIOP_Connector.cpp
- https://github.com/DOCGroup/ACE_TAO/blob/master/TAO/orbsvcs/orbsvcs/SSLIOP/IIOP_SSL_Connector.cpp

Uses in the Acceptor classes:
- https://github.com/DOCGroup/ACE_TAO/blob/master/TAO/tao/IIOP_Acceptor.cpp (line 30, 481, 1243)
- https://github.com/DOCGroup/ACE_TAO/blob/master/TAO/orbsvcs/orbsvcs/SSLIOP/SSLIOP_Acceptor.cpp (line 488)

Initialization of the port span:

- https://github.com/DOCGroup/ACE_TAO/blob/master/TAO/tao/params.h (line 106-110, 339-341)
- https://github.com/DOCGroup/ACE_TAO/blob/master/TAO/tao/params.cpp (line 31)
- https://github.com/DOCGroup/ACE_TAO/blob/master/TAO/tao/params.inl (line 223-234)
- https://github.com/DOCGroup/ACE_TAO/blob/master/TAO/tao/ORB_Core.cpp (line 670)

PR #1070, opened 2020-04-07
Implementation of portspan capability for SSLIOP

wkbrd:
It was determined that SSLIOP was not implemented per #1068. 

Please review and consider adding support for this capability that is valuable for implementing SSLIOP in conjunction with port-based firewall rules.

Thanks!


wkbrd:
@jwillemsen , without this code addition, the portspan was not recognized by the SSLIOP implementation.

jwillemsen:
Iiop acceptor is the base, would expect it to be handled there

Issue #914, opened 2019-05-28
Any plans to support musl libc?


james-aurizn:
### Version

6.5.5

### Host machine and operating system

Alpine linux docker container.

### Compiler name and version (including patch level)

```
# g++ -v
Using built-in specs.
COLLECT_GCC=g++
COLLECT_LTO_WRAPPER=/usr/libexec/gcc/x86_64-alpine-linux-musl/8.3.0/lto-wrapper
Target: x86_64-alpine-linux-musl
Configured with: /home/buildozer/aports/main/gcc/src/gcc-8.3.0/configure --prefix=/usr --mandir=/usr/share/man --infodir=/usr/share/info --build=x86_64-alpine-linux-musl --host=x86_64-alpine-linux-musl --target=x86_64-alpine-linux-musl --with-pkgversion='Alpine 8.3.0' --enable-checking=release --disable-fixed-point --disable-libstdcxx-pch --disable-multilib --disable-nls --disable-werror --disable-symvers --enable-__cxa_atexit --enable-default-pie --enable-default-ssp --enable-cloog-backend --enable-languages=c,c++,objc,fortran,ada --disable-libssp --disable-libmpx --disable-libmudflap --disable-libsanitizer --enable-shared --enable-threads --enable-tls --with-system-zlib --with-linker-hash-style=gnu
Thread model: posix
gcc version 8.3.0 (Alpine 8.3.0)
```

### The $ACE_ROOT/ace/config.h file

https://github.com/LASAGNE-Open-Systems/LASAGNE-Core/blob/release-1.5.1/bin/build/ace/config.h

### The $ACE_ROOT/include/makeinclude/platform_macros.GNU file

https://github.com/LASAGNE-Open-Systems/LASAGNE-Core/blob/release-1.5.1/bin/build/ace/platform_macros.GNU

### AREA/CLASS/EXAMPLE AFFECTED:

ACE.cpp fails to compile, and many if not all following units.

### Synopsis

Looks like a bunch of types are defined differently, i.e. loff_t, ucontext_t, cpu_set_t.

### Repeat by

Based on this https://github.com/LASAGNE-Open-Systems/docker-lasagne-core/blob/master/lasagne-core/Dockerfile except the top updated to read

```
FROM alpine:3.9
RUN apk update && apk add --virtual build-group build-base git figlet perl linux-headers
[...]

PR #1072, opened 2020-04-08
Add a musl compatible fallback for libc detection on Linux

drandreas:
The libc detection code is very glibc centric. At the end of config-linux.h there was already a uclib fixup. I've added an additional section for musl.

Follow up to #914

### DIFF
--- a/ACE/ace/Dev_Poll_Reactor.cpp
+++ b/ACE/ace/Dev_Poll_Reactor.cpp
@@ -1122,10 +1122,10 @@ ACE_Dev_Poll_Reactor::dispatch_io_event (Token_Guard &guard)
 
   // Define bits to check for while dispatching.
 #if defined (ACE_HAS_EVENT_POLL)
-  const __uint32_t out_event = EPOLLOUT;
-  const __uint32_t exc_event = EPOLLPRI;
-  const __uint32_t in_event  = EPOLLIN;
-  const __uint32_t err_event = EPOLLHUP | EPOLLERR;
+  const ACE_UINT32 out_event = EPOLLOUT;
+  const ACE_UINT32 exc_event = EPOLLPRI;
+  const ACE_UINT32 in_event  = EPOLLIN;
+  const ACE_UINT32 err_event = EPOLLHUP | EPOLLERR;
 #else
   const short out_event = POLLOUT;
   const short exc_event = POLLPRI;
@@ -1138,7 +1138,7 @@ ACE_Dev_Poll_Reactor::dispatch_io_event (Token_Guard &guard)
   // is invalid, there's no event there. Else process it. In any event, we
   // have the event, so clear event_ for the next thread.
   const ACE_HANDLE handle = this->event_.data.fd;
-  __uint32_t revents      = this->event_.events;
+  ACE_UINT32 revents      = this->event_.events;
   this->event_.data.fd = ACE_INVALID_HANDLE;
   this->event_.events = 0;
   if (handle != ACE_INVALID_HANDLE)
--- a/ACE/ace/config-linux.h
+++ b/ACE/ace/config-linux.h
@@ -242,6 +242,10 @@
 
 #endif /* ACE_HAS_MUSL */
 
+#if !defined (__GLIBC__) && !defined (__UCLIBC__) && !defined (ACE_HAS_MUSL)
+#  error "Could not determine C Standard Library"
+#endif /* !defined (__GLIBC__) && !defined (__UCLIBC__) && !defined (ACE_HAS_MUSL) */
+
 #include /**/ "ace/post.h"
 
 #endif /* ACE_CONFIG_LINUX_H */
--- a/ACE/ace/config-linux.h
+++ b/ACE/ace/config-linux.h
@@ -204,6 +204,44 @@
 
 #endif /* __UCLIBC__ */
 
+// To support musl (note: musl devs refuse to add a __MUSL__)
+#if defined (ACE_HAS_MUSL)
+
+// Enable stuff that musl definitly has
+#define ACE_HAS_UCONTEXT_T
+#define ACE_HAS_SIGTIMEDWAIT
+#define ACE_HAS_PTHREADS
+#define ACE_HAS_RECURSIVE_MUTEXES
+#define ACE_HAS_PTHREADS_UNIX98_EXT
+#define ACE_HAS_CPU_SET_T
+#define ACE_HAS_SIGINFO_T
+#define ACE_HAS_SOCKLEN_T
+
+// Mask some features musl lacks
+#define ACE_LACKS_SIGINFO_H
+#define ACE_LACKS_SYS_SYSCTL_H
+#define ACE_LACKS_ISCTYPE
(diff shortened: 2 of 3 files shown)


ACE/ace/config-linux.h
```diff
@@ -204,6 +204,49 @@
...
+#    undef __GLIBC__
+#  endif /* __GLIBC__ */
+
+#  if defined(ACE_HAS_SEMUN)
+#    undef ACE_HAS_SEMUN
+#  endif /* ACE_HAS_SEMUN */
+
+// Fixup strange datatypes
+#define __uint32_t uint32_t
+#define loff_t uint32_t
```
iguessthislldo, line 246: I think `loff_t` is supposed to have the ability to be 64 bits, so this isn't ideal.

Looking at my copy of musl that comes along with zig, I see this is actually defined by `fcntl.h` if `_GNU_SOURCE` is defined. You would also have to make sure `ace/os_include/sys/os_types.h`, which where `loff_t` is needed, is including `fcntl.h` (maybe if `ACE_HAS_MUSL` is defined). If you define `_GNU_SOURCE` it might also remove the need for some of these other lines as well.

If you don't want to use `_GNU_SOURCE`, then maybe just typedefing `ACE_LOFF_T` as `off_t` for musl might be enough since that's what musl is doing, I think.

ACE/ace/config-linux.h
```diff
@@ -204,6 +204,49 @@
...
+#  if defined(__GLIBC__)
+#    undef __GLIBC__
+#  endif /* __GLIBC__ */
+
+#  if defined(ACE_HAS_SEMUN)
+#    undef ACE_HAS_SEMUN
+#  endif /* ACE_HAS_SEMUN */
+
+// Fixup strange datatypes
+#define __uint32_t uint32_t
```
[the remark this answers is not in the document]
drandreas: This `__uint32_t`  define Is only used in conjunction with epoll:
```
ACE/ace/Dev_Poll_Reactor.cpp:  const __uint32_t out_event = EPOLLOUT;
ACE/ace/Dev_Poll_Reactor.cpp:  const __uint32_t exc_event = EPOLLPRI;
ACE/ace/Dev_Poll_Reactor.cpp:  const __uint32_t in_event  = EPOLLIN;
ACE/ace/Dev_Poll_Reactor.cpp:  const __uint32_t err_event = EPOLLHUP | EPOLLERR;
ACE/ace/Dev_Poll_Reactor.cpp:  __uint32_t revents      = this->event_.events;
```

I assume the idea was to match `epoll.h` from a particular libc. But I don't know what OS had it defined as `__uint32_t`. The excerpt below is from glibc-2.29 and it uses `uint32_t` for `events`:
```
struct epoll_event
{
  uint32_t events;      /* Epoll events */
  epoll_data_t data;    /* User data variable */
} __EPOLL_PACKED;
```
 Should we simply replace `__uint32_t ` with `__uint32_t` at those places?

ACE/ace/config-linux.h
drandreas: I'm not particularly fond of defining `_GNU_SOURCE` for all sources that include `ace.h`. I will try your last suggestion and push an update.

ACE/ace/config-linux.h
```diff
@@ -204,6 +204,49 @@
 
 #endif /* __UCLIBC__ */
 
+// To support musl (note: musl devs refuse to add a __MUSL__)
+#if !defined (__GLIBC__) && !defined (__UCLIBC__)
```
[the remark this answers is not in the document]
drandreas: Quote from https://wiki.musl-libc.org/faq.html

It’s a bug to assume a certain implementation has particular properties rather than testing. So far, every time somebody’s asked for this with a particular usage case in mind, the usage case was badly wrong, and would have broken support for the next release of musl. The official explanation: http://openwall.com/lists/musl/2013/03/29/13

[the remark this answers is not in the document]
drandreas: As far as I know ACE does not have any configure script that could do such detections. If it has we perhaps should peruse that path.

drandreas:
We switched from glibc to musl to ease the maintenance of a very old hardware. One of the advantages of musl is its Linux Kernel independence. It simply "works" with Kernels 2.6.39 and up and it can even be back ported to older Kernels with little effort:
https://www.openwall.com/lists/musl/2020/03/10/10

ACE/ace/Dev_Poll_Reactor.cpp
```diff
@@ -1122,10 +1122,10 @@ ACE_Dev_Poll_Reactor::dispatch_io_event (Token_Guard &guard)
 
   // Define bits to check for while dispatching.
 #if defined (ACE_HAS_EVENT_POLL)
-  const __uint32_t out_event = EPOLLOUT;
```
iguessthislldo, line 1125: ACE isn't guaranteed to have `uint32_t` and friends on all platforms. Instead use `ACE_UINT32`.

drandreas: I'm little skeptical if this makes sense. On one hand the type I'm suggesting to be replaced was already non ACE_UINT32. On the other hand `#if defined (ACE_HAS_EVENT_POLL)` will never be true for any OS but Linux, since event poll is Linux specific (not part of POSIX).

drandreas: After all we want to match the datatype of the system interface and not just define a 32bit integer.

drandreas: I still don't consider it a good idea, but I've changed it anyway.

ACE/ace/config-linux.h
```diff
@@ -204,6 +204,49 @@
...
+#define ACE_HAS_SOCKLEN_T
+
+// Mask some features musl lacks
+#define ACE_LACKS_SIGINFO_H
+#define ACE_LACKS_SYS_SYSCTL_H
+#define ACE_LACKS_RWLOCK_T
+#define ACE_LACKS_ISCTYPE
+#define ACE_LACKS_NETDB_REENTRANT_FUNCTIONS
+
+// Following the example set by uclib undef some festures
```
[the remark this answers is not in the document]
drandreas: I do not feel competent enough to perform this task. I would be thankful if this could be handled in another PR by someone else.

ACE/ace/Dev_Poll_Reactor.cpp
jwillemsen: I think very early versions of poll used __uint32_t, newer versions use uint32_t

iguessthislldo: Yea, I think I said somewhere else if it sensitive about the type, then I was less concerned about making it "proper" ACE code. If it's not right to `ACE_UINT32` then we don't have to use it.

ACE/ace/config-linux.h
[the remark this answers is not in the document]
iguessthislldo: That's understandable. ACE is a labyrinth of macros.

drandreas:
closed due to inactivity

PR #1077, opened 2020-04-15
Update to debian/* for upload to Debian

sudipm-mukherjee:


jwillemsen:
Given the file rename can you also check the release script at https://github.com/DOCGroup/ACE_TAO/blob/master/ACE/bin/make_release.py whether any different debian file renames with a new release have to be done?

jwillemsen:
As we don't guarantee binary compatibility between releases I have my doubts about the symbol files in the archive, are they really required? Would not like to maintain those as that asks for issues

jwillemsen:
Did notice that the symbol files have a version number in the content, can we remove from our archive, that is something we don't want to maintain in the main repo, maybe create them as part of the packaging step?

sudipm-mukherjee:
> As we don't guarantee binary compatibility between releases I have my doubts about the symbol files in the archive, are they really required? Would not like to maintain those as that asks for issues

Debian policy says:
`If your package provides a shared library, you should arrange to include a symbols control file following the format described above in that package. You must include either a symbols control file or a shlibs control file.`

ref: https://www.debian.org/doc/debian-policy/ch-sharedlibs.html#providing-a-symbols-file

sudipm-mukherjee:
> Did notice that the symbol files have a version number in the content, can we remove from our archive, that is something we don't want to maintain in the main repo, maybe create them as part of the packaging step?

Yes, it will be better to do it at the packaging step, the packaging script will also warn if some symbols has been removed or added and so accordingly it will provide the diff. I will remove them from the PR.

sudipm-mukherjee:
> Given the file rename can you also check the release script at https://github.com/DOCGroup/ACE_TAO/blob/master/ACE/bin/make_release.py whether any different debian file renames with a new release have to be done?

It won't need any change. I just did a micro release to test the release script and all the files in ACE/debian were renamed properly.

jwillemsen:
Are the changes complete @sudipm-mukherjee ?

sudipm-mukherjee:
> Are the changes complete @sudipm-mukherjee ?

Yes. symbols files have been removed (it can be added as a Debian packaging step). and make_release.py will not need any changes.

jwillemsen:
Thanks! Merged now, when there are more changes, please open a new PR

mitza-oci:
Is there a way to make these files (as they sit in the repo) version neutral?  There's a lot of churn for a new version.

sudipm-mukherjee:
> Is there a way to make these files (as they sit in the repo) version neutral? There's a lot of churn for a new version.

I think that can be done. If you look at the some of the file `install.in` file I created contains a generic location which is then replaced with the actual path by  debian/rules file. similar thing can be done other files also and they can be renamed based on the version.

jwillemsen:
> > Is there a way to make these files (as they sit in the repo) version neutral? There's a lot of churn for a new version.
> 
> I think that can be done. If you look at the some of the file `install.in` file I created contains a generic location which is then replaced with the actual path by debian/rules file. similar thing can be done other files also and they can be renamed based on the version.

Could you try to add that @sudipm-mukherjee, that would making releases easier

sudipm-mukherjee:
> > > Is there a way to make these files (as they sit in the repo) version neutral? There's a lot of churn for a new version.
> > 
> > 
> > I think that can be done. If you look at the some of the file `install.in` file I created contains a generic location which is then replaced with the actual path by debian/rules file. similar thing can be done other files also and they can be renamed based on the version.
> 
> Could you try to add that @sudipm-mukherjee, that would making releases easier

sure. it should not be too difficult.

PR #1066, opened 2020-04-02
When we have C++11 support make use of the C++ (u)intxx_t types

jwillemsen:
* ACE/ace/Basic_Types.h:

PR #1078, opened 2020-04-16
Define ACE_HAS_(U)INT*_T for basic types when C++11 is enabled

sonndinh:
Added hash function objects for `long long` and `unsigned long long`. Right now they're mimicking the hash for ACE_INT64 and ACE_UINT64. 

Note: It looks like the hash functions for `long long` and `unsigned long long` when they are 64-bit, and 
ACE_INT64, ACE_UINT64 may cause collisions if `unsigned long` is less than 64-bit. I'm not sure yet about the requirement for those hash functions and whether the possible collisions may cause a problem or not.


mitza-oci:
Travis CI error on macOS:
```
error: use of overloaded operator '<<' is ambiguous (with operand types 'TAO_OutStream' and 'ACE_CString::size_type' (aka 'unsigned long')
```

ACE/ace/CDR_Base.h
```diff
@@ -205,7 +205,12 @@ class ACE_Export ACE_CDR
   typedef ACE_UINT16 UShort;
   typedef ACE_INT32 Long;
   typedef ACE_UINT32 ULong;
-  typedef ACE_UINT64 ULongLong;
+
+#   if defined (ACE_HAS_UINT64_T)
```
jwillemsen, line 209: This doesn't seem a correct solution, when for MacOSX a different type for ACE_UINT64 has to be used then make use of the ACE_UINT64_TYPE macro in the config-macos*.h file(s) and not do this for all platforms in this file just for CDR_Base.h

ACE/ace/Functor.h
```diff
@@ -200,7 +200,22 @@ class ACE_Export ACE_Hash<unsigned long>
...
+};
+
+template<>
+class ACE_Export ACE_Hash<unsigned long long>
+{
+public:
+  unsigned long operator () (unsigned long long t) const;
+};
+
+#if ((ACE_SIZEOF_LONG < 8) && (ACE_SIZEOF_LONG_LONG < 8))
```
jwillemsen, line 218: Are there platforms where long long is smaller 64bit?

ACE/ace/CDR_Base.h
sonndinh: I removed this in a subsequent commit.

ACE/ace/Functor.h
sonndinh: I fixed this in a subsequent commit.

TAO/TAO_IDL/be_include/be_helper.h
```diff
@@ -167,6 +167,25 @@ class TAO_OutStream
   int gen_endif (void);
 
   // =overloaded operators
+#if defined (ACE_HAS_CPP11)
+#include <type_traits>
```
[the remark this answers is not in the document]
sonndinh: Yes, that fixed it. Thanks! All builds passed now except the ones for Visual Studio 2019.

ACE/ace/Basic_Types.h
```diff
@@ -70,6 +70,12 @@
 #  include <inttypes.h>
 # endif
 
+# if defined (ACE_HAS_CPP11)
+#  include <cstdint>
+#  define ACE_HAS_INT64_T
```
jrw972, line 75: @jwillemsen had mentioned adding #defines for the other types as well.  Is that still possible?

ACE/ace/Basic_Types.h
```diff
@@ -70,6 +70,12 @@
 #  include <inttypes.h>
 # endif
 
+# if defined (ACE_HAS_CPP11)
+#  include <cstdint>
```
iguessthislldo, line 74: This also is included a could of lines up

ACE/ace/Basic_Types.h
sonndinh: I haven't thought carefully about those types yet, but I guess it is possible. I'm not sure if we should include those in this PR though.

ACE/ace/Basic_Types.h
sonndinh: I could merge them into one.

ACE/ace/Basic_Types.h
jwillemsen: We should do that at the same moment, see https://github.com/DOCGroup/ACE_TAO/pull/1066

TAO/TAO_IDL/be_include/be_helper.h
```diff
@@ -167,6 +172,22 @@ class TAO_OutStream
...
+  template <typename Dummy = TAO_OutStream &>
+  typename std::enable_if<std::is_same<Dummy, TAO_OutStream &>::value &&
+                          !std::is_same<ACE_CDR::ULongLong, unsigned long>::value &&
+                          !std::is_same<ACE_CDR::ULong, unsigned long>::value,
+                          TAO_OutStream &>::type
+  operator << (const unsigned long num)
+  {
+    ACE_OS::fprintf (this->fp_,
+                     "%lu",
+                     num);
```
jwillemsen, line 186: Will %lu always work? We have format specifiers in ace/Basic_Types.h which are dependent on the size of the type, for example ACE_UINT32_FORMAT_SPECIFIER_ASCII and ACE_UINT64_FORMAT_SPECIFIER_ASCII

jwillemsen: Are these changes in tao_idl still required now that there are no changes anymore in ace/Basic_Types.h?

TAO/tao/Storable_Base.h
```diff
@@ -112,6 +112,33 @@ namespace TAO
     virtual Storable_Base& operator << (ACE_INT64 ) = 0;
     virtual Storable_Base& operator >> (ACE_INT64 &) = 0;
 
+#if defined (ACE_HAS_CPP11)
```
jwillemsen, line 115: Are these changes still required now that there are no changes anymore in ace/Basic_Types.h?

TAO/TAO_IDL/be_include/be_helper.h
sonndinh: Yes, `%lu` should work since the input argument is `unsigned long` exactly. 

These changes are needed because there are functions that write variables of type `ACE_CString::size_type`, which maps to `unsigned long` on some platforms, to `TAO_OutStream`, and there is no `operator<<` for `unsigned long` from the existing code on such platforms.

ACE/ace/Basic_Types.h
sonndinh: Ok, I'll add those types.

TAO/TAO_IDL/be_include/be_helper.h
jwillemsen: When `ACE_CString::size_type` is the problem, shouldn't we have a operator<< for `size_t` and use `ACE_SIZE_T_FORMAT_SPECIFIER_ASCII` as format specifier?

TAO/tao/Storable_Base.h
sonndinh: Yes, some code, e.g., `read` and `write` methods of the `orbsvcs/orbsvcs/PortableGroup/PG_Group_List_Store`class need these overloaded functions because some operands are `unsigned long` but there is no existing `operator<<` for `unsigned long`. Similar to `be_helper.h`, this may change after we have other C++11 types added.

TAO/tao/Storable_Base.h
```diff
@@ -112,6 +112,33 @@ namespace TAO
...
+#include <type_traits>
+
```
jwillemsen, line 117: Include shouldn't be in the middle of the file

TAO/TAO_IDL/be_include/be_helper.h
sonndinh: That seems a better way to do it. I'll make those changes.

TAO/tao/Storable_Base.h
sonndinh: Will move it to the top.

jwillemsen:
As reference, overview of the size of types on MacOSX, the problematic platform triggering this issue https://developer.apple.com/library/archive/documentation/Darwin/Conceptual/64bitPorting/transition/transition.html

TAO/TAO_IDL/be_include/be_helper.h
jwillemsen: So size_t is a distinct type (see my stackoverflow link below), thinking about it, shouldn't we have a `ACE_HAS_DISTINCT_SIZE_T` or something like that, so that in that case we enable the size_t overload, has it really something todo with the size of long/long long?

jwillemsen: Or the calling code has to convert to a known fixed size

sonndinh: The underlying type of `size_t` is still some primitive type. For example, it's implemented as `unsigned long` in my macOS and also in the approved answer from the stackoverflow link. The problem is that the existing set of overloaded operators may not have an operator for the underlying type of `size_t` after `int64_t` and `uint64_t` are enabled. And that could possibly happen to other platforms as well, not just macOS. 

If the calling code can convert to a fixed-size type whose overloaded operator exists, then surely we won't need those changes. I think you, Adam, and Justin have a better idea of which design we should go for.

ACE/ace/Functor.inl
```diff
@@ -104,6 +104,22 @@ ACE_Hash<unsigned long>::operator () (unsigned long t) const
   return t;
 }
 
+#if (ACE_SIZEOF_LONG == 8)
```
jwillemsen, line 107: Shouldn't this also be size_t? The check for the sizeof looks also strang

sonndinh: The check is to avoid redefining hash for `long long` (and `unsigned long long`) twice. If there was no check and size of `long` < 8 bytes, then `ACE_INT64` will be `long long` when C++11 is not enabled ( in case C++11 is enabled, `int64_t` is used and in many platforms it's implemented as `long long` as well) and the hash for `long long` will be defined twice.

TAO/TAO_IDL/be_include/be_helper.h
```diff
@@ -167,6 +172,22 @@ class TAO_OutStream
...
+  // Avoid duplication for the underlying type of size_t
+  template <typename Dummy = TAO_OutStream &>
+  typename std::enable_if<std::is_same<Dummy, TAO_OutStream &>::value &&
+                          !std::is_same<ACE_CDR::ULongLong, size_t>::value &&
+                          !std::is_same<ACE_CDR::ULong, size_t>::value,
+                          TAO_OutStream &>::type
+  operator << (const size_t num)
+  {
+    ACE_OS::fprintf (this->fp_,
+                     ACE_TEXT_ALWAYS_CHAR (ACE_SIZE_T_FORMAT_SPECIFIER),
```
mitza-oci, line 185:
```suggestion
                     ACE_SIZE_T_FORMAT_SPECIFIER_ASCII,
```
Avoids going through an encoding conversion at runtime (if wchar build).

mitza-oci:
Warnings from macOS build, seems to indicate a type mismatch between format specifiers and actual arguments:
```
Storable_FlatFileStream.cpp:479:73: warning:
      format specifies type 'unsigned long' but the argument has type
      'ACE_UINT64' (aka 'unsigned long long') [-Wformat]
    ACE_OS::fprintf (this->fl_, ACE_UINT64_FORMAT_SPECIFIER_ASCII "\n", i);
                                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~       ^
Storable_FlatFileStream.cpp:519:72: warning:
      format specifies type 'long' but the argument has type 'ACE_INT64'
      (aka 'long long') [-Wformat]
    ACE_OS::fprintf (this->fl_, ACE_INT64_FORMAT_SPECIFIER_ASCII "\n", i);
                                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~       ^
```

mitza-oci:
Building OpenDDS using this branch of ACE results in:
```
dds/DCPS/transport/framework/TransportHeader.inl:114:18: error:
      invalid operands to binary expression ('OpenDDS::DCPS::Serializer' and
      'ACE_INT64' (aka 'long long'))
         (reader >> this->source_);
```

It's possible we'd need to change OpenDDS for this, but it would be better not to.
Does this indicate that `ACE_CDR::LongLong` and `ACE_INT64` are now different types?  Is this a departure from what ACE master is doing on Linux and macOS?

jwillemsen:
> dds/DCPS/transport/framework/TransportHeader.inl:114:18: error:
>       invalid operands to binary expression ('OpenDDS::DCPS::Serializer' and
>       'ACE_INT64' (aka 'long long'))
>          (reader >> this->source_);
> ```
> 
> It's possible we'd need to change OpenDDS for this, but it would be better not to.
> Does this indicate that `ACE_CDR::LongLong` and `ACE_INT64` are now different types? Is this a departure from what ACE master is doing on Linux and macOS?

I would expect that ACE_CDR::LongLong is int64 when C++11 is enabled, that matches the IDL to C++11 language mapping also.

mitza-oci:
> 
> 
> > dds/DCPS/transport/framework/TransportHeader.inl:114:18: error:
> > invalid operands to binary expression ('OpenDDS::DCPS::Serializer' and
> > 'ACE_INT64' (aka 'long long'))
> > (reader >> this->source_);
> > ```
> > 
> > It's possible we'd need to change OpenDDS for this, but it would be better not to.
> > Does this indicate that `ACE_CDR::LongLong` and `ACE_INT64` are now different types? Is this a departure from what ACE master is doing on Linux and macOS?
> > ```
> 
> I would expect that ACE_CDR::LongLong is int64 when C++11 is enabled, that matches the IDL to C++11 language mapping also.

That doesn't seem to be the case here:
```
std::int64_t: long long
ACE_INT64: long long
ACE_CDR::LongLong: long
```

jwillemsen:
Looking at the history of Basic_Types.h I found a cleanup action of myself to remove emulated ACE_INT64/ACE_UINT64 (see https://github.com/DOCGroup/ACE_TAO/commit/38d791f18d8a67b8b9319f4b6a9579101628b28c).

Looking at CDR_Base.h, that could be simplified now, use ACE_INT64/ACE_UINT64 instead of trying to sort that out itself. Maybe something to try in a separate PR as a cleanup there first?

mitza-oci:
This is what I ran to test it:
```
#include "ace/CDR_Base.h"
#include <iostream>
#include <typeinfo>

template <typename T>
void printType(const char* label, T&)
{
  std::cout << label << ": " << typeid(T).name() << '\n';
}

int main(int, char*[])
{
  std::int64_t i64 = 0;
  ACE_INT64 ai64 = 0;
  ACE_CDR::LongLong all = 0;
  printType("std::int64_t", i64);
  printType("ACE_INT64", ai64);
  printType("ACE_CDR::LongLong", all);
  return 0;
}
```
pipe the output through `c++filt` to demangle type names

jwillemsen:
This mismatch in types could very likely be the underlying issue when trying to switch ACE_INT64/ACE_UINT64, would be in favor of first separately update CDR_Base.h and see what happens when that is cleaned, could be that a lot of other issues are gone at that moment.

TAO/TAO_IDL/be_include/be_helper.h
sonndinh: The overloaded functions for ACE_CDR::LongLong and ACE_CDR::ULongLong are currently doing this conversion. Do you think we should change those too?

mitza-oci: Yes, I don't think this combination `ACE_TEXT_ALWAYS_CHAR (ACE_SIZE_T_FORMAT_SPECIFIER)` really makes sense when there are at least 2 better alternatives:
- `ACE_SIZE_T_FORMAT_SPECIFIER_ASCII` which has type `char*` on all builds
- Just  `ACE_SIZE_T_FORMAT_SPECIFIER` which has type `wchar_t*` on wchar-enabled builds, so this calls `ACE_OS::fprintf(FILE*, const wchar_t*, ...)`

jwillemsen:
See #1090 for my cleanup to ACE_CDR, that maps CDR LongLong to ACE_INT64

ACE/ace/Functor.h
```diff
@@ -200,6 +200,32 @@ class ACE_Export ACE_Hash<unsigned long>
   unsigned long operator () (unsigned long t) const;
 };
 
+#if (ACE_SIZEOF_LONG == 8)
+/**
+ * @brief Function object for hashing a long long number
+ */
+template<>
+class ACE_Export ACE_Hash<long long>
```
jwillemsen, line 208: What about providing a specialization for all ACE_(U)INTxx defines, looks we do that for ACE_Equal in this file already

TAO/tao/Storable_FlatFileStream.h
```diff
@@ -84,6 +96,37 @@ namespace TAO
     virtual Storable_Base& operator << (ACE_INT64 );
     virtual Storable_Base& operator >> (ACE_INT64 &);
 
+#if defined (ACE_HAS_CPP11)
+    // Avoid duplication for the underlying type of size_t
+    template <typename Dummy = Storable_Base &>
+    typename std::enable_if<std::is_same<Dummy, Storable_Base &>::value &&
+                            !std::is_same<ACE_UINT64, size_t>::value &&
```
jwillemsen, line 103: What when the file is created on a 64bit system and after that read on a 32bit system, wouldn’t that be a problem when we add support for size_t?

TAO/TAO_IDL/be_include/be_helper.h
jwillemsen: At the idl level we normally work with types of a known size, with size_t this looks risky, a length of a string is 32bit in IDL, with a 64bit size_t it could cause issues, shouldnt the caller use CDR::ULong?

ACE/ace/Functor.h
jwillemsen: See #1091 for a separate PR to try that

jwillemsen:
@mitza-oci @sonndinh Can you review #1090, that looks for me the first step, cleanup old stuff first so that ACE_(U)INT64 are used consistently.

TAO/TAO_IDL/be_include/be_helper.h
sonndinh: Those are fixed.

TAO/tao/Storable_FlatFileStream.h
sonndinh: I'm not sure I understand the question. But the operators for `size_t` are guaranteed to be defined only when the underlying type of `size_t` is different than `ACE_UINT64` and `ACE_UINT32`.

TAO/TAO_IDL/be_include/be_helper.h
sonndinh: There is some code in TAO_IDL (in be/be_visitor_operation/ami_cs.cpp if I remember correctly) that calls output operator with a variable of type `ACE_CString::size_type`, which is `size_t`.

TAO/tao/Storable_FlatFileStream.h
jwillemsen: My question is whether we should allow the API usage using size_t, at that moment we marshal something with a different size what happens when the file is read on a different platform? Shouldn't the caller use UINT32/64 explicitly, that way the size on send/receive is known to be of a fixed size.

TAO/TAO_IDL/be_include/be_helper.h
jwillemsen: Shouldn't the caller there not cast first to a CORBA::ULong, the type of the length of a CORBA string, that should always be 32bit, never 64bit, which size_t could be

sonndinh: Maybe the caller should do that. But I'm not experienced enough with ACE and TAO to tell which way the caller should do. That's the code I got to start with.

TAO/tao/Storable_FlatFileStream.h
sonndinh: I think I understand the question and the potential problem with `size_t` now. Do you have any suggestions for this? I guess a simple way to fix is to use `ACE_UINT64_FORMAT_SPECIFIER_ASCII` in both output and input operators for `size_t`. That way we can avoid a potential bug when `size_t` is 64-bit on one platform and 32-bit on the other.

TAO/TAO_IDL/be_include/be_helper.h
jwillemsen: What is the exact code where this is problematic? Can you add a link to it?

TAO/tao/Storable_FlatFileStream.h
jwillemsen: Yes, let the caller cast before writing/reading, he should use a fixed size type when using any CORBA related API

jwillemsen:
Please merge master again in your branch

TAO/TAO_IDL/be_include/be_helper.h
sonndinh: Line 198 in ami_cs.cpp caused the error: [https://github.com/DOCGroup/ACE_TAO/blob/master/TAO/TAO_IDL/be/be_visitor_operation/ami_cs.cpp#L198](https://github.com/DOCGroup/ACE_TAO/blob/master/TAO/TAO_IDL/be/be_visitor_operation/ami_cs.cpp#L198).

mitza-oci: Since it's in TAO_IDL_BE, the ORB isn't directly seeing a size_t.  Instead, the code generator gets a size_t and writes it to the file as a C++ integer literal.  The type of this literal is `int` (for all reasonable string sizes), which is what will be seen by the ORB.

jwillemsen: Ok, not a problem with this addition

jwillemsen:
What about the additions in Storable_Base @mitza-oci, couldn't the addition of size_t lead to an interoperability problem?

mitza-oci:
> 
> 
> What about the additions in Storable_Base @mitza-oci, couldn't the addition of size_t lead to an interoperability problem?

There's a potential issue with this one.  Could we try taking it out?

sonndinh:
> > What about the additions in Storable_Base @mitza-oci, couldn't the addition of size_t lead to an interoperability problem?
> 
> There's a potential issue with this one. Could we try taking it out?

I guess the potential problem with the addition in Storable_Base is similar to that of the Storable_FlatFileStream. If so, I will remove their changes and update PG_Group_List_Store (the one that caused a compilation error) to use existing fixed-size input/output operators.

mitza-oci:
> 
> 
> > > What about the additions in Storable_Base @mitza-oci, couldn't the addition of size_t lead to an interoperability problem?
> > 
> > 
> > There's a potential issue with this one. Could we try taking it out?
> 
> I guess the potential problem with the addition in Storable_Base is similar to that of the Storable_FlatFileStream. If so, I will remove their changes and update PG_Group_List_Store (the one that caused a compilation error) to use existing fixed-size input/output operators. Do you think we should do that?

I looked in to how this is used.  I don't think there's actual distribution (through the ORB) of this data, it's just saved to the local filesystem.  There are potential size_t issues here, but much less of a degree than actually sending it on the network.

Still, it seems like this uncovered a misuse of the Storable_Base interface and it's a smaller change to just update the PG_Group_List_Store to cast to ACE_UINT64 instead of needing to change Storable_Base.h and Storable_FlatFileStream.h,

TAO/tao/Storable_FlatFileStream.h
sonndinh: This is fixed now.

PR #1081, opened 2020-04-19
Remove unnecessary ACE_Select_Reactor_Handler_Repository ctor init parameter.

likema:


ACE/ace/Select_Reactor_Base.cpp
```diff
@@ -109,9 +109,8 @@ ACE_Select_Reactor_Handler_Repository::open (size_type size)
 ACE_Select_Reactor_Handler_Repository::ACE_Select_Reactor_Handler_Repository (ACE_Select_Reactor_Impl &select_reactor)
   : select_reactor_ (select_reactor),
```
jwillemsen, line 110: Leads to a compile error when `ACE_SELECT_REACTOR_BASE_USES_HASH_MAP` is defined

likema: Fixed it.

PR #1084, opened 2020-04-21
Additions related to rpmbuild

egleston:
These are the updates I need to do to build after a new version of ACE-TAO is released:
- BuildRequires: has a dependency on sendmail. Not sure why, but I don't have a mail server on my build machine and this makes it impossible to run `rpmbuild -tb ACE-TAO-src.tar.gz`
- We have legacy code that uses combat-tcl for CORBA. To compile idl into tcl, it uses both `tao_ifr` and `tao_ifr_service`.

### DIFF
--- a/TAO/orbsvcs/IFR_Service/IFR_Service.mpc
+++ b/TAO/orbsvcs/IFR_Service/IFR_Service.mpc
@@ -3,6 +3,7 @@
 // necessary to add minimum_corba to 'avoids' here.
 project(IFR_Service) : orbsvcsexe, install, ifrservice, ifr_client, imr_client, svc_utils {
   exename = tao_ifr_service
+  install = $(ACE_ROOT)/bin
 
   IDL_Files {
   }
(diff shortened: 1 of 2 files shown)


TAO/orbsvcs/IFR_Service/IFR_Service.mpc
```diff
@@ -3,6 +3,7 @@
 // necessary to add minimum_corba to 'avoids' here.
 project(IFR_Service) : orbsvcsexe, install, ifrservice, ifr_client, imr_client, svc_utils {
   exename = tao_ifr_service
+  install = $(ACE_ROOT)/bin
```
jwillemsen, line 6: This changes the location of the tao_ifr_service which can break test scripts and user setups, for installing it as part of the rpm it is not necessary to change the location of where it is build, see for example tao-cosnaming in the spec file

egleston: I think I understand the issue you mentioned and adjusted my updates. I returned the IFR_Service.mpc to its original state and simply have the rpmbuild spec install from the original build location.

jwillemsen:
Thanks!

PR #1086, opened 2020-04-23
Update debian files

sudipm-mukherjee:
Three changes.
1. An error in the copyright which came out in the initial Debian review of the previous upload to Debian. 
2. Modify the files in debian/* which needs updating while releasing a new version and generalise them so that they can be created automatically while creating the Debian package.
3. Modify make_release.py to only update debian/control file.


jwillemsen:
Please don't use force pushes because that makes it much harder to see what changed, it requires a full review again which takes more time

sudipm-mukherjee:
> Please don't use force pushes because that makes it much harder to see what changed, it requires a full review again which takes more time

Sorry about that, it was just a change in Signed-off-by. https://github.com/DOCGroup/ACE_TAO/compare/86d885d117a0d5ae7b1bfb4704ad9bae7c72cc9b..1e13908ba3493cbf71b16a5497394ee7e4e62dcc should have shown you that there has been no real change.                             
Thanks for merging.

PR #1090, opened 2020-04-25
Map ACE_CDR LongLong to ACE_INT64, all platforms for which we needed …

jwillemsen:
…an emulation have been deprecated in the past.

    * ACE/ace/CDR_Base.cpp:
    * ACE/ace/CDR_Base.h:
    * ACE/ace/CDR_Size.h:
    * ACE/ace/README:
    * ACE/ace/config-kfreebsd.h:
    * ACE/ace/config-qnx.h:
    * TAO/TAO_IDL/be/be_visitor_any_extracted_type_decl.cpp:
    * TAO/TAO_IDL/be/be_visitor_argument/vardecl_ss.cpp:
    * TAO/TAO_IDL/be/be_visitor_attr_init.cpp:
    * TAO/TAO_IDL/be/be_visitor_null_return_value.cpp:
    * TAO/TAO_IDL/fe/idl.yy.cpp:
    * TAO/orbsvcs/orbsvcs/Time/TAO_UTO.cpp:
    * TAO/tao/AnyTypeCode/Any_Basic_Impl.cpp:
    * TAO/tao/TypeCodeFactory/TypeCodeFactory_i.cpp:


ACE/ace/Basic_Types.h
```diff
@@ -208,28 +208,28 @@ ACE_BEGIN_VERSIONED_NAMESPACE_DECL
 // typedefs for whatever we can.  Some of these are needed for certain
 // cases of ACE_UINT64, so do them before the 64-bit stuff.
 
-#if defined (ACE_INT8_TYPE)
-  typedef ACE_INT8_TYPE         ACE_INT8;
-#elif defined (ACE_HAS_INT8_T)
+#if defined (ACE_HAS_INT8_T)
```
mitza-oci, line 211: If I'm reading this correctly, it seems like the purpose of `ACE_*_TYPE` (not `ACE_HAS_*_T`) is to let config-myplatform.h override the automated detection.  This PR would change that behavior.

Does ace/README need to change to describe the modified precedence here?  Are there any platforms that may be negatively impacted by it?  On the other hand, are we depending on this change for the related CDR_Base (and possibly macOS) changes?

jwillemsen: Yes, when we have C++11 we enable the `ACE_HAS_*_T` types which we want to use at that moment, when the platform has for non C++11 or very old compilers a different type set using `ACE_*_TYPE` we don't want to use that old definition.

I think this are changes to cleanup CDR_Base so that we get rid of some strange things that could cause problems, including macOS. The only challenge is that I don't have a macOS system available right now, so I can't test #1078 myself

jwillemsen: With this change the types in Basic_Types.h are used also by CDR_Base without exception

mitza-oci: Right, I'm fine with that approach, as long as it's documented.  Looks like you added that in the latest commit.
There's a macOS (Mac Mini) system at our office that I can SSH to, so let me know if you need anything tested that's not easy to do with the CI builds.

jwillemsen: Ok, let me merge this first in master and after that merge master again in #1078 to see what happens.

mitza-oci:
@jwillemsen I just noticed that this changed `TAO/TAO_IDL/fe/idl.yy.cpp`, should change `fe/idl.ll` and regenerate it

PR #1091, opened 2020-04-26
Provide ACE_Hash for ACE_(U)INT16/32/64

jwillemsen:
* ACE/ace/Functor.cpp:
    * ACE/ace/Functor.h:
    * ACE/ace/Functor.inl:


jwillemsen:
Looks this doesn't work, a specialization of unsigned long is required for PortableGroup

PR #1101, opened 2020-05-06
Define ACE_LACKS_IF_NAMEINDEX for Android APIs < 24

jonesc-oci:
The functions if_nameindex() and If_freenameindex() are supported for Android APIs >= 24.


ACE/ace/config-android.h
```diff
@@ -123,6 +123,10 @@
 #  endif
 #endif
 
+#if __ANDROID_API__ < 24
```
iguessthislldo, line 126: NDK r12, the minimum revision that I consider ACE to support, has API 24, but it doesn't have `if_nameindex`, so this also depends on the NDK revision. I checked the NDKs after r12 and it appears in r14:

```suggestion
#if ACE_ANDROID_NDK_LESS_THAN(14, 0) || __ANDROID_API__ < 24
```

I would have just let my matrix catch this but something's wrong with it at the moment.

iguessthislldo:
I went ahead and put this check into the `config.h` of builds in my matrix, but it's [producing errors in the older NDKs](https://travis-ci.org/github/iguessthislldo/OpenDDS-Android/builds/684109790).

iguessthislldo:
Running this through my matrix: https://travis-ci.org/github/iguessthislldo/OpenDDS-Android/builds/686789303

ACE/ace/os_include/net/os_if.h
```diff
@@ -102,7 +102,8 @@ struct  ifconf {
 # define IFF_BROADCAST 0x2
 #endif /* IFF_BROADCAST */
 
-#if defined ACE_HAS_WINSOCK2 && defined ACE_LACKS_IF_NAMEINDEX
+#if (defined ACE_LACKS_IF_NAMEINDEX && defined ACE_HAS_WINSOCK2) || \
+  defined ACE_LACKS_STRUCT_IF_NAMEINDEX
```
jwillemsen, line 106: Can't we add ACE_LACKS_STRUCT_IF_NAMEINDEX to config-win32-common.h, that makes this file again simpler.

iguessthislldo:
So NDK r12 is fixed like you described, but NDK 14r still complaining about `ACE_OS::if_nameindex()` and I couldn't make heads or tails of it. After looking at it locally I think something is wrong with the generation of `net/if.h` for the standalone toolchains in r14 because the file is missing `if_nameindex()` and friends. When I said 14 had it originally I was basing it on the fact it was in the sysroot headers which are kinda different from the toolchains get. I'm thinking we should move the check to NDK r15 instead.

ACE/ace/os_include/net/os_if.h
jwillemsen: ACE_HAS_WINSOCK2 is always defined on Windows, so very likely ACE_LACKS_STRUCT_IF_NAMEINDEX can just be added to config-win32-common.h

iguessthislldo:
[I triggered another build](https://travis-ci.org/github/iguessthislldo/OpenDDS-Android/builds/687109184)

PR #1102, opened 2020-05-07
Update various hyperlinks and refer to github issues for problem reports

jwillemsen:


ACE/ACE-INSTALL.html
```diff
@@ -451,7 +451,7 @@ <h2><a name="win32">Building and Installing ACE on Windows</a></h2>
 "Clean" command will probably not account for all existing files.</p>
 
 <p>For using MPC and our perl based test framework we recommend
-our windows users to use <a href="http://www.activestate.com/Products/ActivePerl/"> Active
+our windows users to use <a href="https://www.activestate.com/products/perl"> Active
```
mitza-oci, line 454: Should we take this opportunity to switch to (or also list) Strawberry Perl?

jwillemsen: We are mostly using activestate perl without problems, will add strawberry perl

PR #1105, opened 2020-05-08
On Windows vista or newer there is a limit of 32767 per environment variable

jwillemsen:


mitza-oci:
I barely remember working on this one so long ago:
https://github.com/DOCGroup/ACE_TAO/commit/2e295eb6ac00ce44602cb14eceed9940eceb6d1d#diff-cc76a7b8b1121f762d081389be34624e

@jwillemsen how do we associate our GitHub accounts with these imported commits?  Seems to work for yours

jwillemsen:
I think the associate happened when converting from svn to ISIS git, as far as I know we can't do that anymore without rewriting the full git history which could cause problems for all forks and clones everywhere

PR #1108, opened 2020-05-11
An incorrect dynamic_cast should throw std::bad_cast which is derived…

jwillemsen:
… from std::exception, let us test that we get std::bad_cast

    * ACE/tests/Compiler_Features_13_Test.cpp:


jwillemsen:
Looks we are running out of disk space again, see https://github.com/actions/virtual-environments/issues/709

PR #1114, opened 2020-05-15
Code changes to workaround the Visual Studio 2019 crash, instead of d…

jwillemsen:
…eferencing the instance point just take the typeid of ACE_Process_Manager directly


jwillemsen:
The name argument looks only to be there for debugging purposes, it is stored inside of ACE_Cleanup_Info_Node but not actively used

mitza-oci:
> 
> 
> The name argument looks only to be there for debugging purposes, it is stored inside of ACE_Cleanup_Info_Node but not actively used

The only time it would be different is when the user sets a custom process manager as the `instance_` singleton.  Should this change be conditional on `_MSC_VER == 1925`?

jwillemsen:
> > The name argument looks only to be there for debugging purposes, it is stored inside of ACE_Cleanup_Info_Node but not actively used
> 
> The only time it would be different is when the user sets a custom process manager as the `instance_` singleton. Should this change be conditional on `_MSC_VER == 1925`?

In the getter function we create the ACE_Process_Manager, so it can't be different, in the getter I have used *t, that worked as far as I remember from yesterday, testing that right now on azure.

jwillemsen:
> See other comment in PR. While we're looking at this one, it would be good to remove some of this duplication and improperly indented code. For example, create a macro that expands to either `ACE_Process_Manager_cleanup` or `ACE_Process_Manager::cleanup`

Done, added a macro for that

ACE/ace/Process_Manager.cpp
```diff
@@ -157,18 +151,10 @@ ACE_Process_Manager::instance (ACE_Process_Manager *tm)
...
-                                0,
-                                typeid (*ACE_Process_Manager::instance_).name ());
-#else
   ACE_Object_Manager::at_exit (ACE_Process_Manager::instance_,
-                                ACE_Process_Manager::cleanup,
+                                ACE_PROCESS_MANAGER_CLEANUP_FUNCTION,
                                 0,
-                                typeid (*ACE_Process_Manager::instance_).name ());
-#endif /* ACE_HAS_SIG_C_FUNC */
+                                typeid (*t).name ());
```
mitza-oci, line 157:
```suggestion
                                typeid (*tm).name ());
```

jwillemsen: I first had that also, but when this method is called ACE_Process_Manager::instance_ still points to the value it had, where tm points to the new value. I doubt if the code is correct, but using t matches the current behavior

mitza-oci: So `at_exit` has the wrong pointer?  I agree that fixing it would be outside. the scope of this PR.

jwillemsen: Yes, that looks the case, but something that is outside of the scope, will create a new issue so that this is on the todo list

jwillemsen:
See https://github.com/DOCGroup/ACE_TAO/issues/1117 for the new PR

PR #1116, opened 2020-05-15
Modify Android link on Windows hosts to avoid error

jonesc-oci:
Modify the Android link step on Windows hosts to avoid a 'command line too long' error when linking a large number of object files.

### DIFF
--- a/ACE/include/makeinclude/platform_android.GNU
+++ b/ACE/include/makeinclude/platform_android.GNU
@@ -180,11 +180,10 @@ LIBS += -llog
 
 # link step to avoid 'command line too long' error on Windows
 ifeq ($(OS), Windows_NT)
-  SHOBJS_FILE = $(VSHDIR)tmpfile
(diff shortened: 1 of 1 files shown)


ACE/include/makeinclude/platform_android.GNU
```diff
@@ -177,3 +177,14 @@ endif
...
 # Link To Android Logging Library for Log_Msg_Android_Logcat
 LIBS += -llog
+
+# link step to avoid 'command line too long' error on Windows
+ifeq ($(OS), Windows_NT)
+  SHOBJS_FILE = $(VSHDIR)tmpfile
+  CLEANUP_OBJS += $(SHOBJS_FILE)
+  define SHLIBBUILD
+    echo " " > $(SHOBJS_FILE)
+    $(foreach myobj, $(VSHOBJS), echo $(myobj) >> $(SHOBJS_FILE);)
```
mitza-oci, line 187: Would `$^` work in place of `$(VSHOBJS)`?  The difference is that other places in the makefiles could add prerequisites to this rule (not sure that ever happens in practice).

PR #1120, opened 2020-05-25
Regenerated with bison 3.6.2

jwillemsen:
* TAO/TAO_IDL/fe/idl.tab.cpp:
    * TAO/TAO_IDL/fe/idl.tab.hpp:


iguessthislldo:
Because regenerating the parser is always a decent sized patch, I'm not sure we should be doing this just for the sake of using a new bison unless we're also changing the grammar or there's a specific problem with the generated code that a new version fixes. I have some annotation work coming up this week where I was going to do this anyway.

jwillemsen:
The advantage of a separate upgrade is that we can cleabnly separate the changes from the upgrade and changes from the grammar.

Issue #1118, opened 2020-05-19
New warnings from #1005


mitza-oci:
Looks we have some new warnings, see https://buildlogs.remedy.nl/rhel70_atcd_release_ndds523rev0/2020_05_19_10_39_Brief.html#section_5

```
g++ -fvisibility=hidden -fvisibility-inlines-hidden -Wnon-virtual-dtor -O3 -m64 -pthread -fno-strict-aliasing -Wall -W -Wpointer-arith -pipe -D_GNU_SOURCE -DNDEBUG -I/home/build/jenkins/workspace/rhel70_atcd_release_ndds523rev0/BUILD/DOC_ROOT/ACE -DACE_NDEBUG -D__ACE_INLINE__ -I/home/build/jenkins/workspace/rhel70_atcd_release_ndds523rev0/BUILD/DOC_ROOT/ACE -I.. -Iinclude -Ife -DTAO_IDL_FE_BUILD_DLL -c -fPIC -o .shobj/util/utl_global.o /home/build/jenkins/workspace/rhel70_atcd_release_ndds523rev0/BUILD/DOC_ROOT/TAO/TAO_IDL/util/utl_global.cpp
/home/build/jenkins/workspace/rhel70_atcd_release_ndds523rev0/BUILD/DOC_ROOT/TAO/TAO_IDL/util/utl_global.cpp: In member function ‘void IDL_GlobalData::eval(const char*, bool)’:
/home/build/jenkins/workspace/rhel70_atcd_release_ndds523rev0/BUILD/DOC_ROOT/TAO/TAO_IDL/util/utl_global.cpp:1979:63: warning: ‘old.{anonymous}::OldState::default_streambuf_’ may be used uninitialized in this function [-Wmaybe-uninitialized]
ACE_DEFAULT_LOG_STREAM->rdbuf (default_streambuf_);
^
/home/build/jenkins/workspace/rhel70_atcd_release_ndds523rev0/BUILD/DOC_ROOT/TAO/TAO_IDL/util/utl_global.cpp:2000:12: note: ‘old.{anonymous}::OldState::default_streambuf_’ was declared here
OldState old (disable_output);
^
```

_Originally posted by @jwillemsen in https://github.com/DOCGroup/ACE_TAO/pull/1005#issuecomment-630845908_

mitza-oci:
@knutpett Please take a look at this one

jwillemsen:
@knutpett Any updates related to this warning?

[...]

PR #1122, opened 2020-05-27
Fixed warning for maybe-uninitialized

knutpett:


knutpett:
This will solve https://github.com/DOCGroup/ACE_TAO/issues/1118

PR #1124, opened 2020-05-27
Set ACE_HAS_CPP11|14|17|20 based on __cplusplus for clang embarcadero

jwillemsen:


jwillemsen:
Needs #1123 to get our examples/tests to compile

jwillemsen:
Depends on #1121

PR #1130, opened 2020-05-31
Change TAO/tests/IDLv4/TestIDLv4.idl into valid IDLv4

okellogg:


okellogg:
> So perhaps do a force-push here or a new PR?

I'll go for new PR then.

PR #1135, opened 2020-06-05
Anonymous Types in IDL4

iguessthislldo:


TAO/TAO_IDL/util/utl_global.cpp
```diff
@@ -1844,19 +1844,22 @@ IDL_GlobalData::anon_type_diagnostic (
 bool
 IDL_GlobalData::anon_error (void) const
 {
-  return (this->anon_type_diagnostic_ == ANON_TYPE_ERROR);
+  return anon_type_diagnostic_ == ANON_TYPE_ERROR &&
+    !(idl_version_ >= IDL_VERSION_4);
```
mitza-oci, line 1848:
```suggestion
    idl_version_ < IDL_VERSION_4;
```

TAO/tests/IDLv4/anonymous_types/test.idl
```diff
@@ -0,0 +1,17 @@
...
+
+struct TestStruct {
+  short short_array[10];
+  sequence<long> long_sequence;
+  sequence<long, 3> bounded_long_sequence;
+};
+
+// Another struct, different lengths
+struct AnotherTestStruct {
+  short short_array[9];
```
mitza-oci, line 14: Would be good to also have the same array type appear twice, to make sure the generated code doesn't have multiple-definition errors.

iguessthislldo: Oh yea, good point.

jwillemsen:
Not all language mappings do support anonymous types, the IDL to C++11 language mapping doesn't support them because it is impossible to use IDL::traits for template meta programming when anonymous types are used in user IDL. 

Personally I was not in favor of adding them back to IDL4, there was a reason to deprecate them in the past.

TAO/TAO_IDL/util/utl_err.cpp
```diff
@@ -185,9 +185,11 @@ error_string (UTL_Error::ErrorCode c)
     case UTL_Error::EIDL_KEYWORD_WARNING:
       return "Warning - spelling differs from IDL keyword only in case: ";
     case UTL_Error::EIDL_ANONYMOUS_ERROR:
-      return "Error: anonymous types are deprecated by OMG spec";
+      return "Error: anonymous types are deprecated by OMG spec in IDL3, "
+        "but are valid again in IDL4";
     case UTL_Error::EIDL_ANONYMOUS_WARNING:
-      return "Warning - anonymous types are deprecated by OMG spec";
+      return "Warning - anonymous types are deprecated by OMG spec in IDL3, "
+        "but are valid in IDL4";
```
jwillemsen, line 192: Maybe add "but not supported in all IDL language mappings"

mitza-oci:
> 
> 
> Not all language mappings do support anonymous types, the IDL to C++11 language mapping doesn't support them because it is impossible to use IDL::traits for template meta programming when anonymous types are used in user IDL.

Language mappings based on IDL3 should use the IDL3 settings of tao_idl.


> Personally I was not in favor of adding them back to IDL4, there was a reason to deprecate them in the past.

We can't do anything about that at this point, since they are widely used in XTypes and other DDS specs, and also in ROS2.

jwillemsen:
> We can't do anything about that at this point, since they are widely used in XTypes and other DDS specs, and also in ROS2.

:-( regret to hear that, it is a small effort to just add a typedef and use that. It would make template meta programming using IDL::traits so much easier for the user, especially in the DDS world where a lot of APIs are just templated APIs.

TAO/TAO_IDL/util/utl_err.cpp
mitza-oci: And/Or... "in those Profiles of IDL4 which include the Anonymous Types Building Block"

Or, since utl_err.cpp is part of TAO_IDL_FE, this could be changed to customize the error/warning messages based on which backend is used.  The backends are more directly related to language mappings (although this is not an entirely clean separation, it's enough to be useful).

iguessthislldo: I don't really want to do anything else here. I think it's up to the backend to complain about typedefs if it doesn't support them, which it can do easily when it encounters a `AST_Typedef`.

iguessthislldo: I mixed that up, I meant that I'm assuming the fact that it's an anonymous type can be detected by the backend. If that's not possible or easy, then I'm fine with making changes for that.

okellogg:
> :-( regret to hear that, it is a small effort to just add a typedef and use that.

+1

> It would make template meta programming using IDL::traits so much easier [...]

Similar in Ada where they create problems due to the language's particularly strong typing.
Since early on we have had a design rule prohibiting anon types.

jwillemsen:
We should make it possible for the user to enable IDL4 but disallow anonymous types so that any developments using TAO doesn't break Ada/C++11 parts of the same system later on

iguessthislldo:
Alright now it should be possible to override the IDL4 anonymous behavior from the backend or the command line the same as with IDL3.

TAO/TAO_IDL/util/utl_err.cpp
```diff
@@ -185,9 +185,11 @@ error_string (UTL_Error::ErrorCode c)
     case UTL_Error::EIDL_KEYWORD_WARNING:
       return "Warning - spelling differs from IDL keyword only in case: ";
     case UTL_Error::EIDL_ANONYMOUS_ERROR:
-      return "Error: anonymous types are deprecated by OMG spec";
+      return "Error: anonymous types are deprecated by OMG spec in IDL3, "
```
jwillemsen, line 188: Shouldn't these messages be updated to just say that anonymous types are disabled without referring to deprecated/IDL4 now that it is up to the user/backend to decide when they are not allowed or just trigger a warning?

iguessthislldo: By default it is still based on the IDL version, so I'd like to keep that message when that is the case since it explicitly says what the problem is. I can do this by switching the error message based on if the setting has been explicitly set. For the warning message I can make it generic unconditionally.

TAO/TAO_IDL/util/utl_err.cpp
```diff
@@ -185,9 +185,12 @@ error_string (UTL_Error::ErrorCode c)
     case UTL_Error::EIDL_KEYWORD_WARNING:
       return "Warning - spelling differs from IDL keyword only in case: ";
     case UTL_Error::EIDL_ANONYMOUS_ERROR:
-      return "Error: anonymous types are deprecated by OMG spec";
+      return "anonymous types require the IDL version to be 4 or later or must "
```
jwillemsen, line 188: Shouldn't be here Error/Warning be used in the message?

iguessthislldo: "Warning" and "Error"  are already printed by `warning_header` and `error_header` functions respectively.

jwillemsen: Ok

jwillemsen:
When you tested this I am ok

PR #1141, opened 2020-06-14
tao_idl: only write "processing <filename>" to stderr in verbose mode

mitza-oci:


mitza-oci:
> I think I'm okay with this. I can't think of a situation where the message is particularly useful. Maybe if a CI test was processing multiple IDL files in one command and exiting with an error without saying which IDL file is causing the problem, but I had to think for a minute to come up with that situation.

Before this change it's skipping that line for "multi_file_input" (maybe there's other output for that case somewhere else).

By the way - the motivation behind this change is that some build systems consider it a problem (maybe not an error, more like a warning) when any build step writes to stderr.

iguessthislldo:
> By the way - the motivation behind this change is that some build systems consider it a problem (maybe not an error, more like a warning) when any build step writes to stderr.

Ah yes, I've always wondered if this would be a problem somewhere.

PR #1157, opened 2020-07-14
TP Reactor was excessively calling select

danielkempenich:
The TP Reactor implementation of the clear_dispatch_mask was implemented
by clearing the entire ready set instead of just removing the handler
that was suspended.  This forced the TP Reactor to potentially dispatch
only a single handler before re-calling an expensive select() call and
restarting the dispatch with the new set returned from select().  This
heavily favored lower order fds and was inefficient.  This originated
with the fix for ACE bugzilla 1890 which appears to have mainly focused
on the single threaded select reactor.

### DIFF
--- a/ACE/ace/TP_Reactor.inl
+++ b/ACE/ace/TP_Reactor.inl
@@ -105,12 +105,22 @@ ACE_TP_Reactor::clear_handle_read_set (ACE_HANDLE handle)
 }
 
 ACE_INLINE void
-ACE_TP_Reactor::clear_dispatch_mask (ACE_HANDLE ,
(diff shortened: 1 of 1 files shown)


jwillemsen:
What about removing the implementation in TP_Reactor completely and just use the implementation in ACE_Select_Reactor_Impl?

danielkempenich:
Sure thing, can do.

danielkempenich:
Though the base class sets state_changed_ to true, won't that reset all the handles in get_event_for_dispatching?  I can try to take a deeper look comparing with the older behavior as this was degrading performance for us.

PR #1158, opened 2020-07-18
Add LICENSE to the repo

chenrui333:


jwillemsen:
ACE/TAO have its own license, not Apache-2, see https://github.com/DOCGroup/ACE_TAO/blob/master/ACE/COPYING and https://github.com/DOCGroup/ACE_TAO/blob/master/TAO/COPYING

PR #1162, opened 2020-07-21
glibc >= 2.32 has removed sys/sysctl.h see

kraj:
https://sourceware.org/git/?p=glibc.git;a=commit;h=076f09afbac1aa57756faa7a8feadb7936a724e4

This check therefore ensures that its only used on linux when glibc has support for it

Signed-off-by: Khem Raj <raj.khem@gmail.com>

### DIFF
--- a/ACE/ace/config-linux.h
+++ b/ACE/ace/config-linux.h
@@ -20,6 +20,10 @@
 
 #include "ace/config-linux-common.h"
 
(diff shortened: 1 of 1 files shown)


iguessthislldo:
I think this is covered by #1144.

jwillemsen:
This should have been fixed in the x.5.10 release (see https://github.com/DOCGroup/ACE_TAO/releases/tag/ACE%2BTAO-6_5_10), can you try that, when that doesn't work please reopen this

PR #1005, opened 2019-12-09
Improve exception safety for IDL_GlobalData::eval()

knutpett:
If eval() parses code which causes an exception, then
the output won't be re-enabled. Fixed by using a helper
class that re-enables in the dtor.

PR #1165, opened 2020-07-27
TAO_IDL: Memory management for eval

knutpett:
I discovered that with the version introducing the OldState helper class there is a problem with seems to be double delete. This was discovered in our code base that does something similar to store/restore state of IDL_GlobalData before invoking eval().

The easiest way for me to reproduce the error in plain ACE/TAO was to instanciate OldState twice which can seem to be a
little far fetched and shouldn't be there in the final code. So I temporarily did this to make the tests fail, and then fixed the bug before removing the temp code.
Please have a look at the code and let me know it there is a better way to do this.

Before merging to master I'd like to to a final check on the code so we're sure the result is correct.

### DIFF
--- a/TAO/TAO_IDL/util/utl_global.cpp
+++ b/TAO/TAO_IDL/util/utl_global.cpp
@@ -2023,7 +2023,6 @@ void
 IDL_GlobalData::eval (const char *string, bool disable_output)
 {
   OldState old (disable_output);
-  OldState old2 (disable_output);
 
   // Set up Flex to read from string
   tao_yy_scan_string (string);
--- a/TAO/TAO_IDL/util/utl_global.cpp
+++ b/TAO/TAO_IDL/util/utl_global.cpp
@@ -1953,11 +1953,12 @@ namespace
   {
   public:
     explicit OldState (bool disable_output = false)
-        : old_filename_ (idl_global->filename () ? new UTL_String (idl_global->filename (), true) : 0), // need a copy because IDL_GlobalData::set_filename() destroys previous value
+        : old_filename_ (idl_global->filename () ? new UTL_String (idl_global->filename (), true) : 0),
+          // need a copy because IDL_GlobalData::set_filename() destroys previous value
           old_lineno_ (idl_global->lineno ()),
           old_idl_src_file_ (idl_global->idl_src_file ()),
           disable_output_ (disable_output),
-          default_streambuf_ (0),
+          default_streambuf_ (ACE_DEFAULT_LOG_STREAM->rdbuf ()),
           flags_ (ACE_LOG_MSG->flags ())
       {
         idl_global->in_eval_ = true;
@@ -1971,19 +1972,18 @@ namespace
 
         if (disable_output_)
           {
-            default_streambuf_ = ACE_DEFAULT_LOG_STREAM->rdbuf ();
             ACE_DEFAULT_LOG_STREAM->rdbuf (0);
             ACE_LOG_MSG->clr_flags (ACE_Log_Msg::STDERR);
             ACE_LOG_MSG->clr_flags (ACE_LOG_MSG->flags ());
          }
       }
 
-      UTL_String pseudo_filename ()
+      UTL_String pseudo_filename () const
       {
         // Name this pseudo-file "builtin-N"
         static char buffer[64];
         ACE_OS::snprintf (&buffer[0], sizeof buffer, "builtin-%u", pseudo_filename_counter_);
-        UTL_String filename (&buffer[0], true);
+        const UTL_String filename (&buffer[0], true);
 
         return filename;
       }
@@ -2010,11 +2010,11 @@ namespace
       }
 
     private:
-      UTL_String *old_filename_;
-      long old_lineno_;
-      UTL_String *old_idl_src_file_;
-      bool disable_output_;
-      std::streambuf *default_streambuf_;
+      UTL_String *const old_filename_;
+      const long old_lineno_;
+      UTL_String *const old_idl_src_file_;
+      const bool disable_output_;
+      std::streambuf *const default_streambuf_;
       const unsigned long flags_;
       static unsigned pseudo_filename_counter_;
   };
--- a/TAO/TAO_IDL/util/utl_global.cpp
+++ b/TAO/TAO_IDL/util/utl_global.cpp
@@ -2016,6 +2016,7 @@ void
 IDL_GlobalData::eval (const char *string, bool disable_output)
 {
   OldState old (disable_output);
+  OldState old2 (disable_output);
 
   // Set up Flex to read from string
   tao_yy_scan_string (string);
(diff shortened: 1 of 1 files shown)


TAO/TAO_IDL/util/utl_global.cpp
```diff
@@ -1982,6 +1978,17 @@ namespace
          }
       }
 
+      UTL_String get_filename()
```
mitza-oci, line 1981:
```suggestion
      UTL_String get_filename ()
```
Is this chunk broken into is own function so it can be called from somewhere else?  If so, it looks like managing the value of `n` should be handled somewhere else (constructor?) so that `get_filename ()` could be called multiple times on the same object and return the same result.

TAO/TAO_IDL/util/utl_global.cpp
```diff
@@ -1965,11 +1965,7 @@ namespace
         idl_global->set_lineno (1);
         idl_global->set_filename (0);
 
-        // Name this pseudo-file "builtin-N"
-        static char buffer[64];
-        static unsigned n = 1;
-        ACE_OS::snprintf (&buffer[0], sizeof buffer, "builtin-%u", n++);
-        UTL_String utl_string (&buffer[0], true);
+        UTL_String utl_string = get_filename();
```
mitza-oci, line 1968:
```suggestion
        UTL_String utl_string = this->get_filename ();
```

TAO/TAO_IDL/util/utl_global.cpp
```diff
@@ -1953,7 +1953,7 @@ namespace
   {
   public:
     explicit OldState (bool disable_output = false)
-        : old_filename_ (idl_global->filename ()),
+        : old_filename_ (idl_global->filename () == 0 ? 0 : new UTL_String(idl_global->filename (), true)), // need a copy because set_filename() destroys previous value
```
mitza-oci, line 1956:
```suggestion
        : old_filename_ (idl_global->filename () ? new UTL_String (idl_global->filename (), true) : 0), 
        // need a copy because set_filename() destroys previous value
```
I didn't see a set_filename() in this class, so I guess this is IDL_GlobalData's set_filename?  Maybe clarify in the comment.

jwillemsen:
Related to #1005

TAO/TAO_IDL/util/utl_global.cpp
knutpett: Ok, so I can assumed that using "this->" is the preferred "style" in ACE/TAO?

mitza-oci: https://github.com/DOCGroup/ACE_TAO/blob/9097c5600873fd147884c0903fc285f1be0c2de7/ACE/docs/ACE-guidelines.html#L806

TAO/TAO_IDL/util/utl_global.cpp
```diff
@@ -1999,6 +2005,7 @@ namespace
           }
 
         tao_yylex_destroy ();
+        pseudo_filename_counter_++;
```
mitza-oci, line 2008:
```suggestion
        ++pseudo_filename_counter_;
```

TAO/TAO_IDL/util/utl_global.cpp
```diff
@@ -1953,7 +1953,7 @@ namespace
   {
   public:
     explicit OldState (bool disable_output = false)
-        : old_filename_ (idl_global->filename ()),
+        : old_filename_ (idl_global->filename () ? new UTL_String (idl_global->filename (), true) : 0), // need a copy because IDL_GlobalData::set_filename() destroys previous value
```
mitza-oci, line 1956: break up long line

TAO/TAO_IDL/util/utl_global.cpp
```diff
@@ -1982,6 +1978,16 @@ namespace
          }
       }
 
+      UTL_String pseudo_filename ()
```
mitza-oci, line 1981:
```suggestion
      UTL_String pseudo_filename () const
```
Not that it matters all that much, but this should be able to be const (and actually that could extend to static, but that may just be confusing).  I was thinking yesterday that this doesn't seem to need to be its own function if it's not used anywhere else.

TAO/TAO_IDL/util/utl_global.cpp
knutpett: What is the recommended maximum number of characters? It's not easy to break that line very nice - I'll try to break at the comment. Let me know how it looks.

TAO/TAO_IDL/util/utl_global.cpp
knutpett: I'll fix const on that function, and I can also put on const on some of the other pointers. But I don't see how I can put on const on that static since it will be modified?

I know the function is only used once, but I think small helper functions help on readability and SRP. I can revert it if you like.

mitza-oci: I meant the function itself can be static since it doesn't use `this`, but it's fine as-is.

TAO/TAO_IDL/util/utl_global.cpp
mitza-oci: The official guidelines are 80 characters but as long as it's reasonably close to that...

knutpett:
Ok, I'll double check that this works for us. I'll let you know.

TAO/TAO_IDL/util/utl_global.cpp
```diff
@@ -1953,35 +1953,41 @@ namespace
...
       }
 
+      UTL_String pseudo_filename () const
+      {
+        // Name this pseudo-file "builtin-N"
+        static char buffer[64];
+        ACE_OS::snprintf (&buffer[0], sizeof buffer, "builtin-%u", pseudo_filename_counter_);
+        const UTL_String filename (&buffer[0], true);
+
+        return filename;
```
knutpett, line 1988: Ups, UTL_String does not have a copy constructor, so we cannot return a UTL_String. On linux this works by accident, but with visual studio on windows we'll get an access violation because the filename object is "bogus".

I'll revert to how it was before the pseudo_filename() helper function.

knutpett: Everything works now

knutpett:
@jwillemsen: ok to merge into master now?

jwillemsen:
I don't have spare free time to review and merge, this one is for @mitza-oci to merge, he merged your original issue. When he doesn't have spare free time your have to wait until someone has time to review and check all results the 48 hours after the merge. 

Support on github is always best effort unless you have a contract with a commercial support company like Remedy IT for which I work.

mitza-oci:
We will handle it.  @knutpett your last note (unless I missed one) was:
> Ok, I'll double check that this works for us. I'll let you know.

So I was waiting for your confirmation.

knutpett:
Everything works fine now! :)

PR #1167, opened 2020-07-29
Fix #1166: Support the portspan endpoint option in DIOP

### LINKED ISSUE #1166: DIOP endpoint port span support
Using TAO 2.5.5, we observed that using `portspan` with DIOP results in an error "Invalid DIOP option: <portspan>." I confirmed in source code that the `DIOP_Acceptor` does not support `portspan` in the latest master. We need to use the `portspan` endpoint option with DIOP, so we have modified the `DIOP_Acceptor` to support the `portspan` endpoint option (pull request coming after I file this issue).

nickwilliams-zaxiom:
- First, fixed the `ORB_init/Portspan` test, which was failing in master:
  - `Spawn()` returns instantly and does not wait for the process to fail, so the test was failing with the error "ERROR: Last server didn't fail! Err:0."
  - Updated the test to, after spawning, wait a second to make sure the process is still running (or, in the case of the last server, to check that the process has failed).
  - The test now passed.
- Next, updated `ORB_init/Portspan` to test DIOP as well as IIOP. The test then failed.
- Finally, updated `DIOP_Acceptor` to support `portspan` in the same manner as `IIOP_Acceptor` (namely, to loop through the port span until an open port is found). The test now passes again.


jwillemsen:
Looks the test was not listed in https://github.com/DOCGroup/ACE_TAO/blob/master/TAO/bin/tao_orb_tests.lst, please add it there

jwillemsen:
New option needs to be documented in the TAO documentation under TAO/docs

TAO/tao/Strategies/DIOP_Acceptor.cpp
```diff
@@ -1038,6 +1071,20 @@ TAO_DIOP_Acceptor::parse_options (const char *str)
                                  ACE_TEXT ("endpoint priorities no longer supported.\n")),
                                 -1);
             }
+          else if (name == "portspan")
+            {
+              int range = static_cast <int> (ACE_OS::atoi (value.c_str ()));
+              // @@ What's the lower bound on the range?  zero, or one?
+              if (range < 1 || range > ACE_MAX_DEFAULT_PORT)
+                TAOLIB_ERROR_RETURN ((LM_ERROR,
+                                   ACE_TEXT ("TAO (%P|%t) Invalid IIOP endpoint ")
```
jwillemsen, line 1080: DIOP

jwillemsen:
test gets a dependency on strategies library which means the mpc file needs an update for a static build and the build order

jwillemsen:
Would suggest to move the test one directory higher so that it is not hidden within the ORB init directory

TAO/tao/Strategies/DIOP_Acceptor.cpp
nickwilliams-zaxiom: Fixed. My change will show when I push the next commit.

nickwilliams-zaxiom:
> Looks the test was not listed in https://github.com/DOCGroup/ACE_TAO/blob/master/TAO/bin/tao_orb_tests.lst, please add it there. There is a NO_DIOP to disable DIOP related tests in this file when DIOP is not available, that seems not possible anymore to do with your test change, maybe make the protocol a commandline option and list the test twice in the lst file?

> Would suggest to move the test one directory higher so that it is not hidden within the ORB init directory

> New option needs to be documented in the TAO documentation under TAO/docs

> test gets a dependency on strategies library which means the mpc file needs an update for a static build and the build order

Changes applied as requested. They will show when I push the next commit.

nickwilliams-zaxiom:
Okay, I have pushed a new commit that I believe addresses all of the comments.

TAO/tao/Strategies/DIOP_Acceptor.cpp
```diff
@@ -1038,6 +1071,20 @@ TAO_DIOP_Acceptor::parse_options (const char *str)
                                  ACE_TEXT ("endpoint priorities no longer supported.\n")),
                                 -1);
             }
+          else if (name == "portspan")
+            {
+              int range = static_cast <int> (ACE_OS::atoi (value.c_str ()));
```
jwillemsen, line 1076: could be const

TAO/tao/Strategies/DIOP_Acceptor.cpp
```diff
@@ -393,27 +394,59 @@ int
 TAO_DIOP_Acceptor::open_i (const ACE_INET_Addr& addr,
                            ACE_Reactor *reactor)
 {
-  ACE_NEW_RETURN (this->connection_handler_,
-                  TAO_DIOP_Connection_Handler (this->orb_core_),
-                  -1);
+  unsigned short const requested_port = addr.get_port_number ();
+  ACE_UINT32 last_port = requested_port + this->port_span_ - 1;
+  if (last_port > ACE_MAX_DEFAULT_PORT)
```
jwillemsen, line 399: Could use ACE_MIN here which could make last_port const

TAO/tao/Strategies/DIOP_Acceptor.cpp
jwillemsen: as ACE_OS::atoi returns int a cast is not necessary here

jwillemsen:
See #1220 for some related cleanup

TAO/docs/ORBEndpoint.html
```diff
@@ -578,20 +578,54 @@ <h3>DIOP Endpoint Overview</h3>
...
+          <TH>Option</TH>
+          <TH>Availability</TH>
+          <TH>Description</TH>
+        </TR>
+        <TR>
+          <TD>
+            <CODE>portspan</CODE>
+          </TD>
+          <TD>
+            <CODE>TAO 2.5.11</CODE>
```
jwillemsen, line 605: Needs to be updated

TAO/tao/Strategies/DIOP_Acceptor.cpp
jwillemsen, line 1076:
``` suggestion
              int const range = ACE_OS::atoi (value.c_str ());

jwillemsen (changes requested):

TAO/docs/ORBEndpoint.html
nickwilliams-zaxiom: Updated.

TAO/tao/Strategies/DIOP_Acceptor.cpp
nickwilliams-zaxiom: Great suggestion. Done.

TAO/tao/Strategies/DIOP_Acceptor.cpp
nickwilliams-zaxiom: Good to know. Done per your suggestion below.

nickwilliams-zaxiom: Implemented your suggestion.

TAO/tao/Strategies/DIOP_Acceptor.cpp
jwillemsen: Can you also update SCIOP_Acceptor.cpp and SSLIOP_Acceptor with the similar change?

nickwilliams-zaxiom: Done.

jwillemsen (approved):

jwillemsen:
Thanks!

PR #1170, opened 2020-08-01
Improve Multicast Join by Interface Name on Windows

simpsont-oci:
Currently on windows, joining multicast by interface name doesn't work for IPV4 addresses (a bug) and it only worked for the name accepted by if_nametoindex() for IPV6 addresses, which aren't particularly user-friendly. A better way to handle this is to make use of GetAdaptersAddresses() and compare the provided interface name against either the adapter name or the friendly name, similarly to how the non-if_nametoindex windows version of the code handles things.

For IPV4, I've added a block to do this. For IPV6, I've pulled the existing ifdef'd code up to run before attempting to match via if_nametoindex, when either are applicable.

### DIFF
--- a/ACE/ace/SOCK_Dgram.cpp
+++ b/ACE/ace/SOCK_Dgram.cpp
@@ -845,7 +845,6 @@ ACE_SOCK_Dgram::make_multicast_ifaddr6 (ipv6_mreq *ret_mreq,
 #endif /* ACE_LACKS_IF_NAMETOINDEX */
       if (lmreq.ipv6mr_interface == 0)
         return -1;
-
     }
 #else  /* ACE_WIN32 || !ACE_LACKS_IF_NAMETOINDEX */
     ACE_UNUSED_ARG(net_if);
--- a/ACE/ace/SOCK_Dgram.cpp
+++ b/ACE/ace/SOCK_Dgram.cpp
@@ -673,7 +673,7 @@ ACE_SOCK_Dgram::make_multicast_ifaddr (ip_mreq *ret_mreq,
       ACE_INET_Addr interface_addr;
       if (interface_addr.set (mcast_addr.get_port_number (), net_if) == -1)
         {
-#ifdef (ACE_WIN32)
+#if defined (ACE_WIN32)
           IP_ADAPTER_ADDRESSES tmp_addrs;
           // Initial call to determine actual memory size needed
           ULONG bufLen = 0;
--- a/ACE/ace/SOCK_Dgram.cpp
+++ b/ACE/ace/SOCK_Dgram.cpp
@@ -788,7 +788,6 @@ ACE_SOCK_Dgram::make_multicast_ifaddr6 (ipv6_mreq *ret_mreq,
 
       IP_ADAPTER_ADDRESSES tmp_addrs;
       // Initial call to determine actual memory size needed
-      DWORD dwRetVal;
       ULONG bufLen = 0;
       char *buf = 0;
       if (::GetAdaptersAddresses (AF_INET6, 0, 0, &tmp_addrs, &bufLen)
@@ -799,7 +798,7 @@ ACE_SOCK_Dgram::make_multicast_ifaddr6 (ipv6_mreq *ret_mreq,
 
       // Get required output buffer and retrieve info for real.
       PIP_ADAPTER_ADDRESSES pAddrs = reinterpret_cast<PIP_ADAPTER_ADDRESSES> (buf);
-      if ((dwRetVal = ::GetAdaptersAddresses (AF_INET6, 0, 0, pAddrs, &bufLen)) != NO_ERROR)
+      if (::GetAdaptersAddresses (AF_INET6, 0, 0, pAddrs, &bufLen) != NO_ERROR)
         {
           pAddrs = 0;
         }
--- a/ACE/ace/SOCK_Dgram.cpp
+++ b/ACE/ace/SOCK_Dgram.cpp
@@ -677,30 +677,23 @@ ACE_SOCK_Dgram::make_multicast_ifaddr (ip_mreq *ret_mreq,
           IP_ADAPTER_ADDRESSES tmp_addrs;
           // Initial call to determine actual memory size needed
           ULONG bufLen = 0;
-          if ((::GetAdaptersAddresses (AF_INET,
-                                       0,
-                                       0,
-                                       &tmp_addrs,
-                                       &bufLen)) != ERROR_BUFFER_OVERFLOW)
-            return -1; // With output bufferlength 0 this can't be right.
+          if (::GetAdaptersAddresses (AF_INET, 0, 0, &tmp_addrs, &bufLen)
+              != ERROR_BUFFER_OVERFLOW)
+            {
+              return -1; // With output bufferlength 0 this can't be right.
+            }
 
           // Get required output buffer and retrieve info for real.
           char *buf = 0;
-          ACE_NEW_RETURN (buf,
-                          char[bufLen],
-                          -1);
+          ACE_NEW_RETURN (buf, char[bufLen], -1);
           PIP_ADAPTER_ADDRESSES pAddrs = reinterpret_cast<PIP_ADAPTER_ADDRESSES> (buf);
-          if ((::GetAdaptersAddresses (AF_INET,
-                                       0,
-                                       0,
-                                       pAddrs,
-                                       &bufLen)) != NO_ERROR)
+          if (::GetAdaptersAddresses (AF_INET, 0, 0, pAddrs, &bufLen) != NO_ERROR)
             {
               delete[] buf; // clean up
               return -1;
             }
 
-          interface_addr = ACE_INET_Addr(); // initialize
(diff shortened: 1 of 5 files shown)


ACE/ace/SOCK_Dgram.cpp
```diff
@@ -672,7 +672,63 @@ ACE_SOCK_Dgram::make_multicast_ifaddr (ip_mreq *ret_mreq,
...
+          if ((dwRetVal = ::GetAdaptersAddresses (AF_INET,
+                                                  0,
+                                                  0,
+                                                  &tmp_addrs,
+                                                  &bufLen)) != ERROR_BUFFER_OVERFLOW)
+            return -1; // With output bufferlength 0 this can't be right.
+
+          // Get required output buffer and retrieve info for real.
+          PIP_ADAPTER_ADDRESSES pAddrs;
+          char *buf;
```
jwillemsen, line 690: initialise with 0

ACE/ace/SOCK_Dgram.cpp
```diff
@@ -672,7 +672,63 @@ ACE_SOCK_Dgram::make_multicast_ifaddr (ip_mreq *ret_mreq,
...
+              delete[] buf; // clean up
+              return -1;
+            }
+
+          interface_addr = ACE_INET_Addr(); // initialize
+          int set_result = -1;
+          while (pAddrs && set_result == -1)
+            {
+              if (ACE_OS::strcmp(ACE_TEXT_ALWAYS_CHAR(net_if), pAddrs->AdapterName) == 0 ||
+                  ACE_OS::strcmp(ACE_TEXT_ALWAYS_CHAR(net_if), ACE_Wide_To_Ascii(pAddrs->FriendlyName).char_rep()) == 0)
```
jwillemsen, line 710: Why compare two wchar back to char to compare, why not compare as wchar?

ACE/ace/SOCK_Dgram.cpp
```diff
@@ -746,31 +798,33 @@ ACE_SOCK_Dgram::make_multicast_ifaddr6 (ipv6_mreq *ret_mreq,
       // Initial call to determine actual memory size needed
       DWORD dwRetVal;
       ULONG bufLen = 0;
+      char *buf;
```
jwillemsen, line 801: initialise with 0, then the = 0 on line 808 can go away

jwillemsen:
Any test extensions possible to test this?

ACE/ace/SOCK_Dgram.cpp
simpsont-oci: I copied & modified most of this chunk of code from make_multicast_ifaddr6, but I agree, it could be better. If the values being copied had been the same, I would have made a separate function to avoid duplication. I'll fix these issues in both locations.

ACE/ace/SOCK_Dgram.cpp
simpsont-oci: I'll just initialize during the the actual cast and remove this line.

ACE/ace/SOCK_Dgram.cpp
simpsont-oci: And then I can invert the condition. Gotcha.

ACE/ace/SOCK_Dgram.cpp
```diff
@@ -746,39 +796,35 @@ ACE_SOCK_Dgram::make_multicast_ifaddr6 (ipv6_mreq *ret_mreq,
...
-                      -1);
-      pAddrs = reinterpret_cast<PIP_ADAPTER_ADDRESSES> (buf);
+      PIP_ADAPTER_ADDRESSES pAddrs = reinterpret_cast<PIP_ADAPTER_ADDRESSES> (buf);
       if ((dwRetVal = ::GetAdaptersAddresses (AF_INET6,
                                               0,
                                               0,
                                               pAddrs,
                                               &bufLen)) != NO_ERROR)
         {
-          delete[] buf; // clean up
```
jwillemsen, line 769: Leak of buf?

simpsont-oci: It only needed to be deleted here because of the return on the following line. Since it will fall through to the final delete (line 790) now, there's no need to delete it here.

ACE/ace/SOCK_Dgram.cpp
```diff
@@ -672,7 +672,61 @@ ACE_SOCK_Dgram::make_multicast_ifaddr (ip_mreq *ret_mreq,
...
+                                       0,
+                                       &tmp_addrs,
+                                       &bufLen)) != ERROR_BUFFER_OVERFLOW)
+            return -1; // With output bufferlength 0 this can't be right.
+
+          // Get required output buffer and retrieve info for real.
+          char *buf = 0;
+          ACE_NEW_RETURN (buf,
+                          char[bufLen],
+                          -1);
```
mitza-oci, line 691:
```suggestion
          ACE_NEW_RETURN (buf, char[bufLen], -1);
```

ACE/ace/SOCK_Dgram.cpp
```diff
@@ -672,7 +672,61 @@ ACE_SOCK_Dgram::make_multicast_ifaddr (ip_mreq *ret_mreq,
...
+                                       0,
+                                       0,
+                                       pAddrs,
+                                       &bufLen)) != NO_ERROR)
+            {
+              delete[] buf; // clean up
+              return -1;
+            }
+
+          interface_addr = ACE_INET_Addr(); // initialize
```
mitza-oci, line 703:
```suggestion
          interface_addr = ACE_INET_Addr ();
```

ACE/ace/SOCK_Dgram.cpp
```diff
@@ -672,7 +672,61 @@ ACE_SOCK_Dgram::make_multicast_ifaddr (ip_mreq *ret_mreq,
...
+              pAddrs = pAddrs->Next;
+            }
+
+          delete[] buf; // clean up
+          if (set_result == -1)
+            {
+              return -1;
+            }
+#else
+          return -1;
```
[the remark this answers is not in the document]
simpsont-oci: I added an ACE_NOTSUP_RETURN here, which should set errno to ENOTSUP and (depending on configuration) create a log message. I also set errno to EINVAL manually when the interface was specified by the caller, but couldn't be found / set correctly.

ACE/ace/SOCK_Dgram.cpp
```diff
@@ -788,10 +818,23 @@ ACE_SOCK_Dgram::make_multicast_ifaddr6 (ipv6_mreq *ret_mreq,
...
+      if (lmreq.ipv6mr_interface == 0)
+        {
+          lmreq.ipv6mr_interface = ACE_OS::if_nametoindex (ACE_TEXT_ALWAYS_CHAR (net_if));
+        }
+
+#endif /* ACE_LACKS_IF_NAMETOINDEX */
+      if (lmreq.ipv6mr_interface == 0) {
+        errno = EINVAL;
+        return -1;
+      }
```
mitza-oci, line 833:
```suggestion
      if (lmreq.ipv6mr_interface == 0)
        {
          errno = EINVAL;
          return -1;
        }
```

ACE/ace/SOCK_Dgram.cpp
```diff
@@ -729,56 +777,38 @@ ACE_SOCK_Dgram::make_multicast_ifaddr6 (ipv6_mreq *ret_mreq,
...
-    {
-      lmreq.ipv6mr_interface = ACE_OS::if_nametoindex (ACE_TEXT_ALWAYS_CHAR (net_if));
-    }
-#elif defined (ACE_WIN32)
+#if defined (ACE_WIN32) || !defined (ACE_LACKS_IF_NAMETOINDEX)
   if (net_if != 0)
     {
+#if defined (ACE_WIN32)
       int if_ix = 0;
       bool num_if =
```
jwillemsen, line 785:
``` suggestion
      bool const num_if
```

ACE/ace/SOCK_Dgram.cpp
```diff
@@ -729,56 +777,38 @@ ACE_SOCK_Dgram::make_multicast_ifaddr6 (ipv6_mreq *ret_mreq,
...
-      lmreq.ipv6mr_interface = ACE_OS::if_nametoindex (ACE_TEXT_ALWAYS_CHAR (net_if));
-    }
-#elif defined (ACE_WIN32)
+#if defined (ACE_WIN32) || !defined (ACE_LACKS_IF_NAMETOINDEX)
   if (net_if != 0)
     {
+#if defined (ACE_WIN32)
       int if_ix = 0;
-      bool num_if =
+      bool const num_if
```
mitza-oci, line 785:
```suggestion
      bool const num_if =
```
we will change this line 2 characters at a time until it works

simpsont-oci: I just wanted you both to get an opportunity to get your name on a commit.

jwillemsen:
New compile errors in http://scoreboard.ociweb.com/doc_master_weevil_win2003_vc71_s1d1o0/index.html and https://buildlogs.remedy.nl/win_mingw64_64_tao/index.html

PR #1172, opened 2020-08-04
Removed define of ACE_LACKS_AUTO_PTR for C++17 in general, gcc doesn'…

jwillemsen:
…t remove auto_ptr with C++17 so we have to do this specifically for a compiler who does remove std::auto_ptr with C++17

    * ACE/ace/Auto_Ptr.h:


jwillemsen:
This reverts #1123

PR #1179, opened 2020-08-10
Do not redefine 64-bit formats for MacOS with C++11

### LINKED ISSUE #1173: macOS warnings
I guess this is based on the recent macOS basic types changes (@sonndinh @jrw972)

```
TAO/TAO_IDL/be/be_helper.cpp:374:20: warning: format specifies type 'unsigned long' but the argument has type 'ACE_CDR::ULongLong' (aka 'unsigned long long') [-Wformat]
```
There are a few instances of the same warning

Apple clang version 11.0.3 (clang-1103.0.32.62)
with -std=c++11

jrw972:
Fixes #1173

### DIFF
--- a/.github/workflows/macosx.yml
+++ b/.github/workflows/macosx.yml
@@ -10,12 +10,12 @@ jobs:
   build:
     strategy:
       matrix:
-        cxxstd: [3, 11]
+        cxxstd: ["03", "11"]
         include:
           - os: macos-10.15
             platform_file: include $(ACE_ROOT)/include/makeinclude/platform_macosx.GNU
     runs-on: ${{ matrix.os }}
-    name: "${{ matrix.os }}-${{ matrix.cxxstd }}"
+    name: "${{ matrix.os }}-C++${{ matrix.cxxstd }}"
     env:
       ACE_ROOT: ${{ github.workspace }}/ACE
       TAO_ROOT: ${{ github.workspace }}/TAO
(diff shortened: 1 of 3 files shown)


mitza-oci:
@jwillemsen Should we update GitHub Actions for macOS to use c++11?  If so, do you want to do it as a separate "check" or just change the existing one?

jwillemsen:
What is the C++ default in macOS, is that C++03? Just change the existing one when the default is lower as C++11

mitza-oci:
Defaults to C++03:
```
$ clang++ -v
Apple clang version 11.0.0 (clang-1100.0.33.17)
Target: x86_64-apple-darwin19.2.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
$ clang++ -x c++ /dev/null -E -dM | grep __cplusplus
#define __cplusplus 199711L
```

jwillemsen:
Given the short time of the macOSX build maybe add C++11 as separate configuration, not that we break older support

.github/workflows/macosx.yml
```diff
@@ -10,11 +10,12 @@ jobs:
   build:
     strategy:
       matrix:
+        cxxstd: [3, 11]
         include:
           - os: macos-10.15
             platform_file: include $(ACE_ROOT)/include/makeinclude/platform_macosx.GNU
     runs-on: ${{ matrix.os }}
-    name: ${{ matrix.os }}
+    name: "${{ matrix.os }}-${{ matrix.cxxstd }}"
```
jwillemsen, line 18: Maybe add C++ in the name before the number, that way someone can more easily determine what 3/11 mean, use at that moment 03 instead of 3

Issue #940, opened 2019-08-14
Use git branches instead of tags for Latest_*


mitza-oci:
https://github.com/DOCGroup/ACE_TAO/blob/9d0c34b22345878e61a5765bbd7a48c6791ffe09/ACE/bin/make_release.py#L524

We found a case where Jenkins's git polling didn't detect a change in Latest_*.  Since tags aren't expected to "move," we propose making these branches instead.

jwillemsen:
Feel free to update the script to use branches instead of tags, please make the same changes to CIAO and DAnCE (they use a slightly modified version of the same script).

mitza-oci:
This is causing problems on some of our Jenkins builds.
```
   4d7f65535e..1282610faa  master        -> origin/master
 * [new tag]               ACE+TAO-6_5_8 -> ACE+TAO-6_5_8
 ! [rejected]              Latest_Beta   -> Latest_Beta  (would clobber existing tag)
 ! [rejected]              Latest_Micro  -> Latest_Micro  (would clobber existing tag)
```

jwillemsen:
Same issue now happens according to @mitza-oci for the Latest_ACETAO_Micro tag on MPC

iguessthislldo:
Okay I have the release script setup and working (Or at least [it looks like I did](https://github.com/iguessthislldo-test-org/ACE_TAO/releases)). I got it working with Python 3 along the way.

I have a few questions:
- I'm a little confused why someone why would want to use "Latest_Major" or "Latest_Micro"? Why not just "latest"?
- What am I doing exactly? Am I just making the tags into branches? Or am I keeping the tags and just adding branches.
> please make the same changes to CIAO and DAnCE (they use a slightly modified version of the same script).

- Do each of them have a separate copy, or share one?

jwillemsen:
Latest_Major would be 6.0.0, Latest Minor 6.5.0 and Latest Micro 6.5.10. Personally I always use the latest micro but I know of users who prefer a minor or even major version. That is mostly also history but let us just keep it.

CIAO/DAnCE have their own script.

jwillemsen:
As far as I understood it the proposal from @mitza-oci is to replace the branches with tags because tags are assumed to move where branches are not.

iguessthislldo:
> Personally I always use the latest micro but I know of users who prefer a minor or even major version. That is mostly also history, in the past we had days with multiple beta (micro) releases a day because of basic compile problems due to lack of CI.

[...]

PR #1182, opened 2020-08-12
make_release.py: Use Branches for Latest 

iguessthislldo:
Part of Fix for #940

Assume `Latest_Major`, `Latest_Minor`, and `Latest_Micro`/`Latest_Beta` and the MPC equivalents are branches instead of tags.

The other part of the fix is actually changing the tags into branches.

Also updated the script to work with Python 3 (it still should work with Python 2.7 though), and tried to do some future proofing, mostly by wrapping the argument parser, because `optparse` is deprecated.

Also did some clean up. Simplifying some code, and using more normal Python patterns.

Use of non read-only global variables should be eliminated, because it's unusual to use them in Python, at least nowadays. I started to do this, but back tracked, because it was taking longer to do so than I would've liked to spend on it.

### DIFF
--- a/ACE/bin/make_release.py
+++ b/ACE/bin/make_release.py
@@ -15,6 +15,9 @@ import shlex
 import multiprocessing
 import sys
 import enum
+import os
+from os.path import join
+import shutil
 
 ##################################################
 #### Global variables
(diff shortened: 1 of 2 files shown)


iguessthislldo:
On the Codacy and Code Factor errors:
The only thing they're complaining about that I wrote is assigning to `input` which is necessary for working with Python 2 and 3.
The thing about "len(SEQUENCE)" in Codacy is strange because it's the same thing, the only benefit of removing `len` is that it is shorter and maybe you can set the variable to `None` and it could handle that. Everything else has at least some degree of merit generally, but isn't that important in the context of a script like this.

ACE/bin/make_release.py
```diff
@@ -584,30 +608,14 @@ def push ():
...
-                push_latest_tag ("ACE_TAO", "Beta", tagname)
-                push_latest_tag ("ACE_TAO", "Micro", tagname)
-                push_latest_tag ("MPC", "ACETAO_Minor", tagname)
-                push_latest_tag ("MPC", "ACETAO_Micro", tagname)
-            elif opts.release_type == "micro":
-                push_latest_tag ("ACE_TAO", "Beta", tagname)
-                push_latest_tag ("ACE_TAO", "Micro", tagname)
-                push_latest_tag ("MPC", "ACETAO_Micro", tagname)
+            # Push release branches
+            latest_branch_helper (push_latest_branch, opts.release_type)
```
jwillemsen, line 612: At the moment we release a major also the minor/micro branches needs to be updated, similar for a new minor, that is also a new micro

iguessthislldo: Yes, that's what the [`latest_branch_helper` function](https://github.com/DOCGroup/ACE_TAO/pull/1182/files#diff-b13c989b2269a0d6d0f1bc09744a6f70R550) is for:

```python
>>> def latest_branch_helper (fn, release_type):
...     release_types = ("major", "minor", "micro")
...     do = release_types[release_types.index(release_type):]
...     if "micro" in do:
...         fn ("ACE_TAO", "Beta")
...         fn ("ACE_TAO", "Micro")
...         fn ("MPC", "ACETAO_Micro")
...     if "minor" in do:
...         fn ("ACE_TAO", "Minor")
...         fn ("MPC", "ACETAO_Minor")
...     if "major" in do:
...         fn ("ACE_TAO", "Major")
...         fn ("MPC", "ACETAO_Major")
... 
>>> latest_branch_helper(print, "micro")
ACE_TAO Beta
ACE_TAO Micro
MPC ACETAO_Micro
>>> latest_branch_helper(print, "minor")
ACE_TAO Beta
ACE_TAO Micro
MPC ACETAO_Micro
ACE_TAO Minor
MPC ACETAO_Minor
>>> latest_branch_helper(print, "major")
ACE_TAO Beta
ACE_TAO Micro
MPC ACETAO_Micro
ACE_TAO Minor
MPC ACETAO_Minor
ACE_TAO Major
MPC ACETAO_Major

```

iguessthislldo:
@jwillemsen these are the two kinds of commands I've been using to test the script:

```
python3 ../doc/ACE_TAO/ACE/bin/make_release.py \
  --major \
  --root git@github.com:iguessthislldo-test-org/ACE_TAO.git \
  --mpc_root git@github.com:iguessthislldo-test-org/MPC.git \
  --verbose \
  --update \
  --tag \
  --push

python3 ../doc/ACE_TAO/ACE/bin/make_release.py \
  --verbose \
  --kit
```

I have tested switching `--major` to `--minor` and `--micro`. They seem to be working properly as far I can tell. Is there any other combination of arguments you think I should test out?

jwillemsen:
Commands I use are all listed in https://github.com/DOCGroup/ACE_TAO/blob/master/ACE/docs/bczar/bczar.html.

jwillemsen:
Is python3 adviced? If so, please update bczar.html, from there I copy the commands

iguessthislldo:
> Commands I use are all listed in https://github.com/DOCGroup/ACE_TAO/blob/master/ACE/docs/bczar/bczar.html.

I wasn't aware of that file. The arguments I figured out from trial and error and reading the script are the same as what bczar.html says to run though.

> Is python3 adviced? If so, please update bczar.html, from there I copy the commands

It's not required. I made it compatible with Python 3+, but also kept compatibility with at least 2.7, which I also tested with. I can update bczar.html to say that just to be safe.

jwillemsen:
Please update the script and bczar.html to use and mention python3, that should work on my opensuse box, no need for backward compatibility for this script which is only used forca release

iguessthislldo:
Okay, then do you want me to rip out the compatibility wrappers/shims for Python 2? I can also change the shebang in the script, so the commands in the bczar stay the same, but I would still mention it requires Python 3.

jwillemsen:
Personally I would like to keep it simple, so drop python2 support, what do you think @mitza

mitza-oci:
> 
> 
> Personally I would like to keep it simple, so drop python2 support, what do you think @mitza-oci

I'm OK with dropping support for Python 2

iguessthislldo:
Alright, it's not much code to support it, but I'm fine with dropping it.

jwillemsen:
I would propose to release 6.5.11 soon after you have implemented this so that we have a real release tested with all changes fresh in memory

iguessthislldo:
Alright, then I think it's ready to be review and merged. After that I will create the branches here and in MPC and move on to doing the same to DAnCE and CIAO.

PR #1184, opened 2020-08-13
Export TAO::ORB::default_svc_conf_entries so that it can be used according to t…

jwillemsen:
…he documentation, fixes debian bug #628571

    * TAO/docs/configurations.html:
    * TAO/tao/TAO_Internal.h:


jwillemsen:
Fixed https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=628571

PR #1197, opened 2020-08-14
make_release.py: Fix Use of "is" with Literals

iguessthislldo:


iguessthislldo:
The problem is that `is`/`is not` is for object identity, while `==`/`!=` is for equality. At first I thought this code could never have worked, but then I realized it actually did, but only because in CPython within a range around 0, ints of the same value are actually the same Python object.

```python
>>> x = 10345
>>> y = 10345
>>> x is y
False
>>> x = -1
>>> y = -1
>>> x is y
True
>>> x = 5
>>> y = 5
>>> x is y
True
```

iguessthislldo:
> What about the invalid PRF form update?

Oh, I missed that part of the email... I will look into that.

iguessthislldo:
I think I only checked VERSION.txt or Version.h files if I checked them all, but my test repo does say "7.0.00" in it's PRF.

I isolated the code that updates the version in the PRF, and found it only does this if the exiting minor version is more than one digit. This existing regex `(\d\.)+\d` is not handling that situation correctly. I will replace it with a real version regex. Yet again this something I didn't change, so I'm not sure how it worked before.

Issue #1201, opened 2020-08-18
Problems with fragmented GIOP messages


j18ter:
### Version

6.5.11/2.5.11

### Host machine and operating system

CentOS 7 Linux 3.10.0-1127.13.1.el7.x86_64 on a generic PC with Intel i7 CPU. 

### Target machine and operating system (if different from host)

Same

### Compiler name and version (including patch level)

gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-39)

### The $ACE_ROOT/ace/config.h file

#define ACE_HAS_IPV6
#include <ace/config-linux.h>

### The $ACE_ROOT/include/makeinclude/platform_macros.GNU file

include $(ACE_ROOT)/include/makeinclude/platform_linux.GNU

### Contents of $ACE_ROOT/bin/MakeProjectCreator/config/default.features

This file does not seem to exist.

### AREA/CLASS/EXAMPLE AFFECTED:

A simple variation of the echo example, which returns a string sequence that is slightly bigger
than 4MB. The server is started with command line parameters ```-ORBMaxMessageSize 1048576```
to cause GIOP fragmentation.

### The problem effects:

It affects execution. In this example the client receives incorrect values, in our more complex
application the client encounters ```CORBA::MARSHAL``` or ```CORBA::COMM_FAILURE``` exceptions.

### Synopsis

Returning big sequences from operations can cause client exceptions or incorrect values.

### Description

Initially seen with a total returned size of approximately 8 MB, and client got ```CORBA::MARSHAL```, The ``-ORBMaxMessageSize"
parameter was not used in this case. Using it to force fragmentation causes simple examples to fail.

### Repeat by

A complete example program with MPC and make files can be cloned from Github with this command:

```git clone https://github.com/NetAcquire/tao_fragmentation_bug.git```

Note: Run the included ```run_tests.pl``` script to see the failure. This passes ```-ORBMaxMessageSize 1048576```
to the server. Without this parameter the test succeeds.

### Sample fix/ workaround

Tried to debug, but got lost in the GIOP parsing code.

jwillemsen:
Thanks for reporting this issue and making a reproducer. Could you attach a client and server logfile with ORB logging 10?

[...]

PR #1205, opened 2020-08-19
Repro problem with fragmented GIOP messages

j18ter:
This is a test to reproduce issue #1201. Not sure what needs to
be done to integrate this with the rest of the build, if anything.

Note, that I'm referencing a utility file from the examples
directory, not sure if this is frowned upon?

A free-standing variant of this test, which I was able to build and
test is here: [TAO Fragmentation Bug](https://github.com/NetAcquire/tao_fragmentation_bug)


jwillemsen:
I would prefer not to use an example file from a completely different location

TAO/tests/GIOP_Fragments/Big_String_Sequence/echo.mpc
```diff
@@ -0,0 +1,34 @@
+// -*- MPC -*-
+project(*idl): taoidldefaults {
+  IDL_Files {
+    Echo.idl
+  }
+  custom_only = 1
+}
+
+project(*Server): taoserver, namingexe, iortable, utils, avoids_corba_e_micro {
```
jwillemsen, line 9: base projects need cleanup, way too much looking briefly at the test code

j18ter: The only base project I was able to remove without causing the build to fail was `avoids_corba_e_micro`. Note,
this whole example is just a copy of `examples/Simple/echo` with some tweaks to reproduce this bug. We don't
use MPC in our own projects and I don't know enough about MPC to do more than copy existing stuff with small
and obvious adjustments. I've pushed a change to remove  `avoids_corba_e_micro`.

j18ter:
> I would prefer not to use an example file from a completely different location

Unfortunately, this `Simple_util.cpp` file can't be dropped without completely rewriting the application. It could
be redundantly checked-in in the project directory, but this seems worse to me.

TAO/tests/GIOP_Fragments/Big_String_Sequence/server.cpp
```diff
@@ -0,0 +1,42 @@
+#include "../../../examples/Simple/Simple_util.h"
+#include "Echo_i.h"
+
+// This is the main driver program for the time and date server.
```
jwillemsen, line 4: old comment

jwillemsen:
Please don't use std namespace, it isn't fully supported on all ACE/TAO supported platforms

jwillemsen:
Related to #1239

PR #1216, opened 2020-08-24
Migrate several custom travis configurations to github actions

jwillemsen:
* .github/workflows/linux.yml:
    * .travis.yml:


jwillemsen:
Ubuntu 20.04 has a problem creating a file using powershell https://github.com/actions/virtual-environments/issues/1475

mitza-oci:
> 
> 
> Ubuntu 20.04 has a problem creating a file using powershell [actions/virtual-environments#1475](https://github.com/actions/virtual-environments/issues/1475)

I will offer to remove all PowerShell from the repo.

jwillemsen:
The advantage of powershell would be that we could more easily copy build steps between the various scripts. Looks 20.04 is not ready for this yet. For the moment I replaced all 20.04 with 18.04.

jwillemsen:
Ok, found that I had a different problem in my changes, so I am trying 20.04 again, the error printed in the detailed log isn't flagged as error, maybe it works enough for now

jwillemsen:
Combining wchar with inline doesn't work, see https://github.com/DOCGroup/ACE_TAO/issues/1217

PR #1240, opened 2020-09-17
Removed manual narrow support in favor of C++ dynamic_cast

jwillemsen:


iguessthislldo:
I'm fine with removing calls to the narrows, but it's going to be a least a little annoying to people extending the compiler if the narrows themselves are removed.

jwillemsen:
> I'm fine with removing calls to the narrows, but it's going to be a least a little annoying to people extending the compiler if the narrows themselves are removed.

Removing them from ciao/opendds was straightforward and easy and by replacing it with dynamic_cast works also with older versions. Looks something people could do easily, we should have done this a decado ago ;-)

mitza-oci:
Does this mean an older version of a downstream project can't be built with a newer version of TAO_IDL_FE?

jwillemsen:
It would build when they use dynamic_cast instead of narrow_from_decl/narrow_from_scope as they should (and as we should have done for a long time).

jwillemsen:
I did the cleanup for OpenDDS and CIAO, separate PRs created for that

mitza-oci:
> 
> 
> It would build when they use dynamic_cast instead of narrow_from_decl/narrow_from_scope as they should (and as we should have done for a long time).

Seems like this will be a problem.  Could TAO_IDL_FE maintain compatibility?

jwillemsen:
I could put macros back in a lot of files but that would mean people just keep using them forever and we have to maintain them forever. I already have updated CIAO/OpenDDS, the effort was less then an hour for both. At the moment someone upgrades to a new TAO release he could do this easily.

mitza-oci:
I think the costs of having some extra lines like `IMPL_NARROW_FROM_DECL(AST_Decl)` in the code are very minimal, and we've been fine with those costs over the last few decades.  Carrying these forward doesn't seem to be preventing any other maintenance efforts.  Doing so allows easier compatibility for any downstream project -- we know there are more than just CIAO and OpenDDS.

An example scenario with OpenDDS would be someone downloading the 3.14 zip or git tag and configuring with one of the options like `--ace-github-latest`.  They are an end-user and not a maintainer so they're not expected to edit the code themselves.

jwillemsen:
I will prepare a new PR with the macros back, but dynamic_cast in the code, at a later moment in time we can cleanup the macros so we can keep this PR open for some time

PR #1242, opened 2020-09-18
Remove usage of narrow_from_decl/narrow_from_scope and replace them with dynamic_cast

jwillemsen:


jwillemsen:
Part of the cleanup of #1240, the remaining cleanup of that PR will be done later

PR #1244, opened 2020-10-02
Use portable perl shebang

jwillemsen:
https://stackoverflow.com/questions/10376206/what-is-the-preferred-bash-shebang

### DIFF
--- a/ACE/ace/SPIPE_Addr.cpp
+++ b/ACE/ace/SPIPE_Addr.cpp
@@ -7,8 +7,6 @@
 #include "ace/SPIPE_Addr.inl"
 #endif /* __ACE_INLINE__ */
 
-
-
 ACE_BEGIN_VERSIONED_NAMESPACE_DECL
 
 ACE_ALLOC_HOOK_DEFINE(ACE_SPIPE_Addr)
--- a/ACE/bin/PerlACE/ConfigList.pm
+++ b/ACE/bin/PerlACE/ConfigList.pm
@@ -1,4 +1,4 @@
-#! /usr/bin/perl
+#!/usr/bin/env perl
 
 package PerlACE::ConfigList;
 use strict;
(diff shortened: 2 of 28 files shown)


jwillemsen:
Coming from https://github.com/conda-forge/staged-recipes/pull/12698

mitza-oci:
> 
> 
> https://stackoverflow.com/questions/10376206/what-is-the-preferred-bash-shebang

I'd be interested to hear from @ocielliottc since this impacts MPC and he knows more about these things.

I don't really have any objections but I'm wondering why this has never been a problem before?

There are some comments here https://unix.stackexchange.com/questions/538844/does-posix-guarantee-the-paths-to-any-standard-utilities stating that depending on `env` isn't necessarily correct either, but I'm not enough of an expert to say if any of this matters for us.

jwillemsen:
> I don't really have any objections but I'm wondering why this has never been a problem before?

See https://github.com/conda-forge/staged-recipes/pull/12698#issuecomment-702757492, looks they have a build setup where perl is not /usr/bin/perl but it is added to the path

ocielliottc:
Since perl is in their path, could they not just prefix running the script with perl?

jwillemsen:
> Since perl is in their path, could they not just prefix running the script with perl?

With prj_install.pl that will not work because it is ran through the generated GNU makefiles and it could be override to be a different script which could even be a non-perl one (see https://github.com/DOCGroup/ACE_TAO/blob/a2887097835b629126a011d035e22f366ccdf79e/ACE/bin/MakeProjectCreator/templates/gnu.mpd#L876)

ocielliottc:
Ah, ok.  I guess I'm ok with this then.  It doesn't seem very likely that this will be running on any unix machines that don't have env in /usr/bin.

jwillemsen:
> Ah, ok. I guess I'm ok with this then. It doesn't seem very likely that this will be running on any unix machines that don't have env in /usr/bin.

Yes, the stackoverflow article that I linked to has a lot of upvotes on this proposal. When all CI builds are ready I will merge the ones I have permission for

PR #1268, opened 2020-10-21
Updates for Windows-hosted Android NDK cross-compile

mitza-oci:


ACE/include/makeinclude/wrapper_macros.GNU
```diff
@@ -315,7 +315,7 @@ endif # ! COMPSPEC
 
 ifdef COMSPEC
   #### Assume we're on a WIN32 host.
-  ACE_NUL = nul
+  ACE_NUL = $(if $(findstring /bin/sh,$(SHELL)),/dev/null,nul)
```
iguessthislldo, line 318: I feel like there could be a better check than relying on `$SHELL`, but I'm not sure what that would look like. Maybe something specific to MSYS/MinGW? It's probably good enough though.

iguessthislldo (approved):

ACE/include/makeinclude/wrapper_macros.GNU
mitza-oci: `$(SHELL)` is in the GNU Make docs, so that's the best I could come up with.  Certainly open to other ideas.
For MSYS2, the value reported by `make -p` is:
```
SHELL := C:/msys64/usr/bin/sh.exe
```

PR #1292, opened 2020-11-10
Give an error when we don't have C++11 enabled, this is mandatory for ACE7/TAO3

jwillemsen:


jwillemsen:
clang5 and gcc4 do have c++11 now as explicit default, after merging this PR compilers without C++11 support will not work anymore on master

shuston (approved):

mitza-oci (approved):

PR #1294, opened 2020-11-10
Remove checks for ACE_HAS_CPP11 and remove the code for C++03 now that we require C++11 support

jwillemsen:


ACE/ace/Global_Macros.h
```diff
@@ -61,24 +61,11 @@
...
-    // It just evaporated ;-)  Not pleasant.
-#   define ACE_UNIMPLEMENTED_FUNC(f)
-# else
-#   if defined (ACE_HAS_CPP11)
-#     define ACE_UNIMPLEMENTED_FUNC(f) f = delete;
-#   else
-#     define ACE_UNIMPLEMENTED_FUNC(f) f;
-#   endif
-# endif /* ACE_NEEDS_FUNC_DEFINITIONS */
+# define ACE_UNIMPLEMENTED_FUNC(f) f = delete;
```
iguessthislldo, line 64: I'm guessing we might want to keep macros like these for compatibility for users, but maybe we should replace these in ACE and TAO for more idiomatic code.

iguessthislldo: If that's already the plan you can ignore me.

jwillemsen: Yes, we keep them for the users but want to replace this globally in ACE/TAO, a lot of cleanup can be done, help is welcome ;-)

PR #1300, opened 2020-11-21
Remove deprecated std::iterator

VemundH:
std::iterator was deprecated in C++17.
This replaces inheritance by instead defining the typedefs earlier provided by std::iterator


jwillemsen:
Thanks!

PR #1316, opened 2020-12-02
Add mingw64 github action

jwillemsen:
* .github/workflows/windows.yml:


jwillemsen:
The compiler installed isn't supported by ACE currently

PR #1318, opened 2020-12-03
Support Building with Android NDK Directly

iguessthislldo:
Instead of having to use a standalone toolchain generated from the NDK.

Also:

- `gnuace` no longer supports Android NDKs before r18. This means only clang is now supported when building for Android. This allows us to remove the hacky GCC support and ensure that the NDK has >=C++11 by default.

- Made it easier to use `gnuace`-built libraries in as imported libraries in CMake-based Android Studio native projects.

- Removed support for old `ANDROID_ARCH` make variable. Replaced `ANDROID_ABI` with `android_abi`, but kept the former as an alias of the later for compatibility. Also will no longer default to 32-bit ARM, so `android_abi` or `ANDROID_ABI` must be defined.

- As recommend by Google, building with neon support is now the default when `android_abi` is `armeabi-v7a`. If support for these processors without NEON extensions is needed, put `androind_neon := 0` in `platform_macros.GNU`.

- As recommend by Google, `gnuace` will start using LLD, the LLVM linker, instead of the default GNU linkers. If the NDK being used doesn't default to LLD and you want to use the GNU linkers, put `androind_set_lld := 0` in `platform_macros.GNU`.

TODO:

- [x] Fix Spelling mistake in NEWS File
- [X] ~Add things to enabled C++11 when needed. Some builds are failing on my matrix because they don't default to C++11.~ I decided to remove support for NDKs without C++11 instead.
- [x] Test on my android matrix
- [X] Update ACE-INSTALL.html
- [x] Create PR with changes appropriate for ACE 6 (Done: #1322)


ACE/ace/config-android.h
```diff
@@ -10,14 +10,13 @@
...
-#if !defined (__ANDROID_API__)
-#  include <android/api-level.h>
-#endif
+#include <android/ndk-version.h>
+#include <android/api-level.h>
 
 #define ACE_ANDROID
 #define ACE_PLATFORM_CONFIG config-android.h
 
-#include "ace/config-linux-common.h"
```
jwillemsen, line 24: This will require the user to have the ace dir in his include path which not all ace users do

ACE/ACE-INSTALL.html
```diff
@@ -1837,14 +1786,19 @@ <h3><a name="android-building">Building</a></h3>
     </div>
   </li>
   <li>Setup the Android build<ul>
-    <li>Create <code>ace/config.h</code>: <code>#include "ace/config-android.h"</code></li>
+    <li>Create <code>ace/config.h</code>: <code>#include "config-android.h"</code></li>
```
jwillemsen, line 1789: ace directory should be here

ACE/ace/config-android.h
```diff
@@ -10,74 +10,66 @@
...
+ * Revision | __NDK_MAJOR__ | __NDK_MINOR__ | __NDK__BETA__
+ * r16      | 16            | 0             | 0
+ * r16b     | 16            | 1             | 0
+ * r16c     | 16            | 2             | 0
+ * r22-beta1| 22            | 0             | 1
+ *
+ * __NDK_BETA__ is starts at 1 and increments until release when it is 0.
  *
  * After r16, NDK version macros are defined in android/ndk-version.h Before
  * that they must be defined in platform_macros.GNU before the include of
```
jwillemsen, line 34: Old comment to remove?

ACE/ace/config-android.h
```diff
@@ -10,74 +10,66 @@
...
-#    error ndk-version.h is missing, __NDK_MINOR__ for Android NDK must be defined!
-#  endif
-#endif
+#define ACE_ANDROID_NDK_LESS_THAN(MAJ, MIN, BET) \
+  !ACE_ANDROID_NDK_AT_LEAST((MAJ), (MIN), (BET))
 
 // ucontext.h and clock_settime() were added in r10c
-#if ACE_ANDROID_NDK_AT_LEAST(10, 2)
+#if ACE_ANDROID_NDK_AT_LEAST(10, 2, 0)
 #  define ACE_HAS_UCONTEXT_T
```
jwillemsen, line 53: Some cleanup possible now that r18 is our minimum?

ACE/include/makeinclude/platform_android.GNU
```diff
@@ -19,92 +22,92 @@ PIE	?= -pie
...
+
+else
+  CROSS_COMPILE := $(CROSS_COMPILE)-
+
+  # Export so child processes can use tools from the same toolchain.
+  export CROSS_COMPILE
 endif
 
-FLAGS_C_CC += -W -Wall -Wpointer-arith
+FLAGS_C_CC += -Wpointer-arith
```
jwillemsen, line 105: Why remove these flags, are used on all platforms

ACE/ace/config-android.h
iguessthislldo: No it doesn't. That's how `#include <>` works, but not how `#include ""` works. `#include ""` first checks to see if the path exists as file relative to the including file, then goes one to check the include path.

ACE/ace/config-android.h
iguessthislldo: Yes, I was going to do that, but it slipped my mind.

ACE/include/makeinclude/platform_android.GNU
iguessthislldo: `platform_clang_common.GNU` includes these flags already.

iguessthislldo:
I'm currently retooling my matrix to build using ACE6 and ACE7, toolchain and NDK. I also will need to add a note that using the NDK directly will require r19 instead of r18 like I thought.

iguessthislldo:
There are at least a few more kinks I have to work out in this PR and in #1322: https://github.com/iguessthislldo/OpenDDS-Android/actions/runs/405156855

ACE/ace/config-android.h
jwillemsen: I would prefer to keep `ace` in the path here for consistency, we do it everywhere else

iguessthislldo: Alternatively we could correct it like I'm currently slowing doing in OpenDDS since it gives a false impression about how `#include` works.

iguessthislldo:
[Builds are now clean](https://github.com/iguessthislldo/OpenDDS-Android/actions/runs/452266694) except a build where iconv is having issues, so that has nothing to do with ACE.

jwillemsen (approved):

PR #1322, opened 2020-12-06
[ACE6] Support Building with Android NDK Directly

iguessthislldo:
Changes in https://github.com/DOCGroup/ACE_TAO/pull/1318 modified for capability's sake for ACE6.

- Made it easier to use `gnuace`-built libraries in as imported libraries in CMake-based Android Studio native projects.

- Replaced `ANDROID_ABI` with `android_abi`, but kept the former as an alias of the later for compatibility.

- Remove need to pass a macro to compiler about `android/ndk-version.h` unless the NDK is old

### DIFF
--- a/ACE/include/makeinclude/platform_g++_common.GNU
+++ b/ACE/include/makeinclude/platform_g++_common.GNU
@@ -253,3 +253,6 @@ pipes ?= 1
 
 FLAGS_C_CC += -Wall -W -Wpointer-arith
 CCFLAGS += -Wnon-virtual-dtor
+
+# Suppress "ar: creating *.a" Message
+ARFLAGS += -c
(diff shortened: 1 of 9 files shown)


iguessthislldo:
https://github.com/DOCGroup/ACE_TAO/blame/master/ACE/tests/Compiler_Features_32_Test.cpp#L46

@jwillemsen so the plan with this is to update this line until the end of time?

jwillemsen:
Did I made an error with the change, if so, please correct @iguessthislldo

iguessthislldo:
I'm including your commit here because the `ace6tao2` branch apparently doesn't have it and that caused an error on a mac build. No you didn't make a mistake, my question was probably a too sarcastic. I'm wondering why ACE tests has `Compiler_Features_32_Test.cpp ` that has to be updated for each Apple Clang release when nothing in ACE or TAO appears to use the workaround it has.

jwillemsen:
It is a construct that we use in the IDL to C++11 union mapping. The issue has been fixed in clang11 so hopefully it will be fixed soon in the next apple clang release

ACE/include/makeinclude/platform_g++_common.GNU
```diff
@@ -253,3 +253,6 @@ pipes ?= 1
 
 FLAGS_C_CC += -Wall -W -Wpointer-arith
 CCFLAGS += -Wnon-virtual-dtor
+
+# Suppress "ar: creating *.a" Message
+ARFLAGS += -c
```
mitza-oci, line 258: This appears to be incompatible with `ar` on older Linux distros (RHEL 6?).

PR #1324, opened 2020-12-06
Remove ACE_OS::readdir_r

iguessthislldo:
[glibc](https://man7.org/linux/man-pages/man3/readdir_r.3.html) and [FreeBSD libc](https://www.freebsd.org/cgi/man.cgi?query=readdir&sektion=3) have both marked `readdir_r` deprecated, mainly because it can't deal with arbitrarily long path names, so we should probably remove it since we have a chance to with ACE7. Also removed `ACE_Dirent::read(ACE_DIRENT *, ACE_DIRENT **)` which used it. `ACE_LACKS_READDIR_R` will now always be defined.


iguessthislldo:
After this I'd like to do the same to `mktemp` and `tmpnam`, which are also deprecated in glibc, though I think those will be harder to remove.

jwillemsen (approved):

PR #1344, opened 2020-12-13
Simplify Includes

iguessthislldo:
Used a script to find places where a preprocessor include starts with the directory of the including file and removed that part of the path to turn it into a relative path.


jwillemsen:
Ok, this looks to work but personally I find it less obvious which file is included, I would prefer to keep the paths there

iguessthislldo:
>  I find it less obvious which file is included

It is missing the other part I'm trying to do in OpenDDS which is just using `#include <>` for non-relative includes. That way its more obvious. This is what's recommend by the [C++ Core Guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rs-incform),

jwillemsen:
> It is missing the other part I'm trying to do in OpenDDS which is just using `#include <>` for non-relative includes. That way its more obvious. This is what's recommend by the [C++ Core Guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rs-incform),

Ok, but when we want to use `#include <some_library/common.h>` for ACE/TAO within the unit tests it looks much easier to keep the ace and tao prefix, that makes a global replace much easier than when you remove ace and tao right now

iguessthislldo:
If I understand you correctly, do you mean for example using `ace/Reactor.h` instead something like of `../../ace/Reactor.h` in a test? If so I agree with you and these changes don't do anything like that. There are no changes outside of `ACE/ace`, `ACE/ACEXML`, `TAO/tao`, and `TAO/orbsvcs`.

jwillemsen:
> If I understand you correctly, do you mean for example using `ace/Reactor.h` instead something like of `../../ace/Reactor.h` in a test? If so I agree with you and these changes don't do anything like that. There are no changes outside of `ACE/ace`, `ACE/ACEXML`, `TAO/tao`, and `TAO/orbsvcs`.

I was thinking about <ace/Reactor.h> in a test, didn't notice you only changed core files, not tests

PR #1357, opened 2020-12-29
tao_idl Improvements

iguessthislldo:
- Made IDL 4 the default.
- Added an IDL preprocessor macro for the IDL version, `__TAO_IDL_IDL_VERSION`.
- Added proper documentation for the `__TAO_IDL`  IDL preprocessor macro.
- Remove support for long deprecated `CPP_LOCATION` and `TAO_IDL_DEFAULT_CPP_FLAGS` environment variables.
- Fixed issue encountered in [this OpenDDS PR](https://github.com/objectcomputing/OpenDDS/pull/2161#issuecomment-750808090) where OpenDDS' `idl2jni`, which an `tao_idl`-derived compiler, can't properly handle relative includes because `tao_idl` makes a temporary copy that is passed to the preprocessor. Fixed this by making it preferred to pass the IDL file to the preprocessor directly.


jwillemsen:
Very likely some old preprocessors had issues when passing them a IDL file directly due to the idl extension, some required it to be c/cpp

TAO/TAO_IDL/driver/drv_args.cpp
```diff
@@ -688,21 +684,66 @@ process_long_option(long ac, char **av, long &i)
...
+      bool invalid_argument = no_more_args;
+      if (no_more_args)
+        {
+          ACE_ERROR ((LM_ERROR,
+            ACE_TEXT ("--preprocessor-input is missing its required argument.")));
+        }
+      else
+        {
+          i++;
+          if (!ACE_OS::strcmp (av[i], "guess"))
```
jwillemsen, line 712: Can use std::strcmp now without problems

TAO/TAO_IDL/driver/drv_preproc.cpp
```diff
@@ -577,11 +578,107 @@ DRV_add_include_path (ACE_CString& include_path,
...
+  if (idl_global->preprocessor_input_ == IDL_GlobalData::PreprocessorInputGuess)
+    {
+      const char * const cpp_name = tolower (our_basename (FE_get_cpp_loc_from_env ()));
+      const char * method = "copy";
+      if (idl_global->compile_flags () & IDL_CF_INFORMATIVE)
+        {
+          ACE_DEBUG ((LM_DEBUG, "%C: guessing preprocessor input method using name: %C\n",
+            idl_global->prog_name (), cpp_name));
+        }
+      if (ACE_OS::strstr (cpp_name, "g++")) // g++ or clang++
```
jwillemsen, line 626: can use std::strstr

TAO/TAO_IDL/driver/drv_preproc.cpp
```diff
@@ -577,11 +578,107 @@ DRV_add_include_path (ACE_CString& include_path,
...
+          idl_global->preprocessor_input_ = IDL_GlobalData::PreprocessorInputDirectWithoutE;
+          method = "direct-without-e (cpp)";
+        }
+      else
+        {
+          // If we don't know it seems like it's safer to fallback to the copy
+          // method. Warn about it anyway, because it shouldn't happen.
+          idl_global->preprocessor_input_ = IDL_GlobalData::PreprocessorInputCopy;
+          if (!(idl_global->compile_flags () & IDL_CF_NOWARNINGS))
+            ACE_ERROR ((LM_WARNING,
```
jwillemsen, line 647: ACE_DEBUG?

TAO/TAO_IDL/driver/drv_preproc.cpp
```diff
@@ -577,11 +578,107 @@ DRV_add_include_path (ACE_CString& include_path,
...
 {
+  char idl_version_arg[128];
+  ACE_OS::sprintf (idl_version_arg, "-D__TAO_IDL_IDL_VERSION=%s",
+    idl_global->idl_version_.to_macro ());
+  DRV_cpp_putarg (idl_version_arg);
+
+  // Decide how files will be passed to the preprocessor
+  if (idl_global->preprocessor_input_ == IDL_GlobalData::PreprocessorInputGuess)
+    {
+      const char * const cpp_name = tolower (our_basename (FE_get_cpp_loc_from_env ()));
```
jwillemsen, line 619: What about setting the default through the various config-* files, there we also set the default preprocessor to be used?

jwillemsen:
bcc64/bcc64x seems to support `-x c++ -E` but bcc32c (which is also clang based) doesn't support -E, there is an explicit Embarcadero message that -E isn't supported `Warning: option '-E' is not supported in Clang-based compiler.` and there is no output

iguessthislldo:
So I decided to download it to see what it's doing (this was annoying because they wanted an account, but didn't even give me the 64-bit compilers...). `bcc32x -x c++ -E` works and that's [the one being used by `tao_idl`](https://github.com/DOCGroup/ACE_TAO/blob/master/ACE/ace/config-win32-borland.h#L53-L59). So if `bcc64` also works, then I think it can use the GCC/Clang direct method without an issue.

iguessthislldo:
To give the current context, I'm trying to see what it would take to get this PR merged because I occasionally see `tao_idl` failing because of the temporary file on Windows: `D:/a/OpenDDS/OpenDDS/OpenDDS/build/bin/Release/opendds_idl.exe: Unable to rename temporary file "C:\Users\RUNNER~1\AppData\Local\Temp\tao-idlf_pTNkVl" to "C:\Users\RUNNER~1\AppData\Local\Temp\tao-idlf_pTNkVl.cpp": Input/output error`

[Log link](https://github.com/OpenDDS/OpenDDS/actions/runs/9421458131/job/25955583946#step:10:4868)

This is probably a problem with GitHubs VMs, but I wanted to do this anyways.

I also noticed it just failed in a job from the merge commit: `D:\a\ACE_TAO\ACE_TAO/ACE\bin\tao_idl: Unable to rename temporary file "C:\Users\RUNNER~1\AppData\Local\Temp\tao-idlf_KzZPyr" to "C:\Users\RUNNER~1\AppData\Local\Temp\tao-idlf_KzZPyr.cpp": Input/output error`

[Log link](https://github.com/DOCGroup/ACE_TAO/actions/runs/9434578453/job/25986878286?pr=1357#step:13:4021)

However because this is in this PR, then something's wrong with my current code and it shouldn't have done this. So the guessing probably isn't working on Windows. As Johnny suggested originally I think I'm going to have it work use macros, at least for the defaults, but I need to think about the specifics.

jwillemsen:
Would be nice when we don't need that copy step anymore with most compilers, will speed up the builds a little bit, especially on Windows

PR #1360, opened 2020-12-29
Revert "Revert "Use unspecified_bool_type instead of bool.""

jwillemsen:
Reverts DOCGroup/ACE_TAO#1227 

Should now not be a problem now Visual Studio 7.1 is not supported anymore for ACE 7

### DIFF
--- a/ACE/ace/Refcounted_Auto_Ptr.h
+++ b/ACE/ace/Refcounted_Auto_Ptr.h
@@ -44,6 +44,10 @@ template <class X, class ACE_LOCK> class ACE_Refcounted_Auto_Ptr;
 template <class X, class ACE_LOCK>
 class ACE_Refcounted_Auto_Ptr
 {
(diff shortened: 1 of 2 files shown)


jwillemsen:
@likema Ok?

likema:
Is the last revert for fixing **VC 7.x** ?

jwillemsen:
No, the last revert puts back your original change, ACE 7 will require C++11 support so we don't have to worry anymore about Visual Studio 7.1

likema:
> No, the last revert puts back your original change, ACE 7 will require C++11 support so we don't have to worry anymore about Visual Studio 7.1

OK

It is inspired by `ace/Intrusive_Auto_Ptr.h`

jwillemsen:
Yes, did see that with the previous merge attempt, Visual Studio 7.1 had issues because the changed operation is not implemented inline as ace/Intrusive_Auto_Ptr.h does, but now this is not an issue anymore.

PR #1363, opened 2020-12-30
Don't print the exception, causes the scoreboard scripts to flag this…

jwillemsen:
… test as error, the exception is expected

    * TAO/tests/ORB_portspan/server.cpp:


jwillemsen:
#1166

PR #1366, opened 2020-12-30
[ACE6] tao_idl Improvements

iguessthislldo:
Backport of the appropriate features of #1357 to TAO 2.

- Added an IDL preprocessor macro for the IDL version, `__TAO_IDL_IDL_VERSION`.
- Added proper documentation for the `__TAO_IDL`  IDL preprocessor macro.
- Fixed issue encountered in [this OpenDDS PR](https://github.com/objectcomputing/OpenDDS/pull/2161#issuecomment-750808090) where OpenDDS' `idl2jni`, which is a `tao_idl`-derived compiler, can't properly handle relative includes because `tao_idl` makes a temporary copy that is passed to the preprocessor. Unlike in #1357 `guess` is not the default for `--preprocessor-input`, but `copy` is, which is the current behavior.


TAO/tests/IDL_Relative_Include/IDL_Relative_Include.mpc
```diff
@@ -0,0 +1,7 @@
+project(*): taoidldefaults {
```
mitza-oci, line 1:
```suggestion
project: taoidldefaults {
```

TAO/docs/compiler.html
```diff
@@ -169,8 +198,47 @@ <h3>Generated Files</h3>
...
+  <tr>
+    <td><code>__TAO_IDL</code></td>
+    <td>
+      <p>The version of ACE that <code>tao_idl</code> was built with. The format
+      is one byte for each version field in hex format. For example, in a
+      theoretical TAO version 3.2.1, which would correspond to ACE version
+      7.3.1, <code>__TAO_IDL</code> would be <code>0x070201</code>.</p>
+
+      <p>It's also possible to include <code>tao/Version.h</code> and use the
+      same macros from that header file as you can use in C++.</p>
```
mitza-oci, line 224: These could be different versions?  They probably shouldn't be though

TAO/docs/compiler.html
```diff
@@ -187,316 +255,327 @@ <h3>Operation Demuxing Strategies</h3>
...
 
+<p>
 TAO_IDL can generate collocated stubs using two different collocation
 strategies.  It also allows you to suppress/enable the generation of
 the stubs of a particular strategy.  To gain great flexibility at
 run-time, you can generate stubs for both collocation strategies
-(using both '<a href="#Gp"><code>-Gp</code></a>' and '<a
-href="#Gd"><code>-Gd</code></a>' flags at the same time) and defer the
+(using both '<a href="#Gp"><code>&#8209;Gp</code></a>' and '<a
+href="#Gd"><code>&#8209;Gd</code></a>' flags at the same time) and defer the
```
mitza-oci, line 295: Why is this changing?

TAO/NEWS
```diff
@@ -1,3 +1,16 @@
...
+
+. TAO IDL:
+
+  . Version of IDL specification being used is available as the
+    `__TAO_IDL_IDL_VERSION` preprocessor macro.
+
+  . Added `--preprocessor-input` option to `tao_idl` to make behavior of
+    `#include ".."` consistent with C and C++. This is not enabled by default
+    like it is in TAO 3, though. For details, see the `--preprocessor-input`
+    argument in `docs/compiler.html`.
```
mitza-oci, line 12: I don't know if the TAO 3 note is needed here.  The scope of this section is TAO 2.

TAO/TAO_IDL/driver/drv_args.cpp
```diff
@@ -688,21 +684,66 @@ process_long_option(long ac, char **av, long &i)
...
+    {
+      bool invalid_argument = no_more_args;
+      if (no_more_args)
+        {
+          ACE_ERROR ((LM_ERROR,
+            ACE_TEXT ("--preprocessor-input is missing its required argument.")));
+        }
+      else
+        {
+          i++;
```
mitza-oci, line 711:
```suggestion
          ++i;
```

TAO/TAO_IDL/driver/drv_args.cpp
```diff
@@ -688,21 +684,66 @@ process_long_option(long ac, char **av, long &i)
...
+              invalid_argument = true;
+              ACE_ERROR ((LM_ERROR,
+                ACE_TEXT ("\"%C\" is not a valid argument to --preprocessor-input.\n"), av[i]));
+            }
+        }
+      if (invalid_argument)
+        {
+          ACE_ERROR ((LM_ERROR, ACE_TEXT ("Use \"guess\", \"direct-with-e\", ")
+            ACE_TEXT ("\"direct-without-e\", \"direct-gcc\", or \"copy\".\n"),
+            av[i]));
```
mitza-oci, line 743: Is there supposed to be a `%C` or similar in the format string?

TAO/TAO_IDL/driver/drv_preproc.cpp
```diff
@@ -577,11 +578,107 @@ DRV_add_include_path (ACE_CString& include_path,
...
+      }
+    return path;
+  }
+
+  char *tolower(const char *input)
+  {
+    if (!input) return 0;
+    const size_t size = ACE_OS::strlen (input);
+    char *output = 0;
+    ACE_NEW_RETURN (output, char [size], 0);
```
mitza-oci, line 597:
```suggestion
    ACE_NEW_RETURN (output, char[size], 0);
```

TAO/TAO_IDL/driver/drv_preproc.cpp
```diff
@@ -577,11 +578,107 @@ DRV_add_include_path (ACE_CString& include_path,
...
 {
+  char idl_version_arg[128];
+  ACE_OS::sprintf (idl_version_arg, "-D__TAO_IDL_IDL_VERSION=%s",
+    idl_global->idl_version_.to_macro ());
+  DRV_cpp_putarg (idl_version_arg);
+
+  // Decide how files will be passed to the preprocessor
+  if (idl_global->preprocessor_input_ == IDL_GlobalData::PreprocessorInputGuess)
+    {
+      const char * const cpp_name = tolower (our_basename (FE_get_cpp_loc_from_env ()));
```
mitza-oci, line 619:
```suggestion
      const char *const cpp_name = tolower (our_basename (FE_get_cpp_loc_from_env ()));
```

TAO/TAO_IDL/driver/drv_preproc.cpp
```diff
@@ -577,11 +578,107 @@ DRV_add_include_path (ACE_CString& include_path,
...
+      const char * method = "copy";
```
mitza-oci, line 620:
```suggestion
      const char *method = "copy";
```

TAO/TAO_IDL/driver/drv_preproc.cpp
```diff
@@ -1358,25 +1459,27 @@ DRV_pre_proc (const char *myfile)
...
-      (void) ACE_OS::unlink (t_ifile);
+      if (copy) (void) ACE_OS::unlink (t_ifile);
       throw Bailout ();
     }
 
   if (WIFEXITED ((status)))
     {
       // Child terminated normally?
-      if (WEXITSTATUS ((status)) != 0)
+      const ACE_exitcode narrowed_status = WEXITSTATUS ((status));
```
mitza-oci, line 1469:
```suggestion
      ACE_exitcode const narrowed_status = WEXITSTATUS ((status));
```

TAO/docs/compiler.html
iguessthislldo: These are non-breaking dashes, which are like [non-breaking spaces](https://en.wikipedia.org/wiki/Non-breaking_space). They prevent line breaks around the dash which was making the page look weird in some sections, especially in the options table, where it would look something like this:

```
--
idl-version VERSION
```

TAO/docs/compiler.html
iguessthislldo: I agree that it's unlikely. Maybe if the user was somehow dealing with multiple versions of TAO or maybe more likely they updated `tao/Version.h`, but didn't recompile `tao_idl` or something like that. I think is later is already covered by the version guard.

TAO/NEWS
iguessthislldo: Maybe, but it is a behavior difference between the two, so I feel like it has to be mentioned somewhere.

TAO/TAO_IDL/driver/drv_args.cpp
iguessthislldo: No, it's a old typo I copied from the code for `--unknown-annotations`. I will fix both.

TAO/NEWS
mitza-oci: Does it require time travel to have TAO 2 know what TAO 3 does?

jwillemsen:
See my comments of #1357, why not set the preprocessor handling through config*.h files instead of guessing it in tao_idl.

PR #1389, opened 2021-01-06
Inline ACE_OS::thr_gettid

iguessthislldo:


jwillemsen:
Maybe move this operation inline, looks to be very small and would benefit from being inline

jwillemsen (changes requested):
Please move it to the inline file, it is very small

jwillemsen (approved):

jwillemsen:
@iguessthislldo I moved the function inline, please merge when you are ok with this

iguessthislldo:
Yes, this is fine.

PR #1414, opened 2021-01-18
g++ 4 has a problem that it warns when using a std::array containin…

jwillemsen:
…g std::string even when {{}} is used, disable this warning only for gcc 4.8

    * ACE/include/makeinclude/platform_g++_common.GNU:

### DIFF
--- a/ACE/include/makeinclude/platform_g++_common.GNU
+++ b/ACE/include/makeinclude/platform_g++_common.GNU
@@ -79,6 +79,12 @@ endif
 # gcc 4 has C++03 as default C++ version, enable this to be C++11
 ifeq ($(findstring $(CXX_MAJOR_VERSION),4),$(CXX_MAJOR_VERSION))
   c++11 ?= 1
+  # gcc 4.8 has a problem that it warnings about missing initializers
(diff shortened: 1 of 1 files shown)


ACE/include/makeinclude/platform_g++_common.GNU
```diff
@@ -79,6 +79,12 @@ endif
 # gcc 4 has C++03 as default C++ version, enable this to be C++11
 ifeq ($(findstring $(CXX_MAJOR_VERSION),4),$(CXX_MAJOR_VERSION))
   c++11 ?= 1
+  # gcc 4.8 has a problem that it warnings about missing initializers
+  # for an std::array containing std::string even when {{}} is used
+  # as it should
+  ifeq ($(findstring $(CXX_MINOR_VERSION),8),$(CXX_MINOR_VERSION))
```
mitza-oci, line 85: This should apply to all 4.x GCC versions

mitza-oci (changes requested):

mitza-oci:
Please merge to ACE6 branch

PR #1424, opened 2021-01-25
Don't force c++11 through the github action, that should be the default

jwillemsen:
* .github/workflows/macosx.yml:


jwillemsen:
Looks clang on macosx by default is c++03, something to resolve in the platform files for macosx

PR #1429, opened 2021-02-12
Removed DDS_ROOT from diff-builds-and-group-fixed-tests-only.sh

mitza-oci:


iguessthislldo:
There's also [an OpenDDS option in `auto_run_tests.pl`](https://github.com/DOCGroup/ACE_TAO/blob/be2df3244a3cfb5070fa8b5f93c90d9e81d6edf1/ACE/bin/auto_run_tests.pl#L141) if we're removing reverences to OpenDDS.

jwillemsen (approved):

mitza-oci:
> 
> 
> There's also [an OpenDDS option in `auto_run_tests.pl`](https://github.com/DOCGroup/ACE_TAO/blob/be2df3244a3cfb5070fa8b5f93c90d9e81d6edf1/ACE/bin/auto_run_tests.pl#L141) if we're removing reverences to OpenDDS.

I think that is actually being used by builds that @jwillemsen maintains, so I'll leave it to him to decide what to do with it.

jwillemsen:
It looks I need to update our builds because of changes in OpenDDS related to the tests, so just let me know what you I need to change, I can refer to the lst files directly in the builds so that you can remove the `-d` from the auto_run_tests.pl script @mitza-oci

mitza-oci:
> 
> 
> It looks I need to update our builds because of changes in OpenDDS related to the tests, so just let me know what you I need to change, I can refer to the lst files directly in the builds so that you can remove the `-d` from the auto_run_tests.pl script @mitza-oci

I think it's more maintainable long term to have OpenDDS use the `auto_run_tests` script in its own repository and not depend on details of the script in ACE.
To do this, it looks like the changes would be here:
```
<variable name="auto_run_tests_script_path"  value="%ACE_ROOT%/bin" />
```
to use DDS_ROOT/tests
and
```
 <command name="auto_run_tests"    options="-d ...
```
replacing `-d` with `-a`

jwillemsen:
Ok, made those changes to our buildtools helper scripts, feel free to remove -d from `auto_run_tests.pl`

PR #1450, opened 2021-03-11
Doxygen changes and make use of delete

jwillemsen:


albert-github:
I see that there are a number of `Doxyfile`s involved in the creation of the documentation here. Maybe an idea is to use a master `Doxyfile` and include this in the different `Doxyfile`s with `@INCLUDE = <the master Doxyfile>`  and further in the different `Doxyfile`s just use the differences specific for that sub project. This will work as doxygen uses the last seen setting

Note: that a `Doxyfile` just has to contain the settings different from the default settings. A setting that is not present will be set to its default. An interesting option with doxygen in this case is `-x` showing just the differences fro the default so e.g. `doxygen -x Doxyfile`

jwillemsen:
> Note: that a `Doxyfile` just has to contain the settings different from the default settings. A setting that is not present will be set to its default. An interesting option with doxygen in this case is `-x` showing just the differences fro the default so e.g. `doxygen -x Doxyfile`

Would be nice to have a combination of `-u` and `-x` to upgrade to a newer doxygen file and remove all settings that are default

albert-github:
I don't think the `-u`  is not necessary as in the `Doxyfile`s  of the sub project one only needs the settings different from the default.
The `-u` can still be applied on \<the master Doxyfile\>.
One small disadvantage of the `-x` might be that comment is lost.

Or did I miss a use case?

jwillemsen:
That will work when someone finds time to rework all files, feel free to open a PR

albert-github:
Regarding the Internal inconsistencies I just pushed a proposed patch: https://github.com/doxygen/doxygen/pull/8429

A number of problems can be overcome by not placing the `@namespace` command when documenting a `module` directly before it.
So change e.g.
```
/**
 * @brief Define types shared by the real-time scheduling and event
 * services.
 *
 * These types are defined in a separate module to decouple the
 * real-time event services from the scheduling service interface (and
 * implementation!)
 */
module RtecBase
```
into
```
/**
 * @namespace RtecBase
 *
 * @brief Define types shared by the real-time scheduling and event
 * services.
 *
 * These types are defined in a separate module to decouple the
 * real-time event services from the scheduling service interface (and
 * implementation!)
 */
module RtecBase
```
the proposed patch also contains new commands like `\idlmodule` so `\namespace` is not necessary anymore in the above mentioned case but `\idlmodule` can be used.

Some problems remain when we have an IDL module and in a .h or .cpp file we also have a (forward declared?) namespace with the same name.
(lack of knowledge on my side about IDL and the direct interaction with C++).

jwillemsen:
> Some problems remain when we have an IDL module and in a .h or .cpp file we also have a (forward declared?) namespace with the same name.
> (lack of knowledge on my side about IDL and the direct interaction with C++).

A IDL module foo will result in generated code that uses the C++ namespace Foo but also our code can be in that namespace. It is normal to have IDL modules and C++ namspaces using the same name, it is what is defined by the IDL to C++ language mapping

albert-github:
Thanks for the clarification.
- I also saw that there is the term `library` and `constants` is there a similar mapping from the IDL domain to the C++ domain?
- Is there an official IDL language definition standard / document?

jwillemsen:
IDL constants are mapped to C++ constants. Library is not a formal IDL term. See https://www.omg.org/spec/IDL for the formal IDL standard and https://www.omg.org/spec/CPP for the mapping from IDL to C++

albert-github:
Thanks, I think that there is even an easier solution to the problem shown in the ACE TAO project.

I just did a quick test and this looks very good, no warnings anymore about the "Internal inconsistencies", I'm still a bit in doubt whether I should:
1. just use the patch I just created locally.
2. use the patch from before plus the local patch

- advantages
  1. small 
  2. cleaner
- disadvantages
  1. no clean separation
  2. extra complexity

Well it all depends on the output, so some checks to be done...

albert-github:
I did some tests with executables as indicated in https://github.com/DOCGroup/ACE_TAO/pull/1450#issuecomment-799277574 and the HTML output for the ACE TAO project was for both versions identical, just the generated tag file were (as expected) different.

It is to be decided by @doxygen whether any of the pull requests should be honored or the pull request https://github.com/doxygen/doxygen/pull/8431 should be honored or the pull request https://github.com/doxygen/doxygen/pull/8429.
I think the pull request #8429 is better and more complete but also more complex and introduces some new doxygen commands.

albert-github:
Both mentioned pull requests (doxygen/doxygen#8431 and doxygen/doxygen#8429) have been integrated into the current ,master version of doxygen, maybe a good time to test again ...

PR #1470, opened 2021-04-08
Defer Including rules.tao.GNU for gnuace

iguessthislldo:
Alternative fix for https://github.com/objectcomputing/OpenDDS/pull/2559. See there for details.

https://github.com/DOCGroup/ACE_TAO/pull/1296 changed the order of includes of rules make files to effectively being how the base projects that set them are included. This can lead to the order of `GNUACE_PROJECT_VERSION ?=` being set differently than what the case was in ACE6/TAO2. The result in OpenDDS is that so library files get TAO's version number.

This tires to fix that by including `tao.rules.GNU` at a marker after the `extension` marker.

### DIFF
--- a/ACE/bin/MakeProjectCreator/templates/gnu.mpd
+++ b/ACE/bin/MakeProjectCreator/templates/gnu.mpd
@@ -200,6 +200,7 @@ VSHDIR = <%targetoutdir%>.shobj/
 include $(ACE_ROOT)/include/makeinclude/wrapper_macros.GNU
 
 <%marker(extension)%>
+<%marker(deferred-extension)%>
 
 <%if(version)%>
 GNUACE_PROJECT_VERSION = <%version%>
--- a/TAO/MPC/config/tao_rules.mpb
+++ b/TAO/MPC/config/tao_rules.mpb
@@ -1,5 +1,5 @@
 project {
-  verbatim(gnuace, extension, 1) {
+  verbatim(gnuace, deferred-extension, 1) {
     include $(TAO_ROOT)/rules.tao.GNU
   }
 }


mcorino (changes requested):
The new tao_rules was introduced to create correctly (in our opinion) ordered `rules.*` includes where the rules for a "derived" library gets inserted before a "base" library.
In AXCIOMA the manual insertions for CIAO, DAnCE and TAO and non-ordered `rules.*` includes were messing up the versioning schemes for libraries. This seemed like a more logical approach. I agree the issue with `GNUACE_PROJECT_VERSION` is the real underlying problem but did not have a better idea to fix that.

mitza-oci:
In order to facilitate a smooth migration from TAO 2 to 3 for any downstream project, we should maintain the behavior that
`<%marker(extension)%>` comes before `rules.tao.GNU` is included.
I think it's reasonable (though maybe not what we want long term) for TAO to be handled as a special case since it's in this same repository.

mcorino:
> In order to facilitate a smooth migration from TAO 2 to 3 for any downstream project, we should maintain the behavior that
> `<%marker(extension)%>` comes before `rules.tao.GNU` is included.

The way things were before #1296 the `rules.tao.GNU` would only be included after the `extension` marker for a project for which the `tao` MPC value would evaluate to `true` which is handled by an obscure and unreliable piece of code in `$ACE_ROOT/bin/MakeProjectCreator/modules/GNUACEProjectCreator.pm` based on comparisons of include and library paths with similar setups for `ciao` and `dance`. Using that setup was a disaster waiting to happen and it did for AXCIOMA.
I can live with a reliable solution for inserting the include of `rules.tao.GNU` after the `extension` marker for any project using tao instead of relying on the correct order of using the current `tao_rules` base project as that should work just fine for AXCIOMA without any changes. The proposed solution does exactly that but I had hoped for a solution without the need for yet again another verbatim section.

mitza-oci (approved):

mcorino (approved):

PR #1473, opened 2021-04-12
[ACE6] Fix wstring coerce leak in tao_idl

simpsont-oci:
This updates the wstring case, which has a leak, to behave more like the string case, which does not have a leak.


mitza-oci (approved):

jwillemsen:
Does this change also needs to be applied to master?

mitza-oci:
> 
> 
> Does this change also needs to be applied to master?

We are already working on it.

simpsont-oci:
https://github.com/DOCGroup/ACE_TAO/pull/1475

PR #1478, opened 2021-04-22
MacOSX 11.1/11.2 are both bigsur releases

jwillemsen:
* ACE/include/makeinclude/platform_macosx.GNU:


jwillemsen:
See https://support.apple.com/en-us/HT211896

iguessthislldo:
Why is this file so strict about the macOS version? Shouldn't it just accept it if it's a newer than the latest known version?

jwillemsen:
It does fallback to bigsur for 11.1/11.2, but I just added these to prepare for a future release, at least the detection for 11.1/11.2 is explicit to bigsur

PR #1481, opened 2021-04-27
Support C++ Keywords in `DCPS_DATA_SEQUENCE_TYPE`

iguessthislldo:
Made for https://github.com/objectcomputing/OpenDDS/pull/2608


jwillemsen:
Wouldn't it be easier for maintenance at the TAO side that you check the TAO version in opendds?

mitza-oci:
> 
> 
> Wouldn't it be easier for maintenance at the TAO side that you check the TAO version in opendds, little bit concerned that each bug fix in TAO would result in a new macro to maintain forever

There are a few ways we could manage this.  One downside of TAO versions is that the current version on the base branch is the same as the version of the previous tag.

Another approach is to go ahead and do an ACE_TAO release with this (both ace6tao2 and ace7tao3) and then the code in OpenDDS can assume it's available.

jwillemsen:
Safest and easiest to my opinion is to check for the next tao release, that is also what we do regularly for TAO fixes we need for AXCIOMA

iguessthislldo:
The reason I'm creating macros for this is I don't want to have to check for three different versions that might have this feature if we're merging it into every TAO that OpenDDS supports.

mitza-oci:
> The reason I'm creating macros for this is I don't want to have to check for three different versions that might have this feature if we're merging it into every TAO that OpenDDS supports.

The proposed alternate approach has no version checks in OpenDDS.  It assumes a supported TAO version is being used.

jwillemsen:
Yes, just assume the fix is there, that sounds better, that way we don't add macros we need to keep forever

mitza-oci (approved):

iguessthislldo:
Alright, done.

PR #1500, opened 2021-05-06
[ACE6] Branch Option for Release Script

iguessthislldo:


ACE/bin/make_release.py
```diff
@@ -461,8 +467,8 @@ def update_latest_branch (product, which):
 
     name = "Latest_" + which
 
-    vprint ('Fast-forwarding', name, 'to master')
-    ex ("cd $DOC_ROOT/" + product + " && git fetch . master:" + name)
+    vprint ('Fast-forwarding', name, 'to', opts.branch)
```
jwillemsen, line 470: More places where a space is missing in the output 

``` suggestion
    vprint ('Fast-forwarding', name, 'to ', opts.branch)
```

jwillemsen:
The Latest_* branches should be updated with a new ACE6TAO2 release

ACE/bin/make_release.py
iguessthislldo: This function passes its arguments to Python's builtin `print`, which adds spaces between arguments by default.

mitza-oci (approved):

PR #1508, opened 2021-05-13
[ACE6] Fix SocketConnect::ip_check() Concurrency and Too-Early Request Issues for Windows

simpsont-oci:
Problems:

On Windows, SocketConnect::ip_check() will attempt to get the list of IP addresses associated with local network interfaces in order to determine if IPv6 or IPv4 is enabled, and then it will set a global flag in order to cache the result of this expensive check. Unfortunately, the global flag is toggled back and forth during this check, which is not thread safe. In addition, it appears as though this check can be done "too early" within a windows process (particularly if done as part of a static initialization) which will cause & cache a false negative, since no interfaces addresses are found.

Solutions:
1) Protect the global flag from modification until a real answer is available, and protect against recursive calls using a static initialized behind the existing mutex.
2) Protect against the "too early" problem by defaulting to "enabled" when answers aren't available, but defer caching the result until a real answer is available.

### DIFF
--- a/ACE/ace/Sock_Connect.cpp
+++ b/ACE/ace/Sock_Connect.cpp
@@ -1473,6 +1473,9 @@ ip_check (int &ipvn_enabled, int pf)
     {
 
 #if defined (ACE_WIN32)
+      static int recursing = 0;
+      if (recursing) return recursing;
+
       // as of the release of Windows 2008, even hosts that have IPv6 interfaces disabled
       // will still permit the creation of a PF_INET6 socket, thus rendering the socket
       // creation test inconsistent. The recommended solution is to get the list of
@@ -1480,14 +1483,22 @@ ip_check (int &ipvn_enabled, int pf)
       ACE_INET_Addr *if_addrs = 0;
       size_t if_cnt = 0;
 
-      ipvn_enabled = 1; // assume enabled to avoid recursion during interface lookup.
+      // assume enabled to avoid recursion during interface lookup.
+      recursing = 1;
       ACE::get_ip_interfaces (if_cnt, if_addrs);
-      ipvn_enabled = 0;
-      for (size_t i = 0; ipvn_enabled == 0 && i < if_cnt; i++)
+      recursing = 0;
+
+      bool found = false;
+      for (size_t i = 0; !found && i < if_cnt; i++)
         {
-          ipvn_enabled = (if_addrs[i].get_type () == pf);
+          found = (if_addrs[i].get_type () == pf);
         }
       delete [] if_addrs;
+
+      // If the list of interfaces is empty, we've tried too quickly. Assume enabled, but don't cache the result
+      if (!if_cnt) return true;
+
+      ipvn_enabled = found ? 1 : 0;
 #else
       // Determine if the kernel has IPv6 support by attempting to
       // create a PF_INET6 socket and see if it fails.
--- a/ACE/ace/Sock_Connect.cpp
+++ b/ACE/ace/Sock_Connect.cpp
@@ -1474,7 +1474,7 @@ ip_check (int &ipvn_enabled, int pf)
 
 #if defined (ACE_WIN32)
       static int recursing = 0;
-      if (recursing) return recursing;
+      if (recursing) return 1;
 
       // as of the release of Windows 2008, even hosts that have IPv6 interfaces disabled
       // will still permit the creation of a PF_INET6 socket, thus rendering the socket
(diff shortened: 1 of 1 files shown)


ACE/ace/Sock_Connect.cpp
```diff
@@ -1473,21 +1473,32 @@ ip_check (int &ipvn_enabled, int pf)
...
+      bool found = false;
+      for (size_t i = 0; !found && i < if_cnt; i++)
         {
-          ipvn_enabled = (if_addrs[i].get_type () == pf);
+          found = (if_addrs[i].get_type () == pf);
         }
       delete [] if_addrs;
+
+      // If the list of interfaces is empty, we've tried too quickly. Assume enabled, but don't cache the result
+      if (!if_cnt) return true;
```
mitza-oci, line 1499:
```suggestion
      if (!if_cnt) return 1;
```

ACE/ace/Sock_Connect.cpp
```diff
@@ -1473,21 +1473,32 @@ ip_check (int &ipvn_enabled, int pf)
     {
 
 #if defined (ACE_WIN32)
+      static int recursing = 0;
+      if (recursing) return recursing;
```
mitza-oci, line 1477:
```suggestion
      if (recursing) return 1;
```
same effect, but I think this is more clear to the reader

ACE/ace/Sock_Connect.cpp
```diff
@@ -1473,21 +1473,32 @@ ip_check (int &ipvn_enabled, int pf)
...
+
+      bool found = false;
+      for (size_t i = 0; !found && i < if_cnt; i++)
         {
-          ipvn_enabled = (if_addrs[i].get_type () == pf);
+          found = (if_addrs[i].get_type () == pf);
         }
       delete [] if_addrs;
+
+      // If the list of interfaces is empty, we've tried too quickly. Assume enabled, but don't cache the result
```
mitza-oci, line 1498: Since this "assumes enabled" we might have a false positive.  What's the implication of this for the code (like INET_Addr) that needs an answer?

mitza-oci (changes requested):

ACE/ace/Sock_Connect.cpp
simpsont-oci: For recursion, the previous code was also "assuming enabled" so hopefully this isn't a horrible thing to assume. I also don't think I ever saw anything in the code that used ipv4_enabled, only ipv6_enabled and that was only called when ACE_HAS_IPV6 is defined.

That said, it's possible this will return a different result for a Windows machine with ACE_HAS_IPV6 defined if the result is called early enough, but unlike the previous code it should eventually settle on the correct answer (rather than being stuck with the wrong one).

ACE/ace/Sock_Connect.cpp
```diff
@@ -1473,21 +1473,32 @@ ip_check (int &ipvn_enabled, int pf)
     {
 
 #if defined (ACE_WIN32)
+      static int recursing = 0;
```
jwillemsen, line 1476: bool with false/true?

ACE/ace/Sock_Connect.cpp
mitza-oci: I guess the question I had may really be about how this works prior to your PR.

The case where it matters is a Windows + IPv6 build.  Since this code will be called by `ACE_SOCK_Acceptor::open`, `ACE_SOCK_Dgram::open`, `ACE_INET_Addr` constructor, etc., it seems like getting the wrong answer (even temporarily) is still the wrong thing to do.

simpsont-oci: For windows, I'm not sure how to guarantee an answer. Testing in Github Actions was definitely showing that sometimes the early calls were generating empty / bogus lists of addresses. Blocking until an answer is really available doesn't seem like a workable solution. Perhaps we could add compile-time defines for which assumptions to make, that way users could have it default one way or the other until the real answer is known? Or we could add a default run-time parameter and try to update code in the other places that use this.

mitza-oci (approved):

jwillemsen:
Do you also apply the same fixes to master and can you add this to the news file to master and ace6tao2?

jwillemsen:
@mitza-oci @simpsont-oci  Shouldn't this be also merged to master?

mitza-oci:
I think we wanted to see how it did in testing first.  @simpsont-oci please merge if ready.

simpsont-oci:
https://github.com/DOCGroup/ACE_TAO/pull/1549

PR #1519, opened 2021-05-23
Sync debian/* files for ACE

sudipm-mukherjee:


ACE/debian/copyright
```diff
@@ -1,141 +1,251 @@
...
+                docs/ACE-subsets.html
+                docs/ACE-guidelines.html
 
-It is now maintained by:
- Debian ACE+TAO maintainers <pkg-ace-devel@lists.alioth.debian.org>
+Files:     *
+Copyright: 1993-2019 Douglas C. Schmidt and his research group at
+           Washington University, University of California, Irvine,
+           and Vanderbilt University
+License:   other-BSD
```
jwillemsen, line 24: There is a formal DOC license, see https://spdx.org/licenses/DOC

ACE/debian/libace-VERSION.docs
```diff
@@ -1,8 +1,8 @@
-README
-VERSION.txt
-AUTHORS
+usr/share/ace/README
```
jwillemsen, line 1: Will these explicit paths not break things, not sure how this works for example on opensuse build service

jwillemsen (changes requested):

ACE/debian/not-installed
```diff
@@ -0,0 +1,4 @@
+usr/share/ace/ace-devel.sh
+usr/share/ace/bin/ace_install_pkgconfig.pl
+usr/share/ace/bin/ace_gperf
```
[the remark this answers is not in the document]
sudipm-mukherjee: After the build and install step, there are 2 ace_gperf.                       
```
debian/tmp/usr/bin/ace_gperf
debian/tmp/usr/share/ace/bin/ace_gperf
```
`ace_gperf` packake is installing the file `debian/tmp/usr/bin/ace_gperf` and the other file is not installed which has been mentioned in this list.

ACE/debian/copyright
sudipm-mukherjee: Great. I will change the name in the next Debian upload.

ACE/debian/libace-VERSION.docs
sudipm-mukherjee: without the path the files are installed from the source folder and not from the `debian/tmp` where the files to be installed are kept. And so the debhelper thinks these files are not installed. With the paths mentioned debhelper looks for them in the debian/tmp and installs from there. I am not sure how `opensuse build service` works but confused why a Debian specific change in a Debian specific file will break `opensuse build service`.

jwillemsen: Ok, so it should keep working

jwillemsen (approved):

ACE/debian/libace-VERSION.docs
sudipm-mukherjee: @jwillemsen yes. Debian has a tool called `debdiff` which can show the difference between deb packages and I verified before uploading that files were not missing from the binary packages.

PR #1524, opened 2021-06-01
Build TAO IDL_Test as part of the github CI actions

jwillemsen:


jwillemsen:
This looks to be more work, in not all configurations IDL_Test can be compiled

jwillemsen:
The MacOSX github action did fail on the IDL_Test with the errors as reported on https://github.com/RemedyIT/taox11/issues/156

PR #1555, opened 2021-06-16
applying solution to issue 1550 as found in ace6tao2

pmesnier:
bringing over the same change applied in pr#1551.


pmesnier:
The one failure in the CI suite is apparently a defect in windows parallel builds otherwise unrelated to the proposed changes.

PR #1558, opened 2021-06-17
Changed ipaddr_format argument of addr_to_string from int to bool

jwillemsen:
* ACE/ace/ATM_Addr.h:
    * ACE/ace/INET_Addr.cpp:
    * ACE/ace/INET_Addr.h:
    * ACE/ace/MEM_Addr.cpp:
    * ACE/ace/MEM_Addr.h:
    * ACE/ace/QoS/QoS_Session_Impl.cpp:
    * ACE/ace/SOCK_Dgram_Mcast.cpp:
    * ACE/ace/UNIX_Addr.cpp:


jwillemsen:
Could break user code when they override addr_to_string and don't use override to detect the signature change of the base

PR #1573, opened 2021-06-23
msvc and borland should define ACE_HAS_WIN32_STRUCTURAL_EXCEPTIONS, r…

jwillemsen:
…emoves mingw check from the common header

    * ACE/ace/config-win32-borland.h:
    * ACE/ace/config-win32-common.h:
    * ACE/ace/config-win32-msvc.h:


ACE/ace/config-win32-borland.h
```diff
@@ -60,9 +60,10 @@
 
 # include "ace/config-win32-common.h"
 
-# define ACE_WSTRING_HAS_USHORT_SUPPORT 1
-# define ACE_HAS_DIRENT
+#define ACE_WSTRING_HAS_USHORT_SUPPORT 1
+#define ACE_HAS_DIRENT
 
+#define ACE_HAS_WIN32_STRUCTURAL_EXCEPTIONS
```
mitza-oci, line 66: "Structural Exceptions" should be "Structured Exceptions"
https://docs.microsoft.com/en-us/windows/win32/debug/structured-exception-handling

jwillemsen: Yes, I know, renamjng the incorrect macro could break user code, not sure whether we should do that

mitza-oci: ACE 7 may be the opportunity to change it.  Is this for the case where a user customizes `config.h` or where the user is checking the configuration in code that uses ACE?

jwillemsen: I think more where the user checks this define is his own code

jwillemsen:
See https://github.com/DOCGroup/ACE_TAO/pull/1578, old macro is defined for backwards compatibility for the moment

PR #1585, opened 2021-07-01
Fix HP-UX g++ DCCFLAGS.

likema:
Build ACE by GCC on HP-UX 11.31 IA64 :

```bash
make optimize=0 inline=0 debug=1
```

**No debug info** is produced because `DCCFLAGS` is not set to `-g`

### DIFF
--- a/ACE/include/makeinclude/platform_hpux_gcc.GNU
+++ b/ACE/include/makeinclude/platform_hpux_gcc.GNU
@@ -44,6 +44,7 @@ endif
 #CCFLAGS		+= -fstrict-prototype
 #endif
 DCFLAGS		+= -g
+DCCFLAGS	+= -g
 DLD		= $(CXX)
 LD		= $(CXX)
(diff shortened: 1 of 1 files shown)


ACE/include/makeinclude/platform_hpux_gcc.GNU
```diff
@@ -44,6 +44,7 @@ endif
 #CCFLAGS		+= -fstrict-prototype
 #endif
 DCFLAGS		+= -g
+DDCFLAGS	+= -g
```
jwillemsen, line 47: DCCFLAGS?

jwillemsen (changes requested):
DCCFLAGS

ACE/include/makeinclude/platform_hpux_gcc.GNU
likema: Sorry for typo

jwillemsen (approved):

PR #1594, opened 2021-07-08
Fix Coverity Issues and Compiler Warnings

iguessthislldo:
Follow up to #840


mitza-oci (approved):

mitza-oci (approved):

jwillemsen (approved):

jwillemsen:
One warning left, see https://buildlogs.remedy.nl/win_vs2019_opendds_debug/2021_07_11_02_18_Full.html#warning_1

iguessthislldo:
That looks like it would have been there before, but okay.

iguessthislldo:
#1602

mitza-oci:
> 
> 
> That looks like it would have been there before, but okay.

I think it's due to precedence.  The prior code had one more set of parens `(float) ((ev->u.bval == true) ? 1.0 : 0.0)`, the current code applies the cast to first operand of the `?:`.  One more reason why C++ casts are better than C casts.

PR #1627, opened 2021-07-22
Escape Double Quotes in tao_idl PP Args on Windows

iguessthislldo:
To fix issue with #1621 on the scoreboards.

### DIFF
--- a/TAO/TAO_IDL/driver/drv_preproc.cpp
+++ b/TAO/TAO_IDL/driver/drv_preproc.cpp
@@ -140,71 +140,69 @@ DRV_cpp_putarg (const char *str)
       throw Bailout ();
     }
 
-  const ACE_TCHAR *arg_to_add = nullptr;
+  char *replace = nullptr;
+  if (str)
+    {
+      const char *const first_quote = ACE_OS::strchr (str, '"');
+      bool allocate_error = false;
 
-  if (str && ACE_OS::strchr (str, ' ') && !ACE_OS::strchr (str, '"'))
-    {
-      ACE_TCHAR *buf = nullptr;
-      ACE_NEW_NORETURN (buf, ACE_TCHAR[ACE_OS::strlen (str) + 3]);
-      if (buf)
+      if (ACE_OS::strchr (str, ' ') && !first_quote)
         {
-          buf[0] = ACE_TEXT ('"');
-          ACE_OS::strcpy (buf + 1, ACE_TEXT_CHAR_TO_TCHAR (str));
-          ACE_OS::strcat (buf, ACE_TEXT ("\""));
-          arg_to_add = buf;
-        }
-      else
-        {
-          idl_global->err()->misc_error ("DRV_cpp_putarg failed to allocate buffer!");
-          throw Bailout ();
-        }
-    }
-  else if (str)
-    {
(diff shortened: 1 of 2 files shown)


iguessthislldo:
... And vcpkg is kaput for a Windows specific PR. Perfect.

TAO/TAO_IDL/driver/drv_preproc.cpp
```diff
@@ -147,14 +151,60 @@ DRV_cpp_putarg (const char *str)
...
     }
-  else
+  else if (str)
     {
-      DRV_arglist[DRV_argcount++] =
-        ACE::strnew (ACE_TEXT_CHAR_TO_TCHAR (str));
+#ifdef ACE_WIN32
+      // Escape Doublequotes on Windows
+      size_t quote_count = 0;
+      for (const char* quote = ACE_OS::strchr (str, '"'); quote; quote = ACE_OS::strchr (quote, '"'))
```
mitza-oci, line 167:
```suggestion
      for (const char *quote = ACE_OS::strchr (str, '"'); quote; quote = ACE_OS::strchr (quote, '"'))
```
also below on 179

TAO/TAO_IDL/driver/drv_preproc.cpp
```diff
@@ -147,14 +151,60 @@ DRV_cpp_putarg (const char *str)
...
+            {
+              ACE_TCHAR *to = buf;
+              for (const char* from = str; *from; ++from)
+                {
+                  if (*from == '"')
+                    {
+                      *to = ACE_TEXT ('\\');
+                      ++to;
+                    }
+                  *to = *from;
```
mitza-oci, line 186: These are (potentially) different types, char and tchar.  Does this assignment always work?  At least add a comment?

iguessthislldo: Assuming `char` is ASCII and `wchar` encoding is ASCII-based it should just should widen the value like an int cast, but yeah that's probably assuming too much. Probably needs to convert the whole str first.

iguessthislldo: Actually I realized that's insane to work with `ACE_TCHAR` when I didn't have to and did the opposite of converting it first by converting it at the very end.

mitza-oci (approved):

TAO/TAO_IDL/driver/drv_preproc.cpp
```diff
@@ -138,22 +140,68 @@ DRV_cpp_putarg (const char *str)
...
     }
-  else
+
+  DRV_arglist[DRV_argcount++] = ACE::strnew (ACE_TEXT_CHAR_TO_TCHAR (replace ? replace : str));
+
+  if (replace)
     {
-      DRV_arglist[DRV_argcount++] =
-        ACE::strnew (ACE_TEXT_CHAR_TO_TCHAR (str));
+      delete replace;
```
jwillemsen, line 203: New coverity error, shouldn't this be delete[] @iguessthislldo

iguessthislldo: Yes: #1633 and #1634.

PR #1628, opened 2021-07-22
[ACE6] Escape Double Quotes in tao_idl PP Args on Windows

iguessthislldo:
This is a backport of https://github.com/DOCGroup/ACE_TAO/pull/1627

### DIFF
--- a/TAO/TAO_IDL/driver/drv_preproc.cpp
+++ b/TAO/TAO_IDL/driver/drv_preproc.cpp
@@ -140,71 +140,69 @@ DRV_cpp_putarg (const char *str)
       throw Bailout ();
     }
 
-  const ACE_TCHAR *arg_to_add = 0;
+  char *replace = 0;
+  if (str)
+    {
+      const char *const first_quote = ACE_OS::strchr (str, '"');
+      bool allocate_error = false;
 
-  if (str && ACE_OS::strchr (str, ' ') && !ACE_OS::strchr (str, '"'))
-    {
-      ACE_TCHAR *buf = 0;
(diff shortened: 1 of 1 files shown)


TAO/TAO_IDL/driver/drv_preproc.cpp
```diff
@@ -138,6 +140,8 @@ DRV_cpp_putarg (const char *str)
       throw Bailout ();
     }
 
+  const ACE_TCHAR *arg_to_add = 0;
+
   if (str && ACE_OS::strchr (str, ' ') && !ACE_OS::strchr (str, '"'))
```
simpsont-oci, line 145: I know this isn't your code... but is this first part even right? What if my string is `abc "bob"` ... it will escape the `"`s but not the ` `, right?

simpsont-oci (approved):
rerunning the windows jobs since they failed do to cancelation / vcpkg issues

TAO/TAO_IDL/driver/drv_preproc.cpp
iguessthislldo: I didn't pay much mind to it, but now that you mention it, I don't know. I don't know why that and what I'm trying to fix is a problem at all since it should be passing in arguments like `execv` in an `argv` style instead of apparently treating the whole dang thing as a single command line string.

mitza-oci: I think the part you're discussing is something I updated not too long ago to handle paths with spaces.  It wasn't meant to cover every single possibility, but just the case where you have an argument that is a filesystem path (for example, when this is called from DRV_add_include_path) with a space that's not already quoted.

jwillemsen:
Backport of #1627

PR #1632, opened 2021-07-22
[GitHub Actions] Disable fail-fast strategy for matrix builds

simpsont-oci:
Problem:
When a single matrix build fails (say, for example, due to vcpkg errors), it cancels all other parallel matrix builds in the same job. This is unhelpful in most situations, as we'd still usually like to see the other test results.

Solution:
Disable the default 'fail fast' strategy for github actions to allow other matrix jobs to continue.

### DIFF
--- a/.github/workflows/face.yml
+++ b/.github/workflows/face.yml
@@ -14,7 +14,7 @@ concurrency:
 jobs:
   build:
     strategy:
-      fast-fail: false
+      fail-fast: false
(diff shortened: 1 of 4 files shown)


iguessthislldo (approved):

jwillemsen (approved):

mitza-oci:
Does this need to be merged to `ace6tao2`?

PR #1641, opened 2021-08-03
[ACE 6] Fixed warnings related to Handle_Set's copy constructor

mitza-oci:
GCC 11:
```
ace/Select_Reactor_T.cpp:1496:18: warning: implicitly-declared 'constexpr ACE_Handle_Set::ACE_Handle_Set(const ACE_Handle_Set&)' is deprecated [-Wdeprecated-copy]
ace/Handle_Set.inl:31:1: note: because 'ACE_Handle_Set' has user-provided 'ACE_Handle_Set& ACE_Handle_Set::operator=(const ACE_Handle_Set&)'
```
Not a problem for ACE 7

### DIFF
--- a/ACE/ace/Handle_Set.h
+++ b/ACE/ace/Handle_Set.h
@@ -113,6 +113,8 @@ public:
   fd_set *fdset (void);
 
 #if defined (ACE_HAS_BIG_FD_SET)
+  ACE_Handle_Set (const ACE_Handle_Set &other);
+
   /// Assignment operator optimizes for cases where <size_> == 0.
   ACE_Handle_Set & operator= (const ACE_Handle_Set &);
 #endif /* ACE_HAS_BIG_FD_SET */
--- a/ACE/ace/Handle_Set.inl
+++ b/ACE/ace/Handle_Set.inl
@@ -27,6 +27,13 @@ ACE_Handle_Set::reset (void)
 }
 
 #if defined (ACE_HAS_BIG_FD_SET)
+ACE_INLINE ACE_Handle_Set::ACE_Handle_Set (const ACE_Handle_Set &other)
+  : size_ (other.size_)
+  , max_handle_ (other.max_handle_)
+  , min_handle_ (other.min_handle_)
+  , mask_ (other.mask_)
+{}
+
 ACE_INLINE ACE_Handle_Set &
 ACE_Handle_Set::operator = (const ACE_Handle_Set &rhs)
 {


jwillemsen:
See #1463 for the fix on master. Can this be for the next micro, our builds are already running

mitza-oci:
> 
> 
> See #1463 for the fix on master.

The code I added is equivalent to what the compiler's already doing, member-wise (trivial) copy construction.

> Can this be for the next micro, our builds are already running

Yes, just want to get the last warning off of our GCC 11 build that uses the `ace6tao2` branch with OpenDDS

jwillemsen:
We should have the same change on master and ace6tao2

mitza-oci:
> 
> 
> We should have the same change on master and ace6tao2

I don't think anything's going to be faster than a trivial copy, compiler can turn it in to a `memcpy`.  But maybe there's something about how this is used that I'm not aware of.

jwillemsen:
In terms of long terms maintenance it looks safer to have the same fix as on master backported and not a different change.

jwillemsen (changes requested):
In terms of long terms maintenance it looks safer to have the same fix as on master backported and not a different change.

mitza-oci:
> 
> 
> In terms of long terms maintenance it looks safer to have the same fix as on master backported and not a different change.

I think there are tradeoffs for each option.  It may be better to have stability for ACE 6.

I'd be fine with closing this PR and merging some other approach, I'd just like to avoid having ACE headers generate warnings in downstream projects using default compiler options.

jwillemsen:
Newer compilers just give a lot of warnings, trying to get ace6tao2 warning free will result in much more changes, even master gives several warnings with newer compilers. Personally I am not in favor of trying to get ace6tao2 warning free, would prefer to get master warning free.

mitza-oci:
> 
> 
> Newer compilers just give a lot of warnings, trying to get ace6tao2 warning free will result in much more changes, even master gives several warnings with newer compilers. Personally I am not in favor of trying to get ace6tao2 warning free, would prefer to get master warning free.

The goal is not to get the build completely warning-free.  The goal is to get warnings out of public headers that are commonly used since they will cause issues for downstream projects.

jwillemsen:
> > Newer compilers just give a lot of warnings, trying to get ace6tao2 warning free will result in much more changes, even master gives several warnings with newer compilers. Personally I am not in favor of trying to get ace6tao2 warning free, would prefer to get master warning free.
> 
> The goal is not to get the build completely warning-free. The goal is to get warnings out of public headers that are commonly used since they will cause issues for downstream projects.

But most of that work for newer compilers is already done on master, when someone has a new compiler that gives these new warnings why not upgrade to ACE7 which has the fixes? These new compilers already have C++11 or even C++17 as default C++ level so it is not that we are forcing users to do something additionally. Is there a real reason why they have to keep using ACE 6?

mitza-oci:
I guess this isn't going to be resolved in time for the upcoming release.  That's fine, it can happen later.

In terms of overall approach to ACE 6, if any of the maintainers don't want to work on this branch then that's OK.  Those that do can maintain it.

jrw972 (approved):

PR #1657, opened 2021-08-20
ACE 6 doesn't support Big Sur

jrw972:
Problem
-------

ACE 6 doesn't compile for macOS Big Sur.

Solution
--------

Update platform_macosx.GNU and related files to support Catalina and
Big Sur.

### DIFF
--- a/ACE/include/makeinclude/platform_macosx.GNU
+++ b/ACE/include/makeinclude/platform_macosx.GNU
@@ -20,18 +20,27 @@ MACOS_CODENAME_VER_10_11  := elcapitan
 MACOS_CODENAME_VER_10_12  := sierra
 MACOS_CODENAME_VER_10_13  := highsierra
 MACOS_CODENAME_VER_10_14  := mojave
-MACOS_CODENAME_VER_latest := mojave
+MACOS_CODENAME_VER_10_15  := catalina
+MACOS_CODENAME_VER_10_latest  := catalina
+MACOS_CODENAME_VER_11_5   := bigsur
+MACOS_CODENAME_VER_11_latest := bigsur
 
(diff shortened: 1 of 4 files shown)


jwillemsen:
Please extend the github CI to test on bigsur, see master

mitza-oci (approved):

ACE/include/makeinclude/platform_macosx.GNU
```diff
@@ -20,18 +20,19 @@ MACOS_CODENAME_VER_10_11  := elcapitan
...
+MACOS_CODENAME_VER_11_5   := bigsur
+MACOS_CODENAME_VER_latest := bigsur
+MACOS_LATEST_MAJOR := 11
+MACOS_LATEST_MINOR := 5
 
 MACOS_CODENAME = $(MACOS_CODENAME_VER_$(MACOS_MAJOR_VERSION)_$(MACOS_MINOR_VERSION))
 
-ifeq ($(MACOS_MAJOR_VERSION),10)
-  ifeq ($(shell test $(MACOS_MINOR_VERSION) -gt 14; echo $$?),0)
+ifeq ($(MACOS_MAJOR_VERSION),$(MACOS_LATEST_MAJOR))
```
iguessthislldo, line 31: I still don't think this check should be here. We should just let try if it's 12.

mitza-oci: ACE 7 doesn't do that as far as I know.  So let's treat it as out of scope for this PR.

iguessthislldo: Yes, that's understandable. I just wanted to point out that we wouldn't have to change this code at all if it already did that because it could just use the last one for 10.

PR #1664, opened 2021-08-25
[ACE6] Update ACE_Atomic_Op_GCC to support newer __atomic operations when available

simpsont-oci:
Problem: ACE_Atomic_Op_GCC_T currently only uses the older __sync style atomic operations, which don't allow for memory ordering, meaning that they often enforce stricter-than-necessary memory barriers in order to achieve synchronization. The `__sync_synchronize()` builtin also isn't supported by clang and GCC's ThreadSanitizer feature, meaning that newer compilers will get false-positive reports of data races when using -fsanitize. ACE_Atomic_Op_GCC_T also appears to be missing load-side synchronization (via `__sync_synchronize()`).

Solution: Update ACE_Atomic_Op_GCC_T's load-side synchronization to use `__sync_synchronize()` and add support for __atomic builtins for GCC versions new enough to support them.

### DIFF
--- a/ACE/ace/Atomic_Op_GCC_T.inl
+++ b/ACE/ace/Atomic_Op_GCC_T.inl
@@ -29,91 +29,149 @@ template <typename T>
 ACE_INLINE T
 ACE_Atomic_Op_GCC<T>::operator++ (void)
 {
-  return __sync_add_and_fetch (&this->value_, 1);
+#if (__GNUC__ > 4) || ((__GNUC__ == 4) && (__GNUC_MINOR__ >= 7))
+  return __atomic_add_fetch (&value_, 1, __ATOMIC_ACQ_REL);
+#else
+  return __sync_add_and_fetch (&value_, 1);
+#endif
 }
 
 template <typename T>
 ACE_INLINE T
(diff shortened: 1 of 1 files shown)


ACE/ace/config-g++-common.h
```diff
@@ -161,7 +161,9 @@
 // GCC >= 4.1 provides __sync_XXXX builtins for use in atomic operations
 // although the builtins are provided globally they are not supported on all platforms
 #if defined (ACE_HAS_THREADS)
-#if (__GNUC__ > 4) || ((__GNUC__ == 4) && (__GNUC_MINOR__ >= 1))
+#if (__GNUC__ > 4) || ((__GNUC__ == 4) && (__GNUC_MINOR__ >= 7))
```
jwillemsen, line 164: gcc 4.7 is the oldest GCC compiler we still support for ACE7, that has C++11 support, couldn't this part be simplified now?

simpsont-oci: For master, yes, it sounds like you could just switch to the __atomics without need for checking compiler versions.

jwillemsen (approved):

jwillemsen:
Do you make similar changes to master?

PR #1675, opened 2021-09-09
Use ACE_OS::strcat, much easier and removes a warning of gcc on strnc…

jwillemsen:
…at usage

    * TAO/TAO_IDL/util/utl_idlist.cpp:


jwillemsen:
New code from #723

PR #1709, opened 2021-10-30
Port to MSVC with /std:c++17 and 20

mitza-oci:


ACE/ace/Malloc.h
```diff
@@ -287,6 +287,8 @@ class ACE_Export ACE_Control_Block
     void dump () const;
   };
 
+  typedef ACE_Malloc_Header *MALLOC_HEADER_PTR;
```
jwillemsen, line 290:
``` suggestion
  using MALLOC_HEADER_PTR = ACE_Malloc_Header *;
```

ACE/ace/Malloc_T.cpp
```diff
@@ -797,18 +797,20 @@ ACE_Malloc_T<ACE_MEM_POOL_2, ACE_LOCK, ACE_CB>::shared_free (void *ap)
...
-                  || blockp < (MALLOC_HEADER *) currp->next_block_))
+                  || blockp < currp->next_block_))
             // Freed block at the start or the end of the memory pool.
             break;
         }
 
       // Join to upper neighbor.
-      if ((blockp + blockp->size_) == currp->next_block_)
+      typename ACE_CB::MALLOC_HEADER_PTR next = blockp;
+      next += static_cast<int> (blockp->size_); // ACE_Based_Pointer has += but not +
```
jwillemsen, line 812: Shouldn't we a type that is 64bit on 64bit systems?

TAO/TAO_IDL/ast/ast_expression.cpp
```diff
@@ -2697,9 +2697,11 @@ dump_expr_val (ACE_OSTREAM_TYPE &o, AST_Expression::AST_ExprValue *ev)
     case AST_Expression::EV_char:
       o << ev->u.cval;
       break;
+#ifndef ACE_HAS_CPP20
```
jwillemsen, line 2700: Instead of not dumping any value, maybe print something to indicate that we have a non-printable character?

ACE/ace/Malloc_T.cpp
mitza-oci: The signature in Based_Pointer_T is `void operator+= (int index);`.  An individual block would be limited to 2GiB.

TAO/TAO_IDL/ast/ast_expression.cpp
mitza-oci: Yes, it will do that using the `default:` case.

ACE/ace/Malloc_T.cpp
jwillemsen: Ok

iguessthislldo (approved):

jwillemsen:
This resulted in some new warnings, see https://buildlogs.remedy.nl/sles150_ace_i686_release/index.html

mitza-oci:
OK, I'm looking into it

mitza-oci:
I see what the problem is. `ACE_Based_Pointer_Basic` has a copy constructor but it doesn't work (just asserts) - this makes it hard to use in generic code where the abstract pointer may be an actual pointer or an `ACE_Based_Pointer_Basic`.  I will try making it actually copyable.

jwillemsen:
At least we should mark these operations as deleted in the header (also add the move variants), and remove the noop implementation when you can't make them copyable

mitza-oci:
> 
> 
> At least we should mark these operations as deleted in the header (also add the move variants), and remove the noop implementation when you can't make them copyable

Done in #1716

PR #1710, opened 2021-11-01
Fix assigning not null-terminated string.

likema:
The result buffer of `ACE_TEXT_RegQueryValueEx` may not be null-terminated.

This bug can be also reproduced on  6.5.x branch.

## Summary by CodeRabbit

- **Refactor**
	- Enhanced the handling of configuration settings to improve overall stability and performance.


jwillemsen:
Please extend one of the unit tests under ACE/tests with a reproducer

likema:
> Please extend one of the unit tests under ACE/tests with a reproducer

It may not be do it easily.

https://docs.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-regqueryvalueexa#remarks shows

>If the data has the REG_SZ, REG_MULTI_SZ or REG_EXPAND_SZ type, the string may not have been stored with the >proper terminating null characters. Therefore, even if the function returns ERROR_SUCCESS, the application should >ensure that the string is properly terminated before using it; otherwise, it may overwrite a buffer. (Note that >REG_MULTI_SZ strings should have two terminating null characters.) One way an application can en`sure that the >string is properly terminated is to use RegGetValue, which adds terminating null characters if needed.

jwillemsen:
Can you check tests/Config_Test, I think it is failing due to this change

PR #1727, opened 2021-11-11
Notify svc: Fixed include

mitza-oci:
The "make install" tree doesn't have two orbsvcs directories


jwillemsen:
There are more files that use `orbsvcs/orbsvcs` as include, shouldn't these all be updated?

mitza-oci:
> 
> 
> There are more files that use `orbsvcs/orbsvcs` as include, shouldn't these all be updated?

This is the only one that changed in #1709 and failed the "make install" testing.

PR #1729, opened 2021-11-12
Installed tree doesn't have double orbvsvcs path

jwillemsen:


jwillemsen:
Similar to https://github.com/DOCGroup/ACE_TAO/pull/1727

PR #1732, opened 2021-11-15
Check also mpc files for tabs

jwillemsen:
* ACE/bin/fuzz.pl:

### DIFF
--- a/ACE/bin/fuzz.pl
+++ b/ACE/bin/fuzz.pl
@@ -535,7 +535,7 @@ sub check_for_tab ()
     return if is_suppressed ();
 
     print "Running tabs check\n";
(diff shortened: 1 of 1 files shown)


jwillemsen:
Some mpc files do need tabs :-(

iguessthislldo:
Yea, our lint script [excludes mpc files from the tabs checks](https://github.com/objectcomputing/OpenDDS/blob/05c82664fc481a2d9cc620d3eb4f00f9911c14ff/tools/scripts/lint.pl#L480). A smarter script could try to parse the MPC files and see if the tab is inside the Make-specific blocks or not. Not sure that's worth that amount of effort though.

jwillemsen:
That is not really worth the trouble.

PR #1739, opened 2021-11-19
fixed comments in tao_idl-generated code

mitza-oci:


iguessthislldo (approved):

jwillemsen:
Which problem does this fix, do you have an example?

mitza-oci:
> Which problem does this fix, do you have an example, 

When I added code in #1709 that used the "ostream" I didn't realize it was that much different from a real `std::ostream` that it didn't support char.  So the char was promoted to an int and the ASCII value of `:` was used instead of a separator between file name and line number.

> some test extension necessary?

No, it's just a utility function with tao_idl

jwillemsen (approved):

PR #1752, opened 2022-01-07
make install, pkgconfig, ace_for_tao fixes

mitza-oci:
Also fixed some ace_for_tao issues


jwillemsen:
Would be nice when we would also test install with MPC on Windows

jwillemsen (approved):

mitza-oci:
> Would be nice when we would also test install with MPC on Windows

Sure.  Since we don't currently have scoreboard builds for it, it's not as simple as copying some commands into the YAML

Due to #1751 I'll try to add pkgconfig checking here.

jwillemsen (approved):

PR #1763, opened 2022-01-15
support thread naming via pthread_setname_np() if available

esohns:


esohns:
please only mind commit 8eae9d8. Something went wrong during branching

jwillemsen:
Please limit this PR to only the change mentioned in the subject, also there is no config file that sets ACE_HAS_PTHREAD_SETNAME_NP, that should be set somewhere?

esohns:
please re-review

jwillemsen:
Please check the diff, very likely you changes line endings, the diff should only show your changes

jwillemsen:
Maybe create a new branch from master and only apply the changes that belong this PR, just redo the change or cherry-pick the original change

ACE/ace/OS_NS_Thread.cpp
```diff
@@ -3615,6 +3615,7 @@ ACE_OS::thr_create (ACE_THR_FUNC func,
 
 # if defined (ACE_HAS_PTHREADS)
   int result;
+  int name_set = 0;
```
jwillemsen, line 3618: Can be bool and only needs to be there when ACE_HAS_PTHREAD_SETNAME_NP is defined

ACE/ace/OS_NS_Thread.cpp
```diff
@@ -4004,6 +4006,18 @@ ACE_OS::thr_create (ACE_THR_FUNC func,
 
 #   endif /* sun && ACE_HAS_ONLY_SCHED_OTHER */
   auto_thread_args.release ();
+
+  // *** Set pthread name (second try)
+#     if defined (ACE_HAS_PTHREAD_SETNAME_NP)
+  if (!name_set && thr_name && *thr_name)
```
jwillemsen, line 4012: Can the bool not be removed when you only do this code when ACE_HAS_PTHREAD_ATTR_SETNAME is not defined?

jwillemsen (changes requested):

PR #1772, opened 2022-01-25
TAO_IDL: Fix Space In Path Handling

ClaytonCalabrese:
There is a problem with handling spaces in paths for TAO_IDL on Win32 systems.  This is due to the places where quotes are added leading to the elseif  in `DRV_cpp_putarg` which incorrectly escapes the quotes. This causes any spaces to not be properly handled.

### DIFF
--- a/TAO/TAO_IDL/driver/drv_args.cpp
+++ b/TAO/TAO_IDL/driver/drv_args.cpp
@@ -186,7 +186,6 @@ DRV_parse_args (long ac, char **av)
   ACE_CString buffer;
   char *s = nullptr;
   long i;
-  bool has_space = false;
 
   // After -- process all arguments as files
   bool just_files = false;
@@ -285,10 +284,7 @@ DRV_parse_args (long ac, char **av)
                   if (i < ac - 1)
                     {
                       idl_global->append_idl_flag (av[i + 1]);
-                      has_space = FE_Utils::hasspace (av[i + 1]);
 
-                      // If the include path has a space, we need to
-                      // add literal "s.
                       ACE_CString arg = av[i];
                       arg += av[i + 1];
 
@@ -303,10 +299,6 @@ DRV_parse_args (long ac, char **av)
                 }
               else
                 {
-                  has_space = FE_Utils::hasspace (av[i]);
-
-                  // If the include path has a space, we need to
-                  // add literal "s.
                   ACE_CString arg (av[i], 2);
                   arg += av[i] + 2;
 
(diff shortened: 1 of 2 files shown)


TAO/TAO_IDL/driver/drv_args.cpp
```diff
@@ -290,9 +290,7 @@ DRV_parse_args (long ac, char **av)
                       // If the include path has a space, we need to
                       // add literal "s.
                       ACE_CString arg = av[i];
```
jwillemsen, line 292: Looks comments also need to be updated

TAO/TAO_IDL/driver/drv_preproc.cpp
```diff
@@ -489,9 +489,7 @@ DRV_sweep_dirs (const char *rel_path,
                   /// case the original path argument included
                   /// spaces. If it didn't, no harm done.
                   ACE_CString incl_arg ("-I ");
-                  incl_arg += '\"';
                   incl_arg += bname;
-                  incl_arg += '\"';
                   DRV_cpp_putarg (incl_arg.c_str ());
```
iguessthislldo, line 493: This one should definitely be two separate calls to `DRV_cpp_putarg`, because it's `-I FILE`.

TAO/TAO_IDL/driver/drv_preproc.cpp
```diff
@@ -188,6 +177,17 @@ DRV_cpp_putarg (const char *str)
             }
         }
 #endif
+      if (ACE_OS::strchr (str, ' ') && !first_quote)
```
iguessthislldo, line 180:
```suggestion
      if (ACE_OS::strchr (str, ' '))
```

I'm reasonably sure we can allow cases where there are quotes and spaces by removing this.

iguessthislldo (changes requested):

TAO/TAO_IDL/driver/drv_preproc.cpp
mitza-oci: Looking at `drv_args.cpp` around line 297, the existing code supports (or attempts to support) either `-I dir` or `-Idir`, though since the latter is documented in the usage message I tend to think of it as preferred.

Based on what we've learned I think it would make sense to either change this (and any other similar calls) to `-Idir` or two calls to `DRV_cpp_putarg`.

ClaytonCalabrese: Do we have a decision on which way we want to handle this? I'm a little concerned about how this would creep the scope of this PR.

mitza-oci: I think it's easier to determine correctness of the fix if each call to `DRV_cpp_putarg` (and thus each entry in `DRV_arglist`) is intended to be a single argument for the preprocessor's `argv`.  It could be a separate PR.

jwillemsen:
We did have various fixes in this area in the past, not sure whether we have any unit tests for this to validate that it works for all cases

jwillemsen:
There is a new warning, see https://buildlogs.remedy.nl/rhel80_opendds_versioned/

PR #1778, opened 2022-02-02
TAO_IDL: Fix Space In Path Handling: Unused Variable Followup

ClaytonCalabrese:
This is a followup to the PR #1772 . 

The `first_quote` variable was unused outside of the `#ifdef ACE_WIN32`, so it was giving a warning on other systems. This PR resolves that by moving its declaration into the ifdef.


mitza-oci (approved):

jwillemsen:
Can you add this change to the TAO NEWS file?

PR #1420, opened 2021-01-21
[TAO2] TAO_IDL: drv_preproc fixed logging and quoting, update the Define_Custom to run from a path with spaces

mitza-oci:

PR #1783, opened 2022-02-11
Fix tao_idl File Error Messages

iguessthislldo:
1. The error message for failing to open an IDL file didn't have the filename in it. This was because the `ACE_ERROR` for it was using a `%m` when it looks like it used to use a `%p`. It looks like https://github.com/DOCGroup/ACE_TAO/pull/1420/commits/faab6830cf4ecf67b452183dba712caae7996be4 went one change too far.

2. First brought up here: https://github.com/objectcomputing/OpenDDS/issues/3308. Inputting a directory to tao_idl results in an empty file that causes a weird error latter. Changed it to ensure the file is a regular file using `fstat`.

### DIFF
--- a/TAO/TAO_IDL/driver/drv_preproc.cpp
+++ b/TAO/TAO_IDL/driver/drv_preproc.cpp
@@ -1119,6 +1119,27 @@ DRV_stripped_name (char *fn)
 void
 DRV_pre_proc (const char *myfile)
 {
+  // Check to see that the file is a normal file. If we don't and the file is a
+  // directory, then the copy would be blank and any error message later might
+  // not be helpful.
+  ACE_stat file_stat;
+  if (ACE_OS::stat (myfile, &file_stat) == -1)
+    {
+      ACE_ERROR ((LM_ERROR,
+                  "%C: ERROR: Unable to open file (stat) \"%C\": %m\n",
+                  idl_global->prog_name (),
+                  myfile));
+      throw Bailout ();
+    }
+  else if ((file_stat.st_mode & S_IFREG) != S_IFREG)
+    {
+      ACE_ERROR ((LM_ERROR,
+                  "%C: ERROR: This is not a normal file: \"%C\"\n",
+                  idl_global->prog_name (),
+                  myfile));
+      throw Bailout ();
+    }
+
   const char* tmpdir = idl_global->temp_dir ();
   static const char temp_file_extension[] = ".cpp";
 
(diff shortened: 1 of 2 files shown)


TAO/TAO_IDL/driver/drv_preproc.cpp
```diff
@@ -1188,15 +1189,47 @@ DRV_pre_proc (const char *myfile)
...
+      // Check to see that the file is a normal file. If we don't and the file
+      // is a directory, then the copy would be blank and any error message
+      // later wouldn't be accurate.
+      const ACE_HANDLE file_desc = ACE_OS::fileno (file);
+      if (file_desc == ACE_INVALID_HANDLE)
+        call_failed = "(fileno) ";
+      else
+        {
+          ACE_stat file_stat;
+          if (ACE_OS::fstat (file_desc, &file_stat) == -1)
```
mitza-oci, line 1208: ACE_OS has a `stat` function which does both of these in one operation. Any reason not to use it?

mitza-oci (changes requested):

TAO/TAO_IDL/driver/drv_preproc.cpp
iguessthislldo: Lack of knowledge of said function.

iguessthislldo: I don't see a `stat` function that takes a `FILE*`, only the one that takes the path.

mitza-oci: I thought it would start with `myfile` since that's the input to the function.   I guess I should have picked a different line to write the comment on.

jwillemsen:
MinGW fails with these changes

mitza-oci:
> MinGW fails with these changes

Probably an ACE bug for that port?

iguessthislldo:
> > MinGW fails with these changes
> 
> Probably an ACE bug for that port?

Maybe, though I don't really want to debug mingw, so I think I'll try to do just a straight stat at the beginning of the function. If that still doesn't work I'm going to make an exception for mingw since this is just an extra check.

jwillemsen:
Ok, can you also create issues as todo, maybe someone else encounters this also and then an issue will help and maybe we can remove the checks from taoidl later

iguessthislldo:
The MINGW64 job is still running, but it got far enough that I know it worked, but the CodeQL thing didn't like that it's now calling `fopen` after `stat`.

TAO/TAO_IDL/driver/drv_preproc.cpp
```diff
@@ -1118,6 +1119,27 @@ DRV_stripped_name (char *fn)
...
+      ACE_ERROR ((LM_ERROR,
+                  "%C: ERROR: Unable to open file (stat) \"%C\": %m\n",
+                  idl_global->prog_name (),
+                  myfile));
+      throw Bailout ();
+    }
+  else if ((file_stat.st_mode & S_IFREG) != S_IFREG)
+    {
+      ACE_ERROR ((LM_ERROR,
+                  "%C: ERROR: This is not a normal file: \"%C\"\n",
```
mitza-oci, line 1137:
```suggestion
                  "%C: ERROR: This is not a regular file: \"%C\"\n",
```

mitza-oci (commented):
NEWS file has a conflict

mitza-oci (approved):

jwillemsen (approved):

PR #1794, opened 2022-02-24
M1 iOS simulator compatibility

ClaytonCalabrese:
This PR adds iOS Simulator compatibility on M1 systems such as the 2021 by allowing the `IPHONE_TARGET` to be `M1_SIMULATOR`.  SImply changing the option passed to `-arch` to `arm64` was not sufficient for clang++ to operate properly, it needed the `-target` to be set instead.

### DIFF
--- a/ACE/include/makeinclude/platform_macosx_iOS.GNU
+++ b/ACE/include/makeinclude/platform_macosx_iOS.GNU
@@ -10,7 +10,7 @@ else
 endif
 
 ifndef IPHONE_TARGET
-  $(error Please set IPHONE_TARGET to SIMULATOR or HARDWARE)
+  $(error Please set IPHONE_TARGET to SIMULATOR, M1_SIMULATOR, or HARDWARE)
 endif
 
 ifeq ($(IPHONE_TARGET), SIMULATOR)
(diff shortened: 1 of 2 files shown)


ACE/include/makeinclude/platform_macosx_iOS.GNU
```diff
@@ -10,15 +10,24 @@ else
 endif
 
 ifndef IPHONE_TARGET
-  $(error Please set IPHONE_TARGET to SIMULATOR or HARDWARE)
+  $(error Please set IPHONE_TARGET to SIMULATOR, M1_SIMULATOR, or HARDWARE)
```
iguessthislldo, line 13: It seems like it would be easier for the user if we detected it. Like what they're talking about [here](https://stackoverflow.com/questions/65259300/detect-apple-silicon-from-command-line).

Also using the name "M1" in the NEWS is probably fine, but if we need a name for it here we should go with a different name. Going by [Apple's naming conventions for their processors](https://en.wikipedia.org/wiki/Apple_silicon#A_series) it's probably not going to be accurate to call it "M1" for long.

iguessthislldo (changes requested):

ACE/include/makeinclude/platform_macosx_iOS.GNU
mitza-oci: Right, this should use the target instruction set architecture `arm64` instead of the name of any specific processor

PR #1803, opened 2022-03-22
Get signal information from PerlACE

sonndinh:


jwillemsen:
Is this something that also needs to be added to master?

mitza-oci:
> > Is this something that also needs to be added to master?
> 
> Probably not. I think it's something specific to OpenDDS.

Yes, changes need to get merged to master

sonndinh:
> > > Is this something that also needs to be added to master?
> > 
> > 
> > Probably not. I think it's something specific to OpenDDS.
> 
> Yes, changes need to get merged to master

Should I make the PR target master instead of the ace6tao2 branch?

mitza-oci:
> > > > Is this something that also needs to be added to master?
> > > 
> > > 
> > > Probably not. I think it's something specific to OpenDDS.
> > 
> > 
> > Yes, changes need to get merged to master
> 
> Should I make the PR target master instead of the ace6tao2 branch?

No, we can make another PR to do that after the ace6tao2 changes are tested

ACE/bin/PerlACE/Process_Unix.pm
```diff
@@ -573,16 +573,17 @@ sub Spawn ()
     return 0;
 }
 
-sub WaitKill ($)
+sub WaitKill ($;$)
 {
     my $self = shift;
     my $timeout = shift;
+    my $opt_signum = shift;
```
iguessthislldo, line 580: This should probably be a hashref like I showed [here](https://github.com/objectcomputing/OpenDDS/pull/3416#discussion_r832467568). That would allow us to add more arguments to function without forcing a caller to be concerned with a signal ref argument if they don't need that info.

ACE/bin/PerlACE/Process_Unix.pm
```diff
@@ -656,13 +660,16 @@ sub check_return_value ($)
...
-        $rc >>= 8;
-        return $rc;
+        return ($rc >> 8);
     }
 
     # Ignore NSK 16-bit completion code
     $rc &= 0xff if $is_NSK;
 
+    if (defined($opt_signum)) {
+        ${$opt_signum} = $rc;
```
iguessthislldo, line 670: Related to the other comment I think we should the separate signal and coredump info, so the caller doesn't have to do bitwise operations.

iguessthislldo (changes requested):

iguessthislldo (approved):

ACE/bin/PerlACE/Process_Unix.pm
```diff
@@ -573,16 +573,17 @@ sub Spawn ()
     return 0;
 }
 
-sub WaitKill ($)
+sub WaitKill ($;$)
```
jwillemsen, line 576: Maybe add some documentation about the signature, to these and other changed operations?

sonndinh: I've added comments for these functions.

PR #1815, opened 2022-04-04
Get signal information from PerlACE

sonndinh:


iguessthislldo:
Follow up to #1803 and #1810

iguessthislldo (approved):

ACE/bin/PerlACE/Process_Unix.pm
```diff
@@ -671,11 +679,24 @@ sub check_return_value ($)
         $dump = 1;
     }
 
-    # check for ABRT, KILL or TERM
-    if ($rc == 6 || $rc == 9 || $rc == 15) {
+    # A undef means the process does not self crash
+    my $self_crash = $opts->{self_crash};
```
jwillemsen, line 683: Please add some documentation for this new `self_crash` option. There are also TAO tests that use self crashes, would they need an update also?

sonndinh: The comment on top of this function has a description of this option.

Can you point me to the TAO tests that do self crashes? Thanks.

jwillemsen: I don't see `self_crash` being explained in any description, it is only used, but not explained. I do know there are TAO tests that let themself crash, but no idea which one, you would have to search, we have thousands of tests

sonndinh: @jwillemsen @mitza-oci I've gone through the TAO and ACE tests that I think need to be updated. Let me know if I miss any other tests.

jwillemsen (approved):
I assume you have done a full branch build with running all tests to make sure you updated all tests?

ACE/bin/PerlACE/Process_Unix.pm
sonndinh: > I assume you have done a full branch build with running all tests to make sure you updated all tests?
After running the whole test suite, I indeed found a few more tests that deliberately send SIGKILL to themself. So I fixed them in the latest commit. I attached the autobuild's report file that has the failed tests (I can't upload the original HTML file, so used a PDF file instead).

The only test left that failed in the report that hasn't been changed in the latest commit is `TAO/orbsvcs/tests/ImplRepo/run_test.pl both_ir`, which received SIGABRT. I tried to search the code for a place where `abort()` is called but couldn't find it. Also, I ran that test again a few times and it passed. So I'm not sure if the received SIGABRT in the autobuild's report for this test was due to an actual error. And if so, this test won't need any update regarding the changes to PerlACE.

[Daily Build Log (Brief).pdf](https://github.com/DOCGroup/ACE_TAO/files/8475295/Daily.Build.Log.Brief.pdf)

mitza-oci:
> I assume you have done a full branch build with running all tests to make sure you updated all tests?

See test log in https://github.com/DOCGroup/ACE_TAO/pull/1815#discussion_r848720355 (note that this was before the [c59927e](https://github.com/DOCGroup/ACE_TAO/pull/1815/commits/c59927e3016c817114e5572d0d9bcfb5fe94ac66) commit)

I went through the log and it looks good to me.  @jwillemsen please let us know if you'd like to review in detail before we merge

jwillemsen:
I think `Deferred_Transient` is failing because of a bug

mitza-oci:
> I think `Deferred_Transient` is failing because of a bug in TAO, see https://buildlogs.remedy.nl/rhel80_acetao_release/index.html also.

That may be a false positive.  In TAO 3 the run_test.pl has `-ORBVerboseLogging 1` which is not there in TAO 2.  Due to this setting, LM_ERROR is part of the log which gets parsed as an error by the scoreboard scripts.

jwillemsen:
> > I think `Deferred_Transient` is failing because of a bug in TAO, see https://buildlogs.remedy.nl/rhel80_acetao_release/index.html also.
> 
> That may be a false positive. In TAO 3 the run_test.pl has `-ORBVerboseLogging 1` which is not there in TAO 2. Due to this setting, LM_ERROR is part of the log which gets parsed as an error by the scoreboard scripts.

I was more under the impression that the problem was just not visible in the past due to the lack of logging, but can't remember all details

PR #1819, opened 2022-04-17
md5sum and sha256sum

vermaete:
* Strip the directory from the filename in the MD5 checksum file.
  Otherwise the md5sum --check does not work.

* Create the sha256 checksum files

Signed-off-by: Jan Vermaete <jan.vermaete@gmail.com>

### DIFF
--- a/ACE/bin/make_release.py
+++ b/ACE/bin/make_release.py
@@ -637,7 +637,8 @@ def move_packages (name, stage_dir, package_dir):
     print ("\tZip file...")
     target_file = join (package_dir, name + ".zip")
     shutil.copy (join (stage_dir, "zip-archive.zip"), target_file)
-    ex ("md5sum " + target_file + " > " + target_file + ".md5")
+    ex ("cd " +  package_dir + "; md5sum " + os.path.basename(target_file) + " > " + os.path.basename(target_file) + ".md5")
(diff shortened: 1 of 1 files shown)


ACE/bin/make_release.py
```diff
@@ -637,7 +637,8 @@ def move_packages (name, stage_dir, package_dir):
     print ("\tZip file...")
     target_file = join (package_dir, name + ".zip")
     shutil.copy (join (stage_dir, "zip-archive.zip"), target_file)
-    ex ("md5sum " + target_file + " > " + target_file + ".md5")
+    ex ("cd " +  package_dir + "; md5sum " + os.path.basename(target_file) + " > " + os.path.basename(target_file) + ".md5")
```
jwillemsen, line 640: Why not use `name`, target_file is build up of a directory and filename, when you put the filename+extension in a new variable you can use that

jwillemsen:
ACE/bin/generate_rel_manpages and ACE/docs/bczar/bczar.html also need to be updated

jwillemsen:
In bzar.html add the additional tool required (even better would be to let make_release.py check if the required tools are there before starting), and update the copy commands to upload

jwillemsen (changes requested):
See comments

jwillemsen:
bin/copy-local-script.sh and bin/copy-script.sh needs an update also

PR #1812, opened 2022-04-01
Debian11 with gcc10

shuston:
Work in progress.

Bringing a customer's changes in to current dev to support Debian 11 w/ gcc10.

PR #1825, opened 2022-04-25
Restore getpid() result cast to int.

shuston:
Resolves compile warning resulting from PR #1812


shuston:
> Can't we use `= default` for the operator= and explicit copy constructor, wouldn't the default compiler version work?

It might; I added code consistent with the existing usage to satisfy warnings. This is (as commented) performance sensitive code so I am hesitant to change it.

shuston:
> Can't we use `= default` for the operator= and explicit copy constructor, wouldn't the default compiler version work?

Now that I study it more, I think the defaulted one would not work better - the added implementations call up to ACE_Atomic_Op_GCC methods that are implemented with intrinsics to invoke special behavior. I'm doubtful that the defaulted methods would know to do that. I'm going to leave it as-is.

PR #1826, opened 2022-04-25
gnuace: Mark `__postbuild__` Target as PHONY

iguessthislldo:
To prevent it from being created as a file when `make --touch` is used.


mitza-oci:
The title and commit comment say it's for postbuild, but the patch is for prebuild.  Should it be both?  

The prebuild target is mentioned in ACE/include/makeinclude/rules.common.GNU

mitza-oci (approved):

PR #1838, opened 2022-05-17
Windows: Supporting Serial Port 

TheGoesen:
Motivation:
We use ace for communications using serial ports.  We do get an error code 10038. [10038 socket error](https://stackoverflow.com/questions/3948164/10038-socket-error)
To my understanding this is caused by ACE using WASend, which simply doesnt work with Serial Ports.
Implementation:
Add a macro to force usage of the WriteFile which does work with Serial Ports. There may be better ways to do this, I am not a windows c programmer...

### DIFF
--- a/ACE/ace/README
+++ b/ACE/ace/README
@@ -100,6 +100,8 @@ ACE_TLI_TCP_DEVICE                      Device the platform uses for TCP on
                                         TLI.  Only needed if not /dev/tcp.
 ACE_USE_POLL                            The OS platform supports the
                                         poll() event demultiplexor
+ACE_USE_WIN32_WRITE_FILE                Uses WriteFile instead of WSASend. This works for 
+                                        serial Ports (COM1...)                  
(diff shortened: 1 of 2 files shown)


jwillemsen:
Wouldn't this change slowdown the use case where you don't use a serial port and set this define? Just the fact that serial ports are used shouldn't slowdown all use cases

TheGoesen:
Yes It would. I gave it some thought if it would make sense to dynamically check the pointer if it is something WASend supports but it appears to me that it would be counterproductive, as that would add more overhead, so the opposite of performance.

To me it seems a bit like: either you have a serial port usecase, than you don't care about performance, because the comport is slow as hell anyways. Or you don't use a comport, than you don't want additional overhead just because some dude has one. And how likely is it that you need to support a serial port and an ultra high performant tcp stack in the same application?...

I am not opposed to do this differently, I simply don't know how much and why and in which Situations WASend is relevantly better. If someone knows of an almost "free" way to check if the handle is something WASends supports, or a better way to deal with Serial Ports, I am all for it.

PR #1843, opened 2022-05-23
Added fix for compiling on gcc12

### LINKED ISSUE #1832: TAO fails to compile using gcc12
### Version

Master branch (and some of the versions used for OpenDDS)

### Host machine and operating system

Fedora 36

### Compiler name and version (including patch level)

`gcc-12.0.1`
`clang-14.0.0`

### The $ACE_ROOT/ace/config.h file

`config-linux.h`

### The $ACE_ROOT/include/makeinclude/platform_macros.GNU file
**This was taken from the github workflow**
```
ipv6=1
xerces3=1
ssl=1
INSTALL_PREFIX =/home/tyler/oss/ACE_TAO/installed
CCFLAGS+=-std=c++20
include /home/tyler/oss/ACE_TAO/ACE/include/makeinclude/platform_linux.GNU
```

### AREA/CLASS/EXAMPLE AFFECTED:
Compilation fails here: 
../../tao/Any_Insert_Policy_T.h:42:14: error: no viable overloaded '<<='

### The problem effects:

Fails to compile

### Synopsis
Compilation fails here: 
../../tao/Any_Insert_Policy_T.h:42:14: error: no viable overloaded '<<='

### Description

There's some more information in this issue on the OpenDDS: 
https://github.com/objectcomputing/OpenDDS/issues/3495

### Repeat by
Seems to be an issue with gcc12

I'd like to help fix this but I'm not sure how to go about that with this one

tmayoff:
This closes #1832

### DIFF
--- a/ACE/ace/config-g++-common.h
+++ b/ACE/ace/config-g++-common.h
@@ -16,7 +16,12 @@
 
 #define ACE_TEMPLATES_REQUIRE_SOURCE
 
+#define GCC_VERSION (__GNUC__ * 10000 \
+                     + __GNUC_MINOR__ * 100 \
+                     + __GNUC_PATCHLEVEL__)
+#if GCC_VERSION > 120000
 #define ACE_ANY_OPS_USE_NAMESPACE
+#endif
 
 #if __cplusplus >= 201103L
 # define ACE_HAS_CPP11
--- a/ACE/ace/config-g++-common.h
+++ b/ACE/ace/config-g++-common.h
@@ -15,11 +15,8 @@
 #define ACE_CC_BETA_VERSION (0)
 
 #define ACE_TEMPLATES_REQUIRE_SOURCE
-
-#define GCC_VERSION (__GNUC__ * 10000 \
-                     + __GNUC_MINOR__ * 100 \
-                     + __GNUC_PATCHLEVEL__)
-#if GCC_VERSION >= 120000
+                
+#if ACE_CC_MAJOR_VERSION >= 12
 #define ACE_ANY_OPS_USE_NAMESPACE
 #endif
 
--- a/ACE/ace/config-g++-common.h
+++ b/ACE/ace/config-g++-common.h
@@ -15,7 +15,7 @@
 #define ACE_CC_BETA_VERSION (0)
 
 #define ACE_TEMPLATES_REQUIRE_SOURCE
-                
+
 #if ACE_CC_MAJOR_VERSION >= 12
 #define ACE_ANY_OPS_USE_NAMESPACE
 #endif
--- a/ACE/ace/config-g++-common.h
+++ b/ACE/ace/config-g++-common.h
@@ -17,7 +17,7 @@
 #define ACE_TEMPLATES_REQUIRE_SOURCE
 
 #if ACE_CC_MAJOR_VERSION >= 12
-#define ACE_ANY_OPS_USE_NAMESPACE
+# define ACE_ANY_OPS_USE_NAMESPACE
 #endif
 
 #if __cplusplus >= 201103L
--- a/ACE/ace/config-g++-common.h
+++ b/ACE/ace/config-g++-common.h
@@ -19,7 +19,7 @@
 #define GCC_VERSION (__GNUC__ * 10000 \
                      + __GNUC_MINOR__ * 100 \
                      + __GNUC_PATCHLEVEL__)
-#if GCC_VERSION > 120000
+#if GCC_VERSION >= 120000
 #define ACE_ANY_OPS_USE_NAMESPACE
 #endif
 
--- a/ACE/ace/config-g++-common.h
+++ b/ACE/ace/config-g++-common.h
@@ -16,6 +16,8 @@
 
 #define ACE_TEMPLATES_REQUIRE_SOURCE
 
+#define ACE_ANY_OPS_USE_NAMESPACE
+
 #if __cplusplus >= 201103L
 # define ACE_HAS_CPP11
 # define ACE_FALLTHROUGH [[gnu::fallthrough]]
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -61,6 +61,12 @@ jobs:
             optional_macros: CCFLAGS+=-std=c++20
             platform_file: include $(ACE_ROOT)/include/makeinclude/platform_linux.GNU
             os: ubuntu-20.04
+          - CC: gcc-12
+            CXX: g++-12
+            PackageDeps: g++-12
+            optional_macros: CCFLAGS+=-std=c++20
+            platform_file: include $(ACE_ROOT)/include/makeinclude/platform_linux.GNU
+            os: ubuntu-22.04
           - CC: clang-5.0
             CXX: clang++-5.0
             PackageDeps: clang-5.0
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -64,7 +64,6 @@ jobs:
           - CC: gcc-12
             CXX: g++-12
             PackageDeps: g++-12
-            Repo: ppa:ubuntu-toolchain-r/ppa
             optional_macros: CCFLAGS+=-std=c++20
             platform_file: include $(ACE_ROOT)/include/makeinclude/platform_linux.GNU
             os: ubuntu-22.04


ACE/ace/config-g++-common.h
```diff
@@ -16,6 +16,13 @@
 
 #define ACE_TEMPLATES_REQUIRE_SOURCE
 
+#define GCC_VERSION (__GNUC__ * 10000 \
```
jwillemsen, line 19: As you add a new macro to ACE please add ACE_ prefix, add some examples so that people know how to use it, and check if there are more places where we can use this macro

tmayoff: I removed that define I didn't realize gcc version is already defined right above it

ACE/ace/config-g++-common.h
```diff
@@ -16,6 +16,10 @@
 
 #define ACE_TEMPLATES_REQUIRE_SOURCE
 
+#if ACE_CC_MAJOR_VERSION >= 12
+#define ACE_ANY_OPS_USE_NAMESPACE
```
jwillemsen, line 20:
``` suggestion
# define ACE_ANY_OPS_USE_NAMESPACE
```

tmayoff:
Okay so I'm at a bit of a loss here, the changes made work on my Fedora 36 machine, as well as an Ubuntu 22.04 VM, but not on Github Actions, feels like I'm missing something obvious

mitza-oci:
> Okay so I'm at a bit of a loss here, the changes made work on my Fedora 36 machine, as well as an Ubuntu 22.04 VM, but not on Github Actions, feels like I'm missing something obvious

Are your local builds including the IDL_Test directory?  The project that's failing is `IDL_Test_DLL`.

Also it may be useful to have a build with a compiler that's already supported that enables `ACE_ANY_OPS_USE_NAMESPACE`.

tmayoff:
Okay so I retried following the workflow, it works on Fedora but not the Ubuntu VM, the only difference seem to be that fedora has g++-12.1.1 and Ubuntu 12.0.1

mitza-oci:
> Okay so I retried following the workflow, it works on Fedora but not the Ubuntu VM, the only difference seem to be that fedora has g++-12.1.1 and Ubuntu 12.0.1

According to GCC upstream, if it's less than 12.1 then it's a pre-release version.
There appears to be a request here https://bugs.launchpad.net/ubuntu/+source/gcc-12/+bug/1970233 to update the Ubuntu 22.04 package to the official release.

If there's another package repository (like a PPA) with an updated compiler, we could try to use it in the GitHub Actions build.

jwillemsen:
Or we can just wait a little bit until ubuntu has updated their package

mitza-oci:
The `linux.yml` GitHub Actions script has a line that includes `apt-add-repository` for installing clang.  You could try adding a similar line but using `ppa:ubuntu-toolchain-r/ppa` -- see https://launchpad.net/~ubuntu-toolchain-r/+archive/ubuntu/ppa?field.series_filter=jammy This should not need a separate `apt-key` step like we have for llvm.org

jwillemsen:
#1852 is for the apt-key step change

jwillemsen:
Looks MinGW64 now has the same problem, that is using gcc 12.1

mitza-oci:
@jwillemsen in this earlier revision it skipped adding the apt repo https://github.com/DOCGroup/ACE_TAO/runs/6777757278?check_suite_focus=true

mitza-oci:
I think we want something like:
```
    - name: Add LLVM apt repo ${{ matrix.Repo }}
      run: |
        wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
        sudo apt-add-repository "deb http://apt.llvm.org/$(lsb_release -cs)/ ${{ matrix.Repo }} main"
      if: startsWith (matrix.Repo, 'llvm-toolchain')
    - name: Add PPA apt repo ${{ matrix.Repo }}
      run: |
        sudo apt-add-repository ${{ matrix.Repo }}
      if: startsWith (matrix.Repo, 'ppa:')
```

...along with restoring the `Repo` matrix config line for g++-12

jwillemsen:
Did try this between some other jobs. The fact that mingw64 also fails now with gcc12.1 I think we need some more code changes, not just action changes

mitza-oci:
> Did try this between some other jobs. The fact that mingw64 also fails now with gcc12.1 I think we need some more code changes, not just action changes

I build locally on Ubuntu with `gcc version 12.1.0 (GCC)` and @tmayoff build on Fedora.  So I think we can get that to work independent of MinGW.

mitza-oci:
I build locally with `gcc version 12.1.0 (Rev2, Built by MSYS2 project)` and it compiled IDL_Test successfully. Was another step failing?

jwillemsen:
Do you have versioned namespaces enabled, that feature is enabled in these builds

jwillemsen:
I thought clang does use `ACE_ANY_OPS_USE_NAMESPACE` already

mitza-oci:
> I think clang does use `ACE_ANY_OPS_USE_NAMESPACE` already

It's in a few of the `config-macosx-*.h` files but not for clang on Linux.  Not sure if that difference would matter, or if clang and gcc behave differently.  We've also seen difference's between Apple's clang and LLVM's clang.

mitza-oci:
Looks like clang works because it has `-fdelayed-template-parsing`

mitza-oci:
Repro on godbolt using GCC 12.1.  Looks like it can be fixed if we change the order of code generation in tao_idl so that the any insertion (etc.) operators are generated before the traits.  Not sure if that would have unintended consequences, but I don't know why they wouldn't be declared right after the struct (or other thing) that's being inserted.

https://godbolt.org/z/csj8WMEbY

mitza-oci:
> Maybe we can get rid of `ACE_ANY_OPS_USE_NAMESPACE` at the same moment?

I take you mean that TAO would always generate the Any-related operator overloads in the namespace of the mapped types as if `ACE_ANY_OPS_USE_NAMESPACE` was defined?

jwillemsen:
> > Maybe we can get rid of `ACE_ANY_OPS_USE_NAMESPACE` at the same moment?
> 
> I take you mean that TAO would always generate the Any-related operator overloads in the namespace of the mapped types as if `ACE_ANY_OPS_USE_NAMESPACE` was defined?

Yes, I created #1858 as a PR to always define `ACE_ANY_OPS_USE_NAMESPACE`, when that works on github actions we can merge that and cleanup the code in TAO_IDL

mitza-oci:
I have theory on what's wrong with the MinGW build.  No idea why it's not behaving the same as the other GCC 12.1 builds though.  When the IDL type is in the global module, even with ACE_ANY_OPS_USE_NAMESPACE, tao_idl is generating the Any operators in the versioned namespace.  Instead it should use the global namespace -- it's not a problem for versioning since the mangled name has `CORBA::Any`'s namespace.

mitza-oci:
I started working on changes to tao_idl to make the declaration of the Any operators more directly part of the client header generation, as in `TAO_CodeGen::TAO_ROOT_CH`, instead of its own pass.

So when this is done, the generated header should look like:
```
BEGIN_USER_VN
namespace IDL_Module {
  struct IDL_Struct {
  };
  void operator<<=(::CORBA::Any&, const IDL_Struct&);
}
END_USER_VM
BEGIN_TAO_VN
// traits, must come after operator<<=
END_TAO_VN
```

It's not really far enough along to be its own PR yet.  There are interesting corner cases like the enums defined inside unions in Bug_3312_Regression.  It's also not clear if this is the right approach.

tmayoff:
Is it worth at all, closing this PR with just the added g++12 CI test (remove the config change), then continue the conversation in the linked issue until a new PR is actually ready?

mitza-oci:
> Is it worth at all, closing this PR with just the added g++12 CI test (remove the config change), then continue the conversation in the linked issue until a new PR is actually ready?

Perhaps... what kind of errors do we get when it doesn't have `ACE_ANY_OPS_USE_NAMESPACE`?
I tried the Ubuntu PPA for GCC 12.1 on my fork https://github.com/mitza-oci/ACE_TAO/tree/g%2B%2B12 and it worked (as in, it installed and used 12.1, not that it passed the tests)

tmayoff:
Seems so, along with a bunch of others types, 
[make-output.txt](https://github.com/DOCGroup/ACE_TAO/files/8953164/make-output.txt)

PR #1858, opened 2022-06-15
From ACE 7.0.8 we assume all compilers support ACE_ANY_OPS_USE_NAMESPACE

jwillemsen:
* ACE/ace/config-all.h:


jwillemsen:
Ok, this doesn't work yet, maybe when the order has been changed we can try this again

PR #1868, opened 2022-06-27
Updated sequence Any operators code generation

mitza-oci:
- Fixed the alternate mapping (std::vector)
- Use leading :: to avoid naming collision with CORBA namespace

### DIFF
--- a/TAO/TAO_IDL/be/be_visitor_sequence/any_op_cs.cpp
+++ b/TAO/TAO_IDL/be/be_visitor_sequence/any_op_cs.cpp
@@ -42,8 +42,6 @@ be_visitor_sequence_any_op_cs::visit_sequence (be_sequence *node)
 
   TAO_INSERT_COMMENT (os);
 
-  *os << be_global->core_versioning_begin () << be_nl;
-
   // These are no-ops for now, so we just generate them and return
   if (be_global->alt_mapping () && node->max_size ()->ev ()->u.ulval == 0)
     {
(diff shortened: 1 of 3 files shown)


TAO/TAO_IDL/be/be_visitor_sequence/any_op_cs.cpp
```diff
@@ -189,14 +191,14 @@ be_visitor_sequence_any_op_cs::visit_sequence (be_sequence *node)
       << "/// Copying insertion." << be_nl
       << "void operator<<= (" << be_idt_nl
       << "::CORBA::Any &_tao_any," << be_nl
-      << "const " << node->name () << " &_tao_elem)"
+      << "const ::" << node->name () << " &_tao_elem)"
       << be_uidt_nl
       << "{" << be_idt_nl
-      << "TAO::Any_Dual_Impl_T<" << node->name () << ">::insert_copy ("
+      << "::TAO::Any_Dual_Impl_T< ::" << node->name () << ">::insert_copy ("
```
jwillemsen, line 197: Do we really need `::TAO` here, in the core we do use `TAO::Any_Dual_Impl_T` without problems

mitza-oci: Is there a downside to it?  I haven't tried this but it seems like a user's IDL could cause a name collision.

jwillemsen: Not sure, just something I observed. I have restarted my TAOX11 test build to make sure it all works there, we depend on versioned namespaces. Will let you know the results tomorrow, already 9PM here ;-)

jwillemsen (approved):

PR #1879, opened 2022-07-10
Fixes for Static Initialization Order Issues Reported By AddressSanitizer

simpsont-oci:
Problem: Some of the Synch_Options static member variables depend (via default parameters) on some of the Time_Value static member variables. Unfortunately, the order of static variable initialization between compilation units is indeterminate, which means that sometimes these will be initialized in the wrong order. AddressSanitizer's stricter checks for initialization found such an instance and reported it.

Solution: There are probably multiple ways to address this, but perhaps the most straightforward way is to break the direct dependency between static member variables by having Synch_Option static member variables use ACE_Time_Value() rather than ACE_Time_Value::zero_value(), which amounts to the same value.


mitza-oci (commented):
Would it be better to change the ACE_Synch_Options constructor declaration?  That may help other users of this class.

ACE/ace/Synch_Options.h
```diff
@@ -84,7 +84,7 @@ class ACE_Export ACE_Synch_Options
    *                ACE_Synch_Options makes no use of it internally.
    */
   ACE_Synch_Options (unsigned long options = 0,
-                     const ACE_Time_Value &timeout = ACE_Time_Value::zero,
+                     const ACE_Time_Value &timeout = ACE_Time_Value(),
```
mitza-oci, line 87:
```suggestion
                     const ACE_Time_Value &timeout = ACE_Time_Value (),
```

mitza-oci (approved):

Issue #1906, opened 2022-08-17
Fixes to enable compiling with -Werror=format-security


lukem:
### Version

ACE 6.5.11 / TAO 2.5.11.
Issues still appear to be present in present in github master branch as at commit 280392c on 2022-08-16.

### Host machine and operating system

RHEL 9.0
Built using specfile based on that in ace-tao-6.5.11-82.1.src.rpm.

### Compiler name and version (including patch level)

g++ 11.2.1

### The $ACE_ROOT/include/makeinclude/platform_macros.GNU file

Adding the following compiler flags will trigger the issue.
```
CCFLAGS += -Werror=format-security
CFLAGS += -Werror=format-security
```

I actually experienced it when using the full suite of RHEL 7/8/9 compiler-hardening flags enabled in the rpm spec file with:
```
cat >> $ACE_ROOT/include/makeinclude/platform_macros.GNU <<EOF
CCFLAGS += ${RPM_OPT_FLAGS}
CFLAGS += ${RPM_OPT_FLAGS}
LDFLAGS += ${RPM_LD_FLAGS}
EOF
```

(See https://src.fedoraproject.org/rpms/redhat-rpm-config/blob/rawhide/f/buildflags.md . I'm using the older synonyms so the flags are added on RHEL 7 as well).

### AREA/CLASS/EXAMPLE AFFECTED:

The following TAO files fail to compile due to format string issues detected by -Werror=format-security:

- TAO/orbsvcs/ImplRepo_Service/ImR_Activator_i.cpp
- TAO/orbsvcs/FTRT_Event_Service/Factory_Service/EventChannelFactory_i.cpp

### Synopsis

Invoking methods that take a printf()-like format string with a non-literal string as the format string. This can be a security issue - see https://fedoraproject.org/wiki/Format-Security-FAQ

### Description

Methods such as `ACE_Process_Options::command_line()` have an overload:
```
int command_line (const ACE_TCHAR *format, ...) ACE_GCC_FORMAT_ATTRIBUTE (printf, 2, 3);
```

[Line 540 of ImR_Activator_i::start_server() in ImR_Activator_i.cpp](https://github.com/DOCGroup/ACE_TAO/blob/master/TAO/orbsvcs/ImplRepo_Service/ImR_Activator_i.cpp#L540) calls this as:
```
proc_opts.command_line (ACE_TEXT_CHAR_TO_TCHAR(cmdline));
```
where the cmdline varible is passed into `start_server()`.

[...]

PR #1907, opened 2022-08-20
Fix -Wformat-security issues

lukem:
Ensure that methods with a printf()-like format string
are invoked with a literal string instead of a string
built from possibly untrusted sources.

Fixes issue #1906.


TAO/orbsvcs/FTRT_Event_Service/Factory_Service/EventChannelFactory_i.cpp
```diff
@@ -124,7 +124,7 @@ CORBA::Object_ptr EventChannelFactory_i::create_process (
 
   ORBSVCS_DEBUG((LM_DEBUG, "Command Line : %s\n", str.c_str()));
 
-  options.command_line(str.c_str());
+  options.command_line(ACE_TEXT("%s"), str.c_str());
```
mitza-oci, line 127: I'm not sure this is correct for wchar builds.
ACE_Process_Options will use `ACE_OS::vsnprintf(wchar_t*, size_t, const wchar_t*, va_list)` which is just `vswprintf` from C99/POSIX.  `vswprintf` can insert a `const wchar_t*` arg but uses the `%ls` formatter for that, not `%s`.  The `%s` formatter (although encoded within a wide string) tells it to read a `const char*` from the va_list.
But I could be missing something...

jwillemsen: Yes, this will not work, there is a macro for that as far as I remembe

lukem: I've added commit c320b80 which conditionally uses ACE_TEXT("%ls") instead of ACE_TEXT("%s"), using similar patterns seen elsewhere (including in ACE/ace/Process.cpp).

lukem: I considered adding overloaded member functions for `ACE_Process_Options::setenv()` and `ACE_Process_Options::command_line()`, but these would require different names because of ambiguous name resolution issues. (A C++11 workaround using the technique in https://stackoverflow.com/a/30761816 is most likely an ABI change, which I suspect is not desirable).

I have experimented with adding methods to ACE_Process_Options:
```
int setenv_value (const ACE_TCHAR *variable_name, const ACE_TCHAR *value);
int command_line_str (const ACE_TCHAR * args);
```
and adapting various callers of `setenv()` and `command_line()` appropriately.

I'm not sure that this is the best approach. Happy to consider different methods based on preferences of experienced ACE contributors.

lukem: I've tried a different approach instead of the overloaded member functions suggested above.

I've defined `ACE_TEXT_PRIs` to `"s"` or `"ls"`, depending upon the WCHAR mode. This similar to PRIxxx macros in <inttypes.h> (e.g., PRIu64). This does not contain the leading `%`, to allow the caller to use flags or width specifiers as required.

I've then replaced the existing -Wformat-security fixes appropriately.

For example, instead of:
```
options.command_line(ACE_TEXT("%s"), str.c_str());
```
it's now
```
options.command_line(ACE_TEXT("%") ACE_TEXT_PRIs, str.c_str());
```

This is in commit f5e5d7759a9 in my pull request.

lukem:
As noted in the thread above, I've refactored the change to:

- Define `ACE_TEXT_PRIs` to either `"s"` or `"ls"` depending upon the WCHAR mode. This is similar to the PRIxxx macros in <inttypes.h> in C99.
- Adapted my previous changes to use this.

Commit f5e5d7759a9 on branch lukem:format-security.

I'll push separate commits with other parts of the repository converted to use `ACE_TEXT_PRIs` instead of needing idioms such as:
```
#if !defined (ACE_WIN32) && defined (ACE_USES_WCHAR)
 printf(ACE_TEXT("%ls"), buf);
#else
 printf(ACE_TEXT("%s"), buf);
#endif
```

jwillemsen:
Let us know when these changes are ready

TAO/orbsvcs/FTRT_Event_Service/Factory_Service/EventChannelFactory_i.cpp
mitza-oci: I agree that having the PRIs macro available is better than not having it.  One complication of this is that it can't be used for format strings that are passed to `ACE_Log_Msg::log()`.  But I guess that's not directly relevant here.

But I suggest we also consider providing methods on these classes that accept strings (not format strings, plain strings) of either narrow or wide characters.  That's much easier to use correctly and it lets the type system enforce correctness instead of going through varargs.

lukem: @mitza-oci : Do you have suggestions on method names for direct string (non-format-string) methods based on existing ACE conventions? I don't have strong opinions on this, versus the names I experimented with above (`setenv_value()` and `command_line_str()`)

lukem:
> Let us know when these changes are ready

The `ACE_TEXT_PRIs` changes seem to be fine - the various CI builds work.

However, I could also add the string-only (without format strings and varargs) variations per @mitza-oci 's comments in the earlier thread. Thoughts?

TAO/orbsvcs/FTRT_Event_Service/Factory_Service/EventChannelFactory_i.cpp
mitza-oci: I'm OK with the names you've listed.  I don't think ACE has a convention for it.

mitza-oci (dismissed):

mitza-oci:
> However, I could also add the string-only (without format strings and varargs) variations per @mitza-oci 's comments in the earlier thread. Thoughts?

I merged a commit that resolves a conflict since it looked trivial.
We could defer the other changes to another PR.

lukem:
The branch is ready as far as I'm concerned.

mitza-oci:
> The branch is ready as far as I'm concerned.

OK, I reviewed it and something strange happened with GitHub which caused it to record "dismissed their [stale review](https://github.com/DOCGroup/ACE_TAO/pull/1907#pullrequestreview-1112911071)"

I'll go ahead and mark it approved.  Does anyone else want to review before we merge?

mitza-oci (approved):

jwillemsen:
Looks we have new compile errors, see https://buildlogs.remedy.nl/sles150_ace_i686_debug/index.html

```
/home/build/jenkins/workspace/sles150_ace_i686_debug/BUILD/DOC_ROOT/ACE/examples/OS/Process/GNUmakefile.OS_Process_Process

g++ -Wnon-virtual-dtor -fvisibility=hidden -fvisibility-inlines-hidden -O3 -ggdb -m32 -pthread -fno-strict-aliasing -Wall -Wextra -Wpointer-arith -pipe -D_GNU_SOURCE   -I/home/build/jenkins/workspace/sles150_ace_i686_debug/BUILD/DOC_ROOT/ACE -D__ACE_INLINE__ -I../../..  -c -o .obj/process.o /home/build/jenkins/workspace/sles150_ace_i686_debug/BUILD/DOC_ROOT/ACE/examples/OS/Process/process.cpp

/home/build/jenkins/workspace/sles150_ace_i686_debug/BUILD/DOC_ROOT/ACE/examples/OS/Process/process.cpp:460:76: error: macro "ACE_TEXT" passed 2 arguments, but takes just 1

   options.command_line (ACE_TEXT ("%") ACE_TEXT_PRIs ACE_TEXT (" -g", argv0);
                                                                            ^

/home/build/jenkins/workspace/sles150_ace_i686_debug/BUILD/DOC_ROOT/ACE/examples/OS/Process/process.cpp: In function ‘void test_setenv(const ACE_TCHAR*)’:
/home/build/jenkins/workspace/sles150_ace_i686_debug/BUILD/DOC_ROOT/ACE/examples/OS/Process/process.cpp:460:54: error: expected ‘)’ before ‘ACE_TEXT’

   options.command_line (ACE_TEXT ("%") ACE_TEXT_PRIs ACE_TEXT (" -g", argv0);
                                                      ^~~~~~~~

/home/build/jenkins/workspace/sles150_ace_i686_debug/BUILD/DOC_ROOT/ACE/examples/OS/Process/process.cpp:453:31: warning: unused parameter ‘argv0’ [-Wunused-parameter]

 test_setenv (const ACE_TCHAR *argv0)
                               ^~~~~

make[1]: *** [/home/build/jenkins/workspace/sles150_ace_i686_debug/BUILD/DOC_ROOT/ACE/include/makeinclude/rules.local.GNU:136: .obj/process.o] Error 1

make[1]: Target 'all' not remade because of errors.
make[1]: Leaving directory '/home/build/jenkins/workspace/sles150_ace_i686_debug/BUILD/DOC_ROOT/ACE/examples/OS/Process'
make[1]: Entering directory '/home/build/jenkins/workspace/sles150_ace_i686_debug/BUILD/DOC_ROOT/ACE/examples/OS/Process'
```

lukem:
Oops. I thought I managed to test the builds of all the examples too.

The fix should be trivial; add a ) between the " -g" and ,

Are you able to do that, or prefer me to do it with a new pull request?

jwillemsen:
Please open a new PR

lukem:
> Please open a new PR

Done! See #1943

PR #1911, opened 2022-08-29
Upgrade from Ubuntu 18.04 to 22.04 as 18.04 is getting deprecated on …

jwillemsen:
…github actions

    * .github/workflows/linux.yml:


jwillemsen:
22.04 doesn’t have gcc4.8

jwillemsen:
And no gcc6/7/8 and older clangs

PR #1913, opened 2022-08-29
Use ACE's Sync(h) traits in AST_Annotation_Appls

mitza-oci:
- interprocess mutexes are not needed
- support threads=0 builds of tao_idl


iguessthislldo (approved):

mitza-oci:
Is this fuzz check correct?
```
Running ACE_SYNCH_MUTEX check
Error: ./TAO/TAO_IDL/include/ast_annotation_appls.h:20: found ACE_SYNCH_MUTEX, use TAO_SYNCH_MUTEX instead
```
Since it's in TAO_IDL, this shouldn't be including TAO headers.

PR #1914, opened 2022-08-30
Fix warnings found by -Wsuggest-override

knutpett:
There are many, many files with this warning. Will update this PR several times, work in progress.


jwillemsen:
It should be virtual or override but not both

jwillemsen:
Make sure you update all operations that are overridden in a class, when you forget some the newer clang versions will give a lot of warnings

jwillemsen:
Removing virtual could mean you have to change the indent of multiple lines when the declaration is using multiple lines, all together a lot of work for a simple change

knutpett:
> Removing virtual could mean you have to change the indent of multiple lines when the declaration is using multiple lines, all together a lot of work for a simple change

It involves quite a lot of files. Would it be better to review and handle if I split this into 2-3 PRs? One for ACE, one for TAO etc...?

If i chose to do, can I just split locally and just force push the new commits?

jwillemsen:
Don't use force push, that makes it impossible for us to see what changed each time and also means we have to review everything again and again

ACE/ace/Asynch_IO_Impl.h
```diff
@@ -311,11 +311,11 @@ class ACE_Export ACE_Asynch_Read_File_Impl : public virtual ACE_Asynch_Read_Stre
 
   /// This starts off an asynchronous read. Upto @a bytes_to_read will
   /// be read and stored in the @a message_block.
-  virtual int read (ACE_Message_Block &message_block,
+  int read (ACE_Message_Block &message_block,
```
knutpett, line 314: Since I've modified this line Codacy has analysed the file and reported what I think is a latent error: https://app.codacy.com/gh/DOCGroup/ACE_TAO/pullRequest?prid=10073223

What should I do? Fixing that issue in this pull request will IMHO complicate this change. The idea was that this PR should only fix warnings related to override.

jwillemsen: It can be that codacy gives new warnings/errors, I think they normally only scan 5k number of files. Just don't fix them on this branch, just fix the warnings related to override

knutpett: Ok. CodeFactor also reports lots of whitespace issues. I think I've fixed most of them. Should I move them to another branch before I push again?

jwillemsen: Do you mean make a separate PR for just whitespace, sounds a good proposal, those will be easy to review ;-)

knutpett: Ok, then I'll do that. But, what about 6486d27457 (Fix CodeFactor issues: blank lines) which I've already pushed to this branch?

jwillemsen: Try to cherry pick that one to a new branch and submit it as separate PR?

jwillemsen:
Now this one has a huge number of merge conflicts

knutpett:
Still not finished. Trying to fix even more

ACE/ACEXML/common/DefaultHandler.h
```diff
@@ -55,117 +55,117 @@ class ACEXML_Export ACEXML_DefaultHandler
...
   // Methods inherit from ACEXML_ContentHandler.
 
   /*
    * Receive notification of character data.
    */
-  virtual void characters (const ACEXML_Char *ch,
+  void characters (const ACEXML_Char *ch,
                            size_t start,
-                           size_t length);
+                           size_t length) override;
```
jwillemsen, line 67: Can you also change the indent of splitted line when you remove the virtual?

jwillemsen:
Looks also the code generated by TAO_IDL needs to be updated, see for example a clang build log, a huge number of warnings given now

knutpett:
> Looks also the code generated by TAO_IDL needs to be updated, see for example a clang build log, a huge number of warnings given now

Are the warnings introduced by my change, or did they exist before?

I'm working on fixing TAO_IDL as well. It takes some time to fix since I use a lot of time figuring out which visitors are used to generate the code with the warnings. And it takes time to test. A lot of back and forth because I end up with adding override to something that should etc, etc. 

Do you prefer that we fix everything in this PR, or could we do this incrementally?

jwillemsen:
I think the warnings are there because of the changes, at the moment you use override for some operations in a class clang gives a warning that you should do all.

We could do this incrementally, but we shouldn't have an increase of warnings which could be impossible, the challenge is that we want to have master always in a state that we can release when needed, not that we have to wait a long time before we are back to the same level.

jwillemsen:
Closing, lot of merge conflicts due to changes on master

PR #1074, opened 2020-04-11
Support fstat larger than 4G while st_size is 64bit on Windows

likema:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved file size handling to ensure accurate reporting for large files that exceed 32-bit limits.

PR #1918, opened 2022-09-05
fixed fstat on win32 (support 64 bit file sizes)

esohns:


jwillemsen:
See also #1074

ACE/ace/OS_NS_sys_stat.inl
```diff
@@ -41,23 +41,28 @@ namespace ACE_OS
...
-        stp->st_mtime = ACE_Time_Value (fdata.ftLastWriteTime).sec ();
-        stp->st_ctime = ACE_Time_Value (fdata.ftCreationTime).sec ();
-        stp->st_nlink = static_cast<short> (fdata.nNumberOfLinks);
-        stp->st_dev = stp->st_rdev = 0; // No equivalent conversion.
-        stp->st_mode = S_IXOTH | S_IROTH |
-          (fdata.dwFileAttributes & FILE_ATTRIBUTE_READONLY ? 0 : S_IWOTH) |
-          (fdata.dwFileAttributes & FILE_ATTRIBUTE_DIRECTORY ? S_IFDIR : S_IFREG);
-      }
+      stp->st_size = fdata.nFileSizeLow;
+    stp->st_atime = ACE_Time_Value (fdata.ftLastAccessTime).sec ();
```
jwillemsen, line 58: Correct indent

esohns: ok.

esohns:
Note: the ACE_OS::stat function (same file lines 221) has the same issue for the WinCE case (ACE_HAS_WINCE). Not sure whether Windows CE ever supported 64 bits though... In that case, it might make more sense to use the alternate approach of 1074 and use the ACE macro (ACE_MAKE_QWORD) instead of my solution (ULARGE_INTEGER comes from winnt.h).

jwillemsen:
Is this the same as #2254?

esohns:
yes, more or less. See comment about WinCE support

jwillemsen:
Windows CE support has been removed on master some time ago

esohns:
what about using ACE_MAKE_QWORD or ACE_COMBINE_PARTS instead of using ULARGE_INTEGER (seems to be windows nt-specific ?)

PR #1919, opened 2022-09-06
make addr_family_ protected so it can be accessed from derived classes

esohns:


jwillemsen:
Reordering members will lead to reorder warnings on the constructor(s), please use uniform initialization and simplify the constructor accordingly, very likely the reorder will not lead to warnings at that moment

jwillemsen (changes requested):
Will lead to warnings

esohns:
ok

esohns:
sorry.

PR #1920, opened 2022-09-08
FIX proposal for a crash that happens because of some misuse of const_cast<>

Nibelheims:
Hello,
First of all, thank you for providing libACE.
This commit is a proposal to fix a crash that can happen with `ACE::sendv_n_i()` and `ACE::writev_n()`, in case the given `iovecs` have been allocated on the heap.
Indeed in case of a partial send, libACE will loop until all the data has been sent. The problem is that it updates its arguments when doing so, leaving the caller with one or more updated pointers inside the table of iovecs. Attempting to free such a pointer results in a fault (the value can even not be aligned properly anymore).
The fix simply copies the iovec data **if necessary** (no perf penalty otherwise), and updates those values instead of the arguments.
Thank you.

### DIFF
--- a/ACE/ace/ACE.cpp
+++ b/ACE/ace/ACE.cpp
@@ -1777,7 +1777,13 @@ ACE::sendv_n_i (ACE_HANDLE handle,
   size_t &bytes_transferred = bt == 0 ? temp : *bt;
   bytes_transferred = 0;
 
-  iovec *iov = const_cast<iovec *> (i);
+  const iovec *iov = i;
+  // as per its prototype, this function shall not modify i, even in case of
+  // partial send, when it is required to update iov->base
+  // in that case, this local copy will be used instead
+  // NOTE: the performance penalty of memory allocation only occurs if
+  // required
+  iovec *shadow_iovec = nullptr;
 
   for (int s = 0;
        s < iovcnt;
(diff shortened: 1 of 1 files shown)


Nibelheims:
Hello,
Thank you for considering this PR. As you can see the modification is small and simple. Do not hesitate to discuss it. I am open to modifications if required.
Thanks and best regards,
Nibelheims

jwillemsen:
What is exactly copied, only the iovec array but not the application data it refers to, correct? Any idea to extend one of the existing ACE unit tests as reproducer?

Nibelheims:
Hello,
Thank you for your reply. Sorry you are right, my phrasing was bad. Yes only the iovec array gets copied, not the data it refers to.
I am quite new to libACE, I actually had to investigate a crash on some application and it turned out that it fell in the "data cannot be sent at once" case. Hence this PR.
That application sends data over the network. I don't know if sending big chunks of data (bigger than the MTU at least) to a socket would be enough to reproduce. The easiest way would be to stub `ACE_OS::sendv` to make it perform a partial send on the input.
I did this PR on my spare time and, unfortunately, could not look at the testing code (not doing so is bad, sorry about that). I had a bit of trouble compiling, I first had to figure out that MPC (which I was unaware of) was required, then I had some trouble making it generating a Makefile. I finally could compile the shared object successfully, but I did not have enough time to go further.

PR #1921, opened 2022-09-08
Added support for VxWorks 21.03, including building for RTP mode.

ocielliottc:


ACE/ace/OS_NS_Thread.cpp
```diff
@@ -5141,7 +5141,8 @@ vx_execae (FUNCPTR entry, char* arg, int prio, int opt, size_t stacksz, ...)
   if (ret == ACE_VX_TASK_ID_ERROR)
     return 255;
 
-  while( ret > 0 && ::taskIdVerify (ret) != ERROR )
+  while( ret > reinterpret_cast <ACE_VX_TASK_ID> (0) &&
```
jwillemsen, line 5144: Why are we checking `ret` here? We check it before for `ACE_VX_TASK_ID_ERROR` and it is const, so will never change.

ocielliottc: Good point.  It doesn't appear to be necessary.

mitza-oci (approved):

jwillemsen (approved):

PR #1936, opened 2022-09-19
Valuetype test extensions

jwillemsen:


TAO/tests/IDL_Test/Versioned_Namespace/Versioned_Namespace.mpc
```diff
@@ -1,6 +1,5 @@
 // -*- MPC -*-
-project(IDL_Test_Versioned_Namespace): taoidldefaults, valuetype {
-  idlflags += -Scdr -Sa
+project(IDL_Test_Versioned_Namespace): taoidldefaults, valuetype, amh, gen_ostream {
```
mitza-oci, line 2: I suggest having two different projects that compile the same IDL source, one with the `-Scdr -Sa` options and one with other options.

PR #1942, opened 2022-09-22
Added const accessors for AST_Decl::local_name () and UTL_String::get_string ()

knutpett:
Small contribution to making it easier to write const correct code for users. Maybe just a drop in the ocean, but it is a start.


jwillemsen:
Why not just make the existing ones const?

knutpett:
> Why not just make the existing ones const?

You're absolutely right. New commit coming up :)

PR #1947, opened 2022-10-03
C++17 or greater implies lack of std::auto_ptr

simpsont-oci:
Problem: Build issues on macOS for setting standard to C++17 (and presumably c++20) due to lack of `std::auto_ptr`.

Solution: Be sure to set ACE_LACKS_AUTO_PTR when C++ standard 17 or greater is detected.

### DIFF
--- a/.github/workflows/macosx.yml
+++ b/.github/workflows/macosx.yml
@@ -16,8 +16,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        cxxstd: ["03", "11"]
-        os: [macos-10.15, macos-11.0]
+        cxxstd: ["03", "11", "17"]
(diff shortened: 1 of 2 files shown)


.github/workflows/macosx.yml
```diff
@@ -16,8 +16,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        cxxstd: ["03", "11"]
-        os: [macos-10.15, macos-11.0]
+        cxxstd: ["03", "11", "17"]
+        os: [macos-10.15, macos-11, macos-12]
```
mitza-oci, line 20:
```suggestion
        os: [macos-11, macos-12]
```

https://github.com/actions/runner-images/issues/5583

mitza-oci (approved):

PR #1991, opened 2022-12-09
[ace6tao2] Correction for wide-char builds

ocielliottc:


simpsont-oci (approved):

simpsont-oci:
Does master / ACE7 have the same issue?

ocielliottc:
It does.  I will be making this change there, as well.

PR #1999, opened 2022-12-14
[ace6tao2] Allow disabling of ACE_OS::readdir_r

ocielliottc:
Only use `readdir` internally since `readdir_r` is deprecated.

This change will not apply to ACE 7, since readdir_r was removed.


mitza-oci:
Should config-linux.h set ACE_DISABLE_READDIR_R based on the glibc version?

ocielliottc:
I thought we would leave it up to the user.

mitza-oci:
> I thought we would leave it up to the user.

OK, we could add to config-linux.h later if someone's motivated to do so.

mitza-oci (approved):

PR #2006, opened 2023-01-02
Fix error in ACE_Process_Options::inherit_environment ()

viktor-ch:
While using ACE_Wide_To_Ascii string conversion class (and correspondingly ::WideCharToMultiByte) there must be expected that the length of the output multi-byte string is not equal to the length of the input Unicode string. That's why we the variable _iter_ used to iterate over the original Unicode environment variables must be increased by the source string length (and not by the length of converted string).

### DIFF
--- a/ACE/ace/Process.cpp
+++ b/ACE/ace/Process.cpp
@@ -932,7 +932,7 @@ ACE_Process_Options::inherit_environment ()
           size_t const idx = temp_narrow_env.size ();
           temp_narrow_env.resize (idx + len + 1, 0);
           ACE_OS::strncpy (&temp_narrow_env[idx], wta.char_rep (), len);
-          iter += len;
+          iter += ACE_OS::strlen(iter);
         }
       temp_narrow_env.push_back (0);
       existing_environment = &temp_narrow_env[0];


jwillemsen:
Please extend one of the unit tests under ACE/tests in such a way that it reproduces the issue you found

viktor-ch:
The error was found in conjunction with Implementation Repository Activator. On one Chinese Windows machine the Activator could not start a managed server. 
Activator starts a server by calling Win API function ::CreateProcess. One of the parameters of ::CreateProcess is the list of environment variables.  While preparing the list the Activator:

1. reads the environment variables using ::GetEnvironmentStringsW as Unicode string
2. iterates over all variables and converts individual Unicode strings to multi-byte strings using ::WideCharToMultiByte (ACE_Wide_To_Ascii )
3. adds extra required variables like TAO_USE_IMR = 1
4. converts variables to Unicode and calls ::CreateProcess

The error occurs on the step 2. The variables occupy a continuous book of memory and are separated by '\0' char in that block. Mistakenly inherit_environment() assumes that the length of the output string of  ::WideCharToMultiByte equals to the length of the input string in characters, what is only try for ASCII strings. But if the input string contains characters which cannot be presented with 1 byte (like Chinese) these characters will be presented with 2 bytes in the output string and the length of this string in chars will be longer.
inherit_environment() moves the pointer to the next variable in the memory by adding the length of the currently converted variable so that the pointer will point not to the beginning of the next variable but further. As a result of this issue the next variable will be added incorrectly to the list and passed to ::CreateProcess. Then Windows will recognize invalid environment variable and refuse starting the process.

**Consider variables:**
PATH=c:\windows;c:\中国人民共和国;
PATHEXT=.COM,.EXE,.BAT

**They are passed to ::CreateProcess as:**
PATH=c:\windows;c:\中国人民共和国;
**_=.COM,.EXE,.BAT_**

(Note that The name PATHEXT is cut from the result list.)


---
ACE was built with ACE_HAS_WCHAR only. (ACE_USES_WCHAR was not defined)

jwillemsen:
Please extend one of the unit tests as reproducer

jwillemsen:
@coderabbitai full review

PR #2009, opened 2023-01-05
CMake Support for the ACE core

ocielliottc:


iguessthislldo (approved):

jwillemsen:
What about a github action with msvc, the action now only tests linux

ocielliottc:
> What about a github action with msvc, the action now only tests linux

The initial plan was to add Linux and add Windows later.  However, we could probably add Windows now.  I don't think it would take much.

alexchandel:
@ocielliottc Were any changes necessary for Linux to build?  When I build 7.0.11 for Linux, I get linker errors that indicate a couple system libraries (pthread & dl) are not provided.  That's something CMake might have to do.

ocielliottc:
I don't remember having any linker issues on Linux.  I did my build on an older debian installation with gcc 6.3.0.

Issue #2015, opened 2023-01-12
Unclear handling of exception passed as parameter to method


okellogg:
Tried this IDL with TAO 3.0.11:

```
interface intfmeth_exc_par {
  exception exc_t {
    string reason;
  };
  void doit (in exc_t ex);
};
```
tao_idl accepts it but the generated intfmeth_exc_parC.h shows:
```
  virtual void doit (
     ex);
```

Is it legal to pass an exception as argument of an interface method?
If not, it would be helpful if tao_idl could produce an error message.

mitza-oci:
Use of an exception type anywhere other than the `raises` clauses is not legal IDL.

jwillemsen:
This is not legal, TAOX11 does reject it (see https://github.com/RemedyIT/taox11/pull/248).

jwillemsen:
The TAO IDL_Test does have some illegal IDL we test, but not this construct

okellogg:
Here is a patch to tao_idl that works for me:

In TAO/TAO_IDL/ast/ast_{field,argument}.cpp constructor handle AST_Decl::NT_except.

```
diff --git a/TAO/TAO_IDL/ast/ast_field.cpp b/TAO/TAO_IDL/ast/ast_field.cpp
index 6b09b061604..bacbf80714c 100644
--- a/TAO/TAO_IDL/ast/ast_field.cpp
+++ b/TAO/TAO_IDL/ast/ast_field.cpp
@@ -115,6 +115,10 @@ AST_Field::AST_Field (AST_Type *ft,
           idl_global->err ()->not_a_type (ft);
         }
     }
+  else if (fnt == AST_Decl::NT_except)
+    {
+      idl_global->err ()->not_a_type (ft);
+    }
 }
 
 // To be used when constructing a node of a subclass of AST_Field.
diff --git a/TAO/TAO_IDL/ast/ast_argument.cpp b/TAO/TAO_IDL/ast/ast_argument.cpp
index 0fbe32d0656..039e5662f23 100644
--- a/TAO/TAO_IDL/ast/ast_argument.cpp
+++ b/TAO/TAO_IDL/ast/ast_argument.cpp
@@ -70,6 +70,11 @@ trademarks or registered trademarks of Sun Microsystems, Inc.
 
 #include "ast_argument.h"
 #include "ast_visitor.h"
+#include "ast_type.h"
+
+#include "utl_err.h"
+
+#include "global_extern.h"
 
 // Static functions.
 
[...]

PR #2025, opened 2023-01-14
Issue #2015 unclear handling of exception (tao_idl)

okellogg:


okellogg:
Not good yet, will try again later

PR #2030, opened 2023-01-16
Add headers and other source files to the list of files to be installed.

lockhart:
This includes stubs and skeletons for IDL interfaces and files required to extend interfaces. This is a superset of files which are known to be required to build external packages.

The current scheme for building the RPM uses dependencies from (some) built objects to deduce which files were required to be installed. But many other files are not in the core product but are required for external projects to use these components.

### DIFF
--- a/ACE/rpmbuild/ace-tao.spec
+++ b/ACE/rpmbuild/ace-tao.spec
@@ -1128,23 +1128,13 @@ cat mmraw.list |\
         sort -u > allhdrs.list
 
 # Add missing headers.
-echo ace/QtReactor/QtReactor.h >> allhdrs.list
-# pick up additional ACE files required to build external packages
 ls ace/*{.h,inl,_T.cpp} >> allhdrs.list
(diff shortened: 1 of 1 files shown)


jwillemsen:
The best solution would be to use the `make install` functionality of the GNU makefiles, the MPC files contain what should be installed. Not sure how much work that would be to try. If you want to keep the approach to just filter on extension it looks to me that some workarounds to add some specific headers can be removed as your wildcard also matches those

lockhart:
Hi Johnny. I wanted to keep (for the moment) those original workarounds which overlap the more inclusive wildcards to show what was in the original work. But I'm happy to re-submit the pull request with some cleanup of that section. I will do that momentarily.

The "make install" technique is a larger effort and I'm happy to pursue that but would like to make usable RPMs available in the meantime. This has been broken-ish since around 7.0/3.0 when, I would guess, refactoring greatly reduced the interdependencies of program elements so the current technique started leaving out chunks of functionality for installation.

jwillemsen:
Thanks!

PR #2035, opened 2023-01-17
Cleanup Solaris support

jwillemsen:


jwillemsen:
Includes removal of support for solaris threads which is a lot of code to be removed

PR #2077, opened 2023-06-16
Fix ACE_Shared_Memory_Pool

jwillemsen:
The ACE_Shared_Memory_Pool doesn't release the shared memory as it should on release. This leads to an increase of used shared memory on a system where part of the processes are restarted. This can be reproduced by using an OpenDDS test using shared memory where the subscriber keeps running and the publishers are each time restarted. After each restart of the publisher run `sudo pmap -p $(pidof subscriber) | grep shmid` and notice that the number of shmid mappings only increases.

In high level overview the following operations should be used by ACE:
- `shmget()` allocates as much physical memory as you request for the segment.
- `shmat()` programs the MMU and sets things up in the kernel so that some address range in your process maps to the segment.
- `shmdt()` does the reverse operation and removes the mapping.
- `shmctl()` with IPC_RMID deallocates the physical memory of the segment (marks it as free).

Before this PR ACE didn't use `shmdt` at all, only `shmctl` was used on release. In order to use `shmdt` we need to store the addresses returned by `shmat` and we also need to keep the shmid of the first segment in a member variable as we can't read it from the shared memory segment itself because after calling `shmdt` we can't read that memory anymore

### DIFF
--- a/ACE/ace/Shared_Memory_Pool.cpp
+++ b/ACE/ace/Shared_Memory_Pool.cpp
@@ -44,7 +44,7 @@ ACE_Shared_Memory_Pool::in_use (ACE_OFF_T &offset,
   SHM_TABLE *st = reinterpret_cast<SHM_TABLE *> (this->shm_addr_table_[0]);
   shmid_ds buf;
 
-  for (counter = 0; counter < this->max_segments_ && st[counter].used_ == true; counter++)
+  for (counter = 0; counter < this->max_segments_ && st[counter].used_ == 1; counter++)
     {
       if (ACE_OS::shmctl (st[counter].shmid_, IPC_STAT, &buf) == -1)
         ACELIB_ERROR_RETURN ((LM_ERROR,
@@ -73,7 +73,7 @@ ACE_Shared_Memory_Pool::find_seg (const void* const searchPtr,
   SHM_TABLE *st = reinterpret_cast<SHM_TABLE *> (this->shm_addr_table_[0]);
   shmid_ds buf;
 
-  for (counter = 0; counter < this->max_segments_ && st[counter].used_ == true; counter++)
+  for (counter = 0; counter < this->max_segments_ && st[counter].used_ == 1; counter++)
     {
       if (ACE_OS::shmctl (st[counter].shmid_, IPC_STAT, &buf) == -1)
         ACELIB_ERROR_RETURN ((LM_ERROR,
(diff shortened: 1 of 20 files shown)


ACE/ace/Shared_Memory_Pool.cpp
```diff
@@ -222,19 +226,19 @@ ACE_Shared_Memory_Pool::handle_signal (int, siginfo_t *siginfo, ucontext_t *)
...
-    file_perms_ (ACE_DEFAULT_FILE_PERMS),
+  : file_perms_ (ACE_DEFAULT_FILE_PERMS),
     max_segments_ (ACE_DEFAULT_MAX_SEGMENTS),
     minimum_bytes_ (0),
     segment_size_ (ACE_DEFAULT_SEGMENT_SIZE)
 {
   ACE_TRACE ("ACE_Shared_Memory_Pool::ACE_Shared_Memory_Pool");
 
-  // Only change the defaults if <options> != 0.
+  this->shm_addr_table_ = std::make_unique<void*[]>(this->max_segments_);
```
mitza-oci, line 236: The use of `shm_addr_table_[0]` and `shm_addr_table_[i]` for any `i > 0` are so different that it may be more clear just to keep `base_addr_` as it was and either have `[0]` be that same address or just keep the additional allocations in `shm_addr_table_`.

ACE/ace/Shared_Memory_Pool.h
```diff
@@ -182,14 +184,16 @@ class ACE_Export ACE_Shared_Memory_Pool : public ACE_Event_Handler
   /// Base shared memory key for the segment.
   key_t base_shm_key_;
 
+  /// Base shared memory id
+  int base_shm_id_ {};
```
mitza-oci, line 188: Is this needed as a separate variable or will it always be available in the SHM_TABLE?

jwillemsen: Gone in the last revision

ACE/ace/Shared_Memory_Pool.cpp
```diff
@@ -402,60 +400,47 @@ ACE_Shared_Memory_Pool::release (int destroy)
...
+          // Get the shared memory id and used flag on the stack as we can't read the shared memory
+          // anymore after we detached it
+          int const shmid = st[counter].shmid_;
+          int const used = st[counter].used_;
+
+          // When we have an address attached for this segment we have to detach it
+          if (this->shm_addr_table_[counter])
             {
-              result = -1;
+              if (ACE_OS::shmdt (shm_addr_table_[counter]) == -1)
```
mitza-oci, line 425:
```suggestion
              if (ACE_OS::shmdt (this->shm_addr_table_[counter]) == -1)
```

mitza-oci (changes requested):

mitza-oci (approved):

PR #2080, opened 2023-06-21
`ACE_INET_Addr::set` errantly succeeds when ACE_LACKS_GETSERVBYNAME

jrw972:
Problem
-------

Calling `set("1.2.3.4")` on an ACE_INET_Addr eventually tries to part it as a port name.  When `ACE_LACKS_GETSERVBYNAME`, this succeeds returning port number 0 with the reset address.

When port names are not expected, this breaks a useful idiom of parsing addresses that may or may not contain port numbers:

    ACE_INET_Addr addr;
    if (addr.set(x) == 0) {
      // Success, address contained a port number
    } else if (addr.set(u_short(0), x) == 0) {
      // Success, address did not contain a port number.
    } else {
      // Fail
    }

Solution
--------

Return -1 for the port number when `ACE_LACKS_GETSERVBYNAME`.

### DIFF
--- a/ACE/tests/INET_Addr_Test.cpp
+++ b/ACE/tests/INET_Addr_Test.cpp
@@ -158,6 +158,43 @@ static bool test_multiple ()
   return success;
 }
 
+static bool test_port_names()
+{
+  bool success = true;
+
+  ACE_INET_Addr addr;
+
+#if defined (ACE_LACKS_GETSERVBYNAME)
+  // Since we don't have getservbyname, the call to set should fail.
+  // Check that the call does in fact fail.
+  if (addr.set("telnet") == 0)
+    {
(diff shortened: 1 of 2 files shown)


mitza-oci (dismissed):

jwillemsen:
Can you extend a unit test for this?

ACE/tests/INET_Addr_Test.cpp
```diff
@@ -158,6 +158,42 @@ static bool test_multiple ()
   return success;
 }
 
+static bool test_port_names()
+{
+  bool success = true;
+
+
+#if defined (ACE_LACKS_GETSERVBYNAME)
+  ACE_INET_Addr addr;
```
jwillemsen, line 167: Can’t we just always run this test, so remove the define check?

iguessthislldo (approved):

PR #2081, opened 2023-06-23
Shared_Memory_Pool: fixed construction with non-default max_segments_

mitza-oci:
shm_addr_table_ needs to be initialized after max_segments_ is known
See #2077


jwillemsen:
Thanks!

PR #2083, opened 2023-06-23
Make more members of ACE_Shared_Memory_Pool const

jwillemsen:


jwillemsen:
Based on the changes of #2081

PR #1842, opened 2022-05-23
Add IDL4 map support to TAO

tmayoff:
This a WIP PR for adding maps to tao_idl. Currently I have the flex/bison and ast generation working (although a tad messy), as well as a simple test to verify it doesn't crash.

https://github.com/objectcomputing/OpenDDS/issues/3236

PR #2106, opened 2023-08-31
Backport IDL Maps to TAO2

iguessthislldo:
Backporting:
- #1329 (To make it easy easier to apply the other changes)
- #1842
- #2100
- #2103
- #2104
- #2107
- #2098
- #2136


mitza-oci (approved):
This also applies to TAO 3 - when reading the NEWS item, will it be clear to users that maps are not yet supported in the ORB?

PR #2133, opened 2023-09-21
Add missing and update .gitignore files to prevent untracked files Git status

nickwilliams-zaxiom:
When I commented `orbsvcs/tests` in the exclude block and then ran MWC and Make, `git status` revealed a host of new untracked files due to MWC and Make creating new files throughout this directory. The best solution I could come up with to keep these files out of Git was to add a bunch of new `.gitignore` files and update several existing `.gitignore` files. This will prevent this problem going forward for other developers. These are the only changes in this pull request.

If the context helps, this is the Bash command that I used to generate these changes automatically after MWC and Make:

```
$ (export IFS=$'\n'; for line in $(git status --short --untracked-files); \
    do if [[ $line == '??'* ]]; \
    then file="${line:3}"; d="$(dirname "${file}")"; f="$(basename "${file}")"; \
    echo "${f}" >> "${d}/.gitignore"; fi; done)
```


mitza-oci:
I suggest adding a cross-platform script to the repo to maintain these files going forward. See https://github.com/OpenDDS/OpenDDS/blob/master/tools/scripts/update_gitignore.pl for an example

.gitignore
```diff
@@ -50,3 +51,10 @@ ipch/
 *.*codeanalysis*
 read.lock
 fileList.bin
+
+# compiled test binaries, so that git status doesn't show untracked files
+/ACE/protocols/examples/INet/ftp_simple_wget
+/ACE/protocols/examples/INet/http_simple_wget
+/ACE/protocols/tests/HTBP/Send_Recv_Tests/client
+/ACE/protocols/tests/HTBP/Send_Recv_Tests/server
+/ACE/protocols/tests/INet/MT_Get/mt_get
```
mitza-oci, line 60: These tend to be easier to manage when they're in the local `.gitignore` for the project's directory

nickwilliams-zaxiom: Moving those now

nickwilliams-zaxiom:
> I suggest adding a cross-platform script to the repo to maintain these files going forward. See https://github.com/OpenDDS/OpenDDS/blob/master/tools/scripts/update_gitignore.pl for an example

Happy to do that, but do you have a suggestion for where?

mitza-oci:
> > I suggest adding a cross-platform script to the repo to maintain these files going forward. See https://github.com/OpenDDS/OpenDDS/blob/master/tools/scripts/update_gitignore.pl for an example
> 
> Happy to do that, but do you have a suggestion for where in the directory structure I should place this?

It could go in `ACE/bin/` but it doesn't have to.  Any suggestions @jwillemsen @iguessthislldo or others?

nickwilliams-zaxiom:
> > > I suggest adding a cross-platform script to the repo to maintain these files going forward. See https://github.com/OpenDDS/OpenDDS/blob/master/tools/scripts/update_gitignore.pl for an example
> > 
> > 
> > Happy to do that, but do you have a suggestion for where in the directory structure I should place this?
> 
> It could go in `ACE/bin/` but it doesn't have to. Any suggestions @jwillemsen @iguessthislldo or others?

I went with `ACE/bin/`, committed, and pushed.

mitza-oci (approved):

iguessthislldo (approved):

PR #2136, opened 2023-09-30
Add Note About Maps Not Usable in TAO

iguessthislldo:


jwillemsen:
This is probably confusing for users, needs some more description of what and what is not supported

iguessthislldo:
How about "Added support for the map data type, but this isn't supported for TAO user applications except for local usage"?

I don't think anything would work unless you just want to declare a map and just use it locally like with a local interface. There's no serialization, so anything with an ORB is out. Serialization and whatever else TAO needs could be added, but maps are not standardized in CORBA as far as I know.

jwillemsen:
Yes, I think that is better, as far as I followed the PR the map datatype is now parsed and the C++ type is generated, it could be used for local interfaces. CDR support, typecodes, dynamic any, etc is not working. 

For TAOX11 I have implemented CDR support, that works there, but also no typecodes, etc, the CORBA spec should be extended for the IDL4 extended types.

mitza-oci (approved):

PR #2141, opened 2023-10-05
IDL map: more TAO2 compatibility

mitza-oci:


jwillemsen:
Looks there are more problems in the backport to ace6tao2, see http://www.dre.vanderbilt.edu/scoreboard/tao2.html @mitza-oci @iguessthislldo

PR #2142, opened 2023-10-05
macOS: sizeof long double is different on ARM64

mitza-oci:
https://developer.apple.com/documentation/xcode/writing-arm64-code-for-apple-platforms#Handle-data-types-and-data-alignment-properly

### DIFF
--- a/ACE/ace/config-macosx-leopard.h
+++ b/ACE/ace/config-macosx-leopard.h
@@ -16,7 +16,7 @@
 #endif /* ! __ACE_INLINE__ */
 
 #if !defined (ACE_SIZEOF_LONG_DOUBLE)
-#  define ACE_SIZEOF_LONG_DOUBLE 16
+#  define ACE_SIZEOF_LONG_DOUBLE __SIZEOF_LONG_DOUBLE__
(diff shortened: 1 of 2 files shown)


jwillemsen (approved):
Looks ok, I don't have a mac so can't test

jwillemsen:
@mitza-oci What about enabling Mergify (see https://docs.mergify.com/commands/backport) to simplify the backporting you have to perform?

mitza-oci:
> @mitza-oci What about enabling Mergify (see https://docs.mergify.com/commands/backport) to simplify the backporting you have to perform?

We could try it.  Merging only takes a minute though, unless there are conflicts.

jwillemsen:
> > @mitza-oci What about enabling Mergify (see https://docs.mergify.com/commands/backport) to simplify the backporting you have to perform?
> 
> We could try it. Merging only takes a minute though, unless there are conflicts.

I have enabled mergify for ACE/TAO and MPC, let us see if that helps. One of the easy things to use is that you can already give the backport command while the PR is being tested, when it is merged the backport PR is automatically created.

PR #2147, opened 2023-10-23
Fixed incorrect preprocessor define check

jwillemsen:
* ACE/examples/APG/Timers/Timers.cpp:


jwillemsen:
@mergify backport ace6tao2

PR #2151, opened 2023-10-25
Fix be_visitor_map_cdr_op_ch

tmayoff:
https://github.com/DOCGroup/ACE_TAO/pull/2141#issuecomment-1778918459

### DIFF
--- a/TAO/TAO_IDL/be_include/be_visitor_map/cdr_op_ch.h
+++ b/TAO/TAO_IDL/be_include/be_visitor_map/cdr_op_ch.h
@@ -34,7 +34,7 @@ public:
   ~be_visitor_map_cdr_op_ch () = default;
 
   /// visit map
-  int visit_map (be_map *node) override;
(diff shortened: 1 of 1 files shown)


mitza-oci:
> For these errors: https://tao.microfocus.com/scoreboard/rhel7_x64_dev/2023_10_24_20_44_Full.html#error_1. Does the generated c++ need to use typedefs rather than using?

Yes, looks like that compiler doesn't have C++11 type aliases.

mitza-oci (approved):

jwillemsen:
Looks there are some more things left to cleanup

```
Error E2088 be_include\be_visitor_map/map_ch.h 33: Bad syntax for pure function definition
Error E2088 be_include\be_visitor_map/map_cs.h 35: Bad syntax for pure function definition
Error E2088 be_include\be_visitor_map/cdr_op_cs.h 32: Bad syntax for pure function definition
```

PR #2169, opened 2023-11-27
wxNamingViewer: build with wxWidgets 3.2

saper:
The patches attached are needed to build and
run wxNamingViewer with wxWidgets 3.2

For most existing code I could not identify which version
of wxWindows/wxWidgets it could have worked last time,
therefore a generic #ifdef has been used.

If no support for old frameworks is needed,
those could be removed.


mitza-oci:
Thanks for working on this.  Would it be feasible to update one of the GitHub Actions builds to enable wxWidgets?

mitza-oci:
It's running the makefile but not building anything yet.
https://github.com/DOCGroup/ACE_TAO/actions/runs/7023541947/job/19110306550?pr=2169#step:18:4354
One of the earlier steps that writes to `platform_macros.GNU` needs to add `wxWindows=1` to it

PR #2170, opened 2023-12-01
Change ACE_Barrier running_threads_ to atomic to ensure accurate values across cores/CPUs

shuston:
Customer experiences ACE_Barrier that (intermittently) never counts down to 1 under heavy system load. Theorizing that the ACE_Sub_Barrier running_threads_ member needs to be atomic to properly be viewed/manipulated across cores/CPUs.


shuston:
@jwillemsen all the Windows tasks fail... do you know if that's "normal"?

jwillemsen:
I think that is a problem with the vcpkg version used in your branch, I would recommend to merge master into your branch, I think that should resolve that compile issue

jwillemsen:
ACE_Barrier does have a guard that is locked always before using `running_threads_`

jwillemsen:
Maybe worth some test extension? Did you hear from your customer whether this change fixes their issue?

shuston:
> ACE_Barrier does have a guard that is locked always before using `running_threads_`

Yes, but we added an ACE_DEBUG inside that lock and 5 threads called wait and the running_threads_ count was 5 in every thread. The problem only occurs under tremendous load. It may take them some time to get the change into QA and run enough load to see if the problem still occurs.

PR #1888, opened 2022-07-18
Upgrade to vcpkg v10

jwillemsen:
* .github/workflows/windows.yml:

PR #2173, opened 2023-12-13
[ACE 6] GitHub Actions: allow "workflow_dispatch" and update vpckg action

mitza-oci:


jwillemsen:
See https://github.com/DOCGroup/ACE_TAO/pull/1888

Issue #2163, opened 2023-11-16
Race condition in ACE_Future double checked locking pattern code


nfrmtkr:
### Version

ACE 7.0.2

### Host machine and operating system

Build machine: CentOS Linux release 7.9.2009

### Target machine and operating system (if different from host)

Runtime: Redhat 8

### Compiler name and version (including patch level)

gcc (GCC) 9.3.1 20200408 (Red Hat 9.3.1-2)

### The $ACE_ROOT/ace/config.h file

#include "ace/config-linux.h"
#define ACE_HAS_IPV6 1
#ifndef ACE_LACKS_STROPTS_H
#define ACE_LACKS_STROPTS_H 1
#endif
#ifndef ACE_LACKS_STRRECVFD
#define ACE_LACKS_STRRECVFD 1
#endif
#undef ACE_HAS_STRBUF_T
#undef ACE_HAS_STREAMS

### The $ACE_ROOT/include/makeinclude/platform_macros.GNU file

not linked. The contents is:

INSTALL_PREFIX=/home/xgbuild/xgen-external/main/install/centos_7/ace-7.0.2
optimize=1
debug=1
ssl=1
buildbits=64
include $(ACE_ROOT)/include/makeinclude/platform_linux.GNU

### Contents of $ACE_ROOT/bin/MakeProjectCreator/config/default.features

File doesn't exist

### AREA/CLASS/EXAMPLE AFFECTED:

ACE compiles and links.
Runtime issue in customer environment.
We can't reproduce the issue localy.
Affected class ACE_Future

### The problem effects:

ACE execution

### Synopsis

We use ACE_Future<int> for handing over an int in a multi threader producer/consumer secenario.
In one of our custumer environments we see from time to time corrupted (out of scope) values in the result of a ACE_Future::get().

### Description

We've debugegd this by forcing a coredump immediatly after we checked for invalid result value of the ACE_Future::get(). The int value was -2013249520, but the future_rep_.value_ value was 0 as expected. 

We asume that in this special case the double checked locking pattern used in the set functions lacks the atomicity rule in line

ACE_NEW_RETURN (this->value_, T (r),  -1);

The stack overflow article seems to match this use case: https://stackoverflow.com/questions/50641740/double-check-locking-issues-c

Should ACE use an std::atomic for the template value? 
Is it a compiler issue, should a compiler switch be added to the make files?

The double checked locking is used at other places as well and may need to be taken into account as well.

### Repeat by

[...]

PR #2179, opened 2024-01-08
Fixed a race condition caused by the double checked locking pattern

nfrmtkr:
Fixed a race condition caused by the double checked locking pattern by using an std::atomic that ensures correct and deterministic create and assignment logic.
Issue #2163


jwillemsen:
Please also open a PR with the unit test extension as reprocuder

nfrmtkr:
The test Future_Test.cpp will test this from functional point of view. 
The issue can't be easily reproduced. It requires fast machines and it depends on the CPU and compiler being used.

nfrmtkr:
Added new test "Future_Stress_Test"

jwillemsen:
Thanks!

PR #2203, opened 2024-02-26
Reset ACE_Process_Options std handles after closed

likema:
When calling `ACE_Process_Options::release_handles` after `ACE_Process::spawn` , std handles must be reset after closed to **avoid** closing again in `ACE_Process_Options` dtor.


jwillemsen:
Thanks!

Issue #2070, opened 2023-04-13
Upgrade minimum C++ requirement to C++17


jwillemsen:
At some point we should upgrade to C++17

jwillemsen:
Feel free to open a pull request with the proposed changes @ClausKlein

ClausKlein:
> Feel free to open a pull request with the proposed changes @ClausKlein

https://github.com/DOCGroup/ACE_TAO/pull/2097

jwillemsen:
See https://github.com/DOCGroup/ACE_TAO/pull/2109

jwillemsen:
g++ 11 is the first release which has c++17 as default, for older versions we have to manually enable C++17

jwillemsen:
Done https://github.com/DOCGroup/ACE_TAO/pull/2224

PR #2212, opened 2024-04-05
Changed the default to C++17 for QNX and g++ versions older as 11 (11…

jwillemsen:
… has C++17 as default)

    * ACE/include/makeinclude/platform_g++_common.GNU:
    * ACE/include/makeinclude/platform_qnx_gcc.GNU:

### DIFF
--- a/ACE/include/makeinclude/platform_g++_common.GNU
+++ b/ACE/include/makeinclude/platform_g++_common.GNU
@@ -74,6 +74,20 @@ endif
 
 CXX_FULL_VERSION := $(shell $(CXX_FOR_VERSION_TEST) --version)
 
+# Minimum C++ level is now C++17, gcc until version 11 have an older version as default
+ifeq ($(findstring $(CXX_MAJOR_VERSION),7),$(CXX_MAJOR_VERSION))
+  c++std ?= c++17
+endif
+ifeq ($(findstring $(CXX_MAJOR_VERSION),8),$(CXX_MAJOR_VERSION))
+  c++std ?= c++17
+endif
+ifeq ($(findstring $(CXX_MAJOR_VERSION),9),$(CXX_MAJOR_VERSION))
+  c++std ?= c++17
+endif
+ifeq ($(findstring $(CXX_MAJOR_VERSION),10),$(CXX_MAJOR_VERSION))
+  c++std ?= c++17
+endif
+
 # Only modify LDFLAGS if DLD has been set.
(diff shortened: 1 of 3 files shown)


jwillemsen:
https://github.com/DOCGroup/ACE_TAO/issues/2070

ACE/include/makeinclude/platform_g++_common.GNU
```diff
@@ -74,6 +74,20 @@ endif
...
+ifeq ($(findstring $(CXX_MAJOR_VERSION),7),$(CXX_MAJOR_VERSION))
+  c++std ?= c++17
+endif
+ifeq ($(findstring $(CXX_MAJOR_VERSION),8),$(CXX_MAJOR_VERSION))
+  c++std ?= c++17
+endif
+ifeq ($(findstring $(CXX_MAJOR_VERSION),9),$(CXX_MAJOR_VERSION))
+  c++std ?= c++17
+endif
+ifeq ($(findstring $(CXX_MAJOR_VERSION),10),$(CXX_MAJOR_VERSION))
```
iguessthislldo, line 87: This looks like this would match versions 17, 18, 19, 27, etc. If this was run with those versions it would downgrade the standard being used if the user didn't set `c++std`.

ACE/include/makeinclude/platform_g++_common.GNU
```diff
@@ -74,6 +74,20 @@ endif
 
 CXX_FULL_VERSION := $(shell $(CXX_FOR_VERSION_TEST) --version)
 
+# Minimum C++ level is now C++17, gcc until version 11 have an older version as default
```
iguessthislldo, line 77: This isn't consistent with what the title of PR is. I'm also still not sure what the purpose of changing the default from the compiler default is. Does ACE still compile with C++14?

ACE/include/makeinclude/platform_g++_common.GNU
jwillemsen: This matches the code we had in the past for C++11, will have a look at this before the next micro, feel free to open a PR

ACE/include/makeinclude/platform_g++_common.GNU
jwillemsen: ACE master currently compiles with C++14, this is a first step towards C++17 as minimum required level. Someone has to dive into the microsoft support, when we have there C++17 as default I plan to move C++17 as minimum level

PR #2224, opened 2024-04-23
C++17 is the minimum required C++ level to compile ACE/TAO after this PR

jwillemsen:
Includes changes to set the C++ compiler default for Visual Studio 2017/2019/2022 to C++17

### DIFF
--- a/ACE/NEWS
+++ b/ACE/NEWS
@@ -1,4 +1,4 @@
-USER VISIBLE CHANGES BETWEEN ACE-7.1.4 and ACE-7.1.5
+SER VISIBLE CHANGES BETWEEN ACE-7.1.4 and ACE-7.2.0
 ====================================================
 
 . ACE/TAO now require C++17 or newer
--- a/ACE/NEWS
+++ b/ACE/NEWS
@@ -1,6 +1,10 @@
 USER VISIBLE CHANGES BETWEEN ACE-7.1.4 and ACE-7.1.5
 ====================================================
 
+. ACE/TAO now require C++17 or newer
+
+. Add support for Embarcadero C++ Builder bcc64x compiler
+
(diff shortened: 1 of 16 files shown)


ACE/NEWS
```diff
@@ -1,6 +1,10 @@
 USER VISIBLE CHANGES BETWEEN ACE-7.1.4 and ACE-7.1.5
```
iguessthislldo, line 1: Please make this a major release if you're requiring a newer C++ standard.

ACE/include/makeinclude/platform_linux_clang.GNU
```diff
@@ -35,6 +41,18 @@ ifeq ($(optimize),0)
...
+  CCFLAGS += -std=c++20
+endif
+
+ifeq ($(c++17),1)
+  CCFLAGS += -std=c++17
+endif
+
 ifeq ($(c++14),1)
   CCFLAGS += -std=c++14
 endif
```
iguessthislldo, line 58: It seems like these should be part of the same if block.

```suggestion
ifneq ($(c++std),)
  CCFLAGS += -std=$(c++std)
else ifeq ($(c++20),1)
  CCFLAGS += -std=c++20
else ifeq ($(c++17),1)
  CCFLAGS += -std=c++17
else ifeq ($(c++14),1)
  CCFLAGS += -std=c++14
endif
```

iguessthislldo (changes requested):

ACE/NEWS
jwillemsen: I will make it a new micro, we are not deprecating specific compilers, just lifting the required C++ standard

iguessthislldo: Stop resolving this. I still don't agree.

Issue #958, opened 2019-09-10
ACE SSL uses deprecated OpenSSL functions


gadh:
SSLv23_method(), SSLv23_server_method(), SSLv23_client_method()
Use of these functions is deprecated. They have been replaced with the above TLS_method(), TLS_server_method() and TLS_client_method() respectively. New code should use those functions instead.

the developer should use the TLS methods instead

jwillemsen:
Please create a pull request with the proposed changes

PR #2231, opened 2024-04-30
Implemented OpenSSL 1.1.0 TLS methods and deprecated SSLv23 ones.

akotulu:
SSLv23_method(), SSLv23_server_method() and SSLv23_client_method() were deprecated and the preferred TLS_method(), TLS_server_method() and TLS_client_method() functions were added in OpenSSL 1.1.0.

All version-specific methods were deprecated in OpenSSL 1.1.0.

[SSLv23_method](https://www.openssl.org/docs/man3.0/man3/SSLv23_method.html)

### DIFF
--- a/ACE/ace/SSL/SSL_Context.cpp
+++ b/ACE/ace/SSL/SSL_Context.cpp
@@ -261,6 +261,23 @@ ACE_SSL_Context::set_mode (int mode)
   SSL_METHOD *method = 0;
 #endif
 
+#if OPENSSL_VERSION_NUMBER >= 0x10100000L
+  switch (mode)
+    {
+    case ACE_SSL_Context::TLS_client:
+      method = ::TLS_server_method ();
+      break;
+    case ACE_SSL_Context::TLS_server:
+      method = ::TLS_client_method ();
+      break;
+    case ACE_SSL_Context::TLS:
+      method = ::TLS_method ();
+      break;
+    default:
+      method = ::TLS_method ();
+      break;
+    }
+#else
   switch (mode)
     {
     case ACE_SSL_Context::SSLv23_client:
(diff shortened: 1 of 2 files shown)


jwillemsen:
Doesn't this change break any user code which uses the SSLv23 ones?

akotulu:
Here is quote from docs.

```
SSLv23_method(), SSLv23_server_method(), SSLv23_client_method()
These functions do not exist anymore, they have been renamed to TLS_method(), TLS_server_method() and TLS_client_method() respectively. Currently, the old function calls are renamed to the corresponding new ones by preprocessor macros, to ensure that existing code which uses the old function names still compiles. However, using the old function names is deprecated and new code should call the new functions instead.
```

jwillemsen:
Ok, but you do change the enum provided by ACE in such a way that when someone uses it, it will not compile unless they change their code

jwillemsen:
https://github.com/DOCGroup/ACE_TAO/issues/958

akotulu:
Ok, I will add the enum values back with corresponding TLS calls.

jwillemsen:
Mark the old enums as deprecated so that we know they are deprecated and can be removed at some point, maybe use the C++14 https://en.cppreference.com/w/cpp/language/attributes/deprecated

akotulu:
Made the changes. Is it ok now?

jwillemsen:
What when ssl version is smaller as 0x10100000L, than the new defines are there. Also in `ACE_SSL_Context::load_trusted_ca` the old enums are not tested

akotulu:
Huh, sorry about the mess. Haven't done such a backwards compatibility stuff before.
It should now be ok.

jwillemsen:
When < 0x10100000L the enums will still give a deprecated warning

akotulu:
Idk if this is the best way to remove the deprecation warning, but here it is.

jwillemsen:
Please fix fuzz errors

akotulu:
Done.

Issue #2286, opened 2024-10-02
/usr/include/ace/SSL/SSL_Context.h not present in 8.0.1 ace-devel packages (it is present in 7.1.4 when built on the same machine)


undergraver:
### Version

8.0.1

### Host machine and operating system
OpenSUSE 15.6 64bit (same target)

### Compiler name and version (including patch level)

# gcc --version
gcc (SUSE Linux) 7.5.0
Copyright (C) 2017 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

iulians@dock-v11r1.4-build:~ # g++ --version
g++ (SUSE Linux) 7.5.0
Copyright (C) 2017 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. 
#

### The $ACE_ROOT/ace/config.h file
Using the default one from tar.gz.

I am building using the .tar.gz file via: "rpmbuild -ta ACE+TAO-src-8.0.1.tar.gz" (tar.gz is taken from github releases and renamed with src in its name)

### The $ACE_ROOT/include/makeinclude/platform_macros.GNU file
Using the default one from tar.gz. See above how I build.

### Contents of $ACE_ROOT/bin/MakeProjectCreator/config/default.features
Using the default one from tar.gz. See above.

### AREA/CLASS/EXAMPLE AFFECTED:

ace-devel doesn't contain SSL_Context.h header file. 

# rpm -qlp /usr/src/packages/RPMS/x86_64/ace-devel-8.0.1-1.x86_64.rpm | grep SSL
/usr/include/ace/SSL
/usr/include/ace/SSL/SSL_Asynch_BIO.h
/usr/include/ace/SSL/SSL_Export.h
/usr/include/ace/SSL/sslconf.h
/usr/lib64/libACE_INet_SSL.so
/usr/lib64/libACE_SSL.so
#

compiling on the same machine latest 7 verson (7.1.4) results in a proper ace-devel package like it can be seen below:

# rpm -qlp /home/iulians/rpmbuild/RPMS/x86_64/ace-devel-7.1.4-1.x86_64.rpm | grep SSL
/usr/include/ace/SSL
/usr/include/ace/SSL/SSL_Asynch_BIO.h
[...]

PR #2288, opened 2024-10-06
fix for #2286 - avoid header compilation failure in %install phase

undergraver:
Adding -std=c++17 to avoid header compilation failure which removes them from the list of headers.
Fixing #2286 - tested on openSUSE15.6 with gcc 7.5 compiler which is not c++17 or newer by default.

The explanation for working with g++ 14.2 (on openSUSE Tumbleweed) is that the compiler is by default c++17 while g++ 7.5 is c++14 even if the compiler is capable of handling c++17 specs - which is actually used in compilation.


jwillemsen:
Thanks!

PR #2301, opened 2024-11-19
Remove duplicate reactor timer management functions

sonndinh:
- Problem: `ACE_Reactor` implements functions `schedule_timer` and `reset_timer_interval` which it already inherits from `ACE_Reactor_Timer_Interface`. The functions are identical in these 2 classes.

- Solution: Remove the functions in `ACE_Reactor` and keep them in the base `ACE_Reactor_Timer_Interface` class.


sonndinh:
@mitza-oci Using `using` is better - it's more concise and saves a call to the base class's member function.

PR #2306, opened 2024-12-11
Fixed some warnings

mitza-oci:
These warnings impact downstream projects since they're in headers and generated code.

PR #2309, opened 2024-12-17
Use =default for skeleton copy constructor

mitza-oci:
Fixes additional warnings introduced in #2306

### DIFF
--- a/TAO/TAO_IDL/be/be_visitor_interface/interface_sh.cpp
+++ b/TAO/TAO_IDL/be/be_visitor_interface/interface_sh.cpp
@@ -115,7 +115,7 @@ be_visitor_interface_sh::visit_interface (be_interface *node)
 
   // Copy constructor and destructor.
   *os << class_name.c_str () << " (const "
-      << class_name.c_str () << "& rhs);" << be_nl
+      << class_name.c_str () << "& rhs) = default;" << be_nl
       << "virtual ~" << class_name.c_str () << " () = default;" << be_nl_2;
 
   // _is_a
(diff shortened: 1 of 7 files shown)


TAO/TAO_IDL/be/be_visitor_interface/interface_sh.cpp
```diff
@@ -115,7 +115,7 @@ be_visitor_interface_sh::visit_interface (be_interface *node)
 
   // Copy constructor and destructor.
   *os << class_name.c_str () << " (const "
-      << class_name.c_str () << "& rhs);" << be_nl
+      << class_name.c_str () << "& rhs) = default;" << be_nl
```
jwillemsen, line 118: `rhs` should be removed, will trigger new warnings with some compilers

TAO/TAO_IDL/be/be_visitor_interface/interface_ss.cpp
```diff
@@ -456,14 +428,6 @@ be_visitor_interface_ss::generate_proxy_classes (be_interface *node)
   return 0;
 }
 
-int
-be_visitor_interface_ss::generate_copy_ctor (be_interface *node,
-                                             TAO_OutStream *os)
-{
-  return node->traverse_inheritance_graph (be_interface::copy_ctor_helper,
```
jwillemsen, line 463: Can't we zap copy_ctor_helper now also?

jwillemsen:
There is also a `-GIc` to enable generation of a copy constructor in the implementation file, the one we generate there looks not correct given your initial changes

TAO/TAO_IDL/be/be_visitor_interface/interface_sh.cpp
mitza-oci: Even for a declaration with no function body?

jwillemsen: Yes, something I did encounter when working on TAOX11, not sure which compilers, very likely one of the older gcc compilers. You could leave it and the scoreboard will tell whether we still support one of those broken ones, if so, it can be removed easily at that moment.

PR #2339, opened 2025-02-26
Fixed warnings and updated bison-generated code for ACE svc.conf

mitza-oci:
Also updated GitHub Actions CI since Ubuntu-20.04 is going away.


iguessthislldo (approved):

ACE/ace/CDR_Base.h
```diff
@@ -335,10 +335,11 @@ class ACE_Export ACE_CDR
          {
            MAX_DIGITS = 31,
            MAX_STRING_SIZE = 4 + MAX_DIGITS, // includes -, 0, ., terminator
-           POSITIVE = 0xc,
-           NEGATIVE = 0xd
          };
 
+         static const Octet POSITIVE = 0xc;
```
jwillemsen, line 340: Why only mode these out of the enum? constexpr instead of const?

ACE/ace/OS_NS_string.cpp
```diff
@@ -101,18 +101,19 @@ ACE_OS::strerror (int errnum)
...
   errmsg = ret_errortext;
   ACE_SECURECRTCALL (strerror_s (ret_errortext, sizeof(ret_errortext), errnum),
                      char *, 0, errmsg);
   if (errnum < 0 || errnum >= _sys_nerr)
     g = EINVAL;
 
   return errmsg;
-#elif defined (ACE_WIN32)
+# else
+#  if defined (ACE_WIN32)
```
jwillemsen, line 113: Why is this changed?

ACE/ace/Asynch_IO_Impl.inl
```diff
@@ -13,90 +13,72 @@ ACE_Asynch_Operation_Impl::ACE_Asynch_Operation_Impl ()
 
 ACE_INLINE
 ACE_Asynch_Read_Stream_Impl::ACE_Asynch_Read_Stream_Impl ()
-  : ACE_Asynch_Operation_Impl ()
```
jwillemsen, line 16: Just declare them as default in the header so that the implementation of all the default constructos can all be removed

ACE/ace/OS_NS_string.cpp
mitza-oci: The previous version results in dead code, which causes a warning from the compiler (based on warning level).

ACE/ace/CDR_Base.h
mitza-oci: They're used differently from the other constants because they are part of expressions that are assigned to values in the byte array representation of `Fixed`.  Having them start with type `Octet` is easier than needing to convert each time they're used.

The compiler treats `static const` members of classes which have integer types and are assigned values in the class definition the same as constexpr, and then the code can exactly match across the different ACE versions.

iguessthislldo (approved):

ACE/ace/Asynch_IO_Impl.h
```diff
@@ -46,7 +46,7 @@ class ACE_Proactor_Impl;
 class ACE_Export ACE_Asynch_Result_Impl
```
jwillemsen, line 46: Do we still need to export these classes as there is nothing to export anymore?

jwillemsen (dismissed):
Just the comment about export

ACE/ace/CDR_Base.cpp
```diff
@@ -210,10 +210,10 @@ ACE_CDR::swap_2_array (char const * orig, char* target, size_t n)
           a = (a1 | a2);
           b = (b1 | b2);
 
-          ACE_UINT32 c1 = static_cast<ACE_UINT16> (a >> 16);
-          ACE_UINT32 c2 = static_cast<ACE_UINT16> (a & 0xffff);
-          ACE_UINT32 c3 = static_cast<ACE_UINT16> (b >> 16);
-          ACE_UINT32 c4 = static_cast<ACE_UINT16> (b & 0xffff);
+          ACE_UINT16 c1 = static_cast<ACE_UINT16> (a >> 16);
```
jwillemsen, line 213: const?

ACE/ace/CDR_Base.cpp
```diff
@@ -1395,8 +1395,8 @@ ACE_CDR::Fixed ACE_CDR::Fixed::div_helper2 (const Fixed &rhs, Fixed &r) const
         top.digit (i, 0);
       for (int i = dig; i < this->digits_; ++i)
         bot.digit (i, 0);
-      bot.digits_ = dig;
-      top.scale_ += dig;
+      bot.digits_ = static_cast<Octet> (dig);
+      top.scale_ += bot.digits_;
```
jwillemsen, line 1399: Test necessary for the change to bot.digits_?

mitza-oci: It's not really changing - the value `dig` is an int that we know is constrained to small positive values.  This adds that cast to silence a warning.

.github/workflows/linux.yml
```diff
@@ -24,146 +24,101 @@ jobs:
...
+            os: ubuntu-24.04
             optional_feature: uses_wchar=1
             optional_macros: inline=0
           - feature: CodeQL
             CC: gcc-10
             CXX: g++-10
             PackageDeps: g++-10
             platform_file: include $(ACE_ROOT)/include/makeinclude/platform_linux.GNU
-            os: ubuntu-20.04
+            os: ubuntu-24.04
```
mitza-oci, line 121: @jwillemsen looks like the CodeQL configuration in GitHub needs to be changed so it can work with 24.04 instead of 20.04

jwillemsen: As far as I can find there is nothing to configure on github

jwillemsen:
The clang based Embarcadero C++ Builder don't link the default, see this link error in some internal builds

```
Error: Unresolved external 'ACE_Asynch_Connect_Impl::~ACE_Asynch_Connect_Impl()' referenced from X:\BUILD\DOC_ROOT\ACE\ACE\DEBUG\ACE\PROACTOR.O
```

jwillemsen:
Seems that is only the old clang based compilers, the newer don't have a problem, so no need to fix this right now @mitza-oci.

jwillemsen:
New warnings on RHEL80

```
[[Details](http://logs.remedy.nl/rhel80_acetao_release/2025_03_01_23_28_Full.html#warning_72)] /home/build/jenkins/workspace/rhel80_acetao_release/BUILD/DOC_ROOT/TAO/tao/Valuetype/ValueBase.cpp: In copy constructor ‘CORBA::DefaultValueRefCountBase::DefaultValueRefCountBase(const CORBA::DefaultValueRefCountBase&)’:
[[Details](http://logs.remedy.nl/rhel80_acetao_release/2025_03_01_23_28_Full.html#warning_73)] /home/build/jenkins/workspace/rhel80_acetao_release/BUILD/DOC_ROOT/TAO/tao/Valuetype/ValueBase.cpp:1401:1: warning: base class ‘class CORBA::ValueBase’ should be explicitly initialized in the copy constructor [-Wextra]
/home/build/jenkins/workspace/rhel80_acetao_release/BUILD/DOC_ROOT/TAO/tao/PortableServer/GNUmakefile.PortableServer
[[Details](http://logs.remedy.nl/rhel80_acetao_release/2025_03_01_23_28_Full.html#warning_74)] /home/build/jenkins/workspace/rhel80_acetao_release/BUILD/DOC_ROOT/TAO/tao/PortableServer/Servant_Base.cpp: In copy constructor ‘TAO_ServantBase::TAO_ServantBase(const TAO_ServantBase&)’:
[[Details](http://logs.remedy.nl/rhel80_acetao_release/2025_03_01_23_28_Full.html#warning_75)] /home/build/jenkins/workspace/rhel80_acetao_release/BUILD/DOC_ROOT/TAO/tao/PortableServer/Servant_Base.cpp:64:1: warning: base class ‘class TAO_Abstract_ServantBase’ should be explicitly initialized in the copy constructor [-Wextra]
```

jwillemsen:
Also visible at https://storage.googleapis.com/oci-scoreboard/ACE_TAO_Ubuntu_22_04_GCC14_x64_Debug/2025_03_01_06_45_Brief.html#section_5, stragely when going to the build at https://www.dre.vanderbilt.edu/scoreboard/integrated.html I get an index page of an old build

mitza-oci:
The warning from compiler on RHEL 8.0 is incorrect / counterproductive, but not sure what we want to do about it -- maybe a pragma?  Here's why:

* ValueBase is a virtual base class of DefaultValueRefCountBase
* DefaultValueRefCountBase is an abstract class because `TypeCode_ptr _tao_type () const` (among others) is pure virtual
* DefaultValueRefCountBase's constructor will never initialize ValueBase because only the most-derived constructor can initialize a virtual base
* Therefore anything we write in the initializer is effectively dead code, and this creates a warning from Visual C++

jwillemsen:
Newer gcc versions show also the same warning

jwillemsen:
gcc14 does show the same warning, see https://storage.googleapis.com/oci-scoreboard/ACE_TAO_Ubuntu_22_04_GCC14_x64_Debug/index.html. 

Maybe easier to disable the call of the base constructor with msvc?

.github/workflows/linux.yml
mitza-oci: OK, I guess it resolved itself, but when I was working on this I saw an error on https://github.com/DOCGroup/ACE_TAO/security/code-scanning/tools/CodeQL/status

PR #917, opened 2019-06-06
Adding support for the MQX RTOS.

ocielliottc:
- Use ACE_OS::send and ACE_OS::recv on MQX, instead of write and read.
- Added static casts where pid_t is assumed to be an int.
- Use configurable macro for the "use_select" parameter.
- MQX does not follow conventional standards when it comes to HTONS and NTOHS and must be configured accordingly.
- Modified ACE_Handle_Set so that the use of fd_array is not conditional upon ACE_WIN32.  It is now enabled through ACE_HANDLE_SET_USES_FD_ARRAY.
- Implement UNIX File IO Functions for MQX in MQX_Filesystem.*
- Use the existing implementation to simulate condition variables.
- Use emulation for diropen on MQX as is done for Windows.
- Convert from MQX error to standard error number.
- Use _mqx_exit in ACE_OS::_exit for MQX.
- Added a nop in ACE_OS::ioctl for MQX.
- Added a definition of ACE_TMAIN for MQX so that MQX is initialized and RTCS and the MQX Filesystem code is initialized before calling ace_main_i().
- Fixed a compile error in SOCK_Dgram.cpp introduced months ago.
- Use struct initialization to avoid using the l_linger member directly.
- Modified ACE_Select_Reactor_Base to use a hash map for handle storage based on ACE_SELECT_REACTOR_BASE_USES_HASH_MAP instead of ACE_WIN32.
- Added the IAR compiler to define ACE_NOTREACHED(a) as empty.
- Made ACE_IPPROTO_TCP conditionally defined.
- Added error values required for ACE.
- Skip the wide-character strtok test if ACE_LACKS_WCSTOK is defined in ACE/tests/OS_Test.cpp.
- Added the ability to compile tests so that they output to stderr instead of a file.

PR #2340, opened 2025-02-28
Updated support for Green Hills INTEGRITY and INTEGRITY-178 RTOS

sonndinh:
## Summary by CodeRabbit

- **New Features**
  - Introduced support for a single-input reactor mode that extends event handling options.
  - Added enhanced support for Green Hills INTEGRITY and INTEGRITY-178 tuMP RTOS, tested on versions 11.4.6 and 5.0.0 respectively.

- **Refactor / Maintenance**
  - Improved compatibility with INTEGRITY-based systems through streamlined error handling, signal processing, threading, and memory management adjustments.
  - Refined conditional configurations, preprocessor directives, and code comments for better clarity and maintainability.
  - Enhanced handling of string views for more efficient string operations.
  - Updated network utility and socket function support for INTEGRITY platforms.

- **Chores**
  - Updated repository ignore settings to exclude additional build artifacts and project-specific files.
  - Added new source files and updated project configurations to support the new single-input reactor functionality.
  - Removed deprecated compiler support for Embarcadero C++ Builder bcc32c and bcc64.


.gitignore
```diff
@@ -17,6 +17,11 @@
 *S_T.inl
 *.bmak
 *.recipe
+*.gpj
+*.int
+*.cmd
```
iguessthislldo, line 22: `cmd` is a common extension for windows batch files. There is [one in the tree](https://github.com/DOCGroup/ACE_TAO/blob/master/TAO/orbsvcs/tests/FT_App/replica.cmd), though I don't think it's important. I'm not sure what we can do about this if integrity is using it though.

TAO/tao/IIOP_Acceptor.cpp
```diff
@@ -1,3 +1,4 @@
+
```
iguessthislldo, line 1:
```suggestion
```

ACE/tests/Dynamic_Priority_Test.cpp
```diff
@@ -47,7 +47,7 @@
 #include "ace/High_Res_Timer.h"
 #include "ace/Sched_Params.h"
 
-#if defined (VXWORKS)
+#if defined (VXWORKS) || defined (INTEGRITY)
 # include "ace/OS_NS_string.h"
 #endif /* VXWORKS */
```
iguessthislldo, line 52:
```suggestion
#endif
```

ACE/ace/ACE.inl
```diff
@@ -216,7 +216,7 @@ ACE::sendv_n (ACE_HANDLE handle,
 ACE_INLINE ssize_t
 ACE::send_i (ACE_HANDLE handle, const void *buf, size_t len)
 {
-#if defined (ACE_WIN32) || defined (ACE_MQX)
+#if defined (ACE_WIN32) || defined (ACE_MQX) || defined (__ghs__)
```
iguessthislldo, line 219: Following convention, shouldn't macros like `__ghs__` and `INTEGRITY` have their own `ACE_*` counterparts?

.gitignore
mitza-oci: I don't think we need in this PR.  In general, we could remove all tool-specific files from .gitignore and let each user configure their own .git/info/exclude or similar.

ACE/ace/CDR_Stream.cpp
```diff
@@ -1707,14 +1707,18 @@ ACE_InputCDR::read_string (std::string& x)
   // the memory is allocated.
   if (len > 0 && len <= this->length())
     {
+#if !defined (ACE_NO_GHS_EXCEPT)
```
iguessthislldo, line 1710: Assuming this macro is new, maybe it could have a better name? Like `ACE_LACKS_EXCEPTIONS` or `ACE_LACKS_STD_EXCEPTIONS`? I don't know what the platform supports, so I don't know what I'm the scope of this is.

mitza-oci:
Since #2339 was recently merged, you may want an explicit merge from the master branch to pick up the updates to the CI configuration

iguessthislldo (changes requested):

ACE/ace/CDR_Stream.cpp
jwillemsen: How does integrity handle a resize error? Shouldn't be there some alternative code instead to handle the error cases?

ACE/ace/OS_NS_Thread.cpp
```diff
@@ -3259,6 +3297,149 @@ ACE_OS::set_scheduling_params (const ACE_Sched_Params &/*sched_params*/, ACE_id_
...
+
+class ACE_Int178_Stack_Manager
+{
+private:
+  typedef Address Int178_Stack[ACE_INT178_DEFAULT_STACK_SIZE / sizeof (Address)];
+  typedef Int178_Stack Int178_Stack_Pool[NUM_DYNAMIC_THREADS];
+
+  struct StackInfo
+  {
+    StackInfo () : id(0), occupied(false) {}
```
jwillemsen, line 3342: You could use uniform initialization to simplify this

ACE/ace/OS_NS_Thread.h
```diff
@@ -224,7 +230,141 @@ ACE_END_VERSIONED_NAMESPACE_DECL
...
+      ACE_thread_mutex_t &lock_;
+    };
+
+#     if defined (ACE_HAS_TSS_EMULATION) && !defined (ACE_HAS_THREAD_SPECIFIC_STORAGE)
+    class ACE_INTEGRITY_TSS_Impl
+    {
+    private:
+      struct TaskInfo
+      {
+        TaskInfo () : id(0), tss_base(0), valid(false) {}
```
jwillemsen, line 296: Uniform initialization would simplify this

ACE/ace/OS_NS_dirent.inl
```diff
@@ -20,8 +20,10 @@ closedir (ACE_DIR *d)
   delete d;
 # elif defined (ACE_HAS_WCLOSEDIR_EQUIVALENT) && defined (ACE_USES_WCHAR)
   ACE_HAS_WCLOSEDIR_EQUIVALENT (d);
-# else /* ACE_WIN32 && ACE_LACKS_CLOSEDIR */
+# elif !defined (ACE_LACKS_CLOSEDIR) /* ACE_WIN32 && ACE_LACKS_CLOSEDIR */
```
jwillemsen, line 23: Find it more readable when you first would check `defined (ACE_LACKS_CLOSEDIR)`, and the normal case in the else part

ACE/ace/config-integritySCA.h
```diff
@@ -84,65 +116,54 @@
...
 #define ACE_LACKS_SEARCH_H
 #define ACE_LACKS_SYS_IPC_H
 #define ACE_LACKS_SYS_SEM_H
-#define ACE_LACKS_PWD_H
 #define ACE_LACKS_STROPTS_H
-#define ACE_LACKS_DLFCN_H
-#define ACE_LACKS_REGEX_H
-#define ACE_LACKS_POLL_H
 #define ACE_LACKS_SYS_SHM_H
-#define ACE_LACKS_TERMIOS_H
```
jwillemsen, line 145: Are there maybe any other parts that could be cleaned because some lacks macros are not used anymore (integrity maybe the last to need a certain workaround)?

ACE/tests/Task_Test.cpp
```diff
@@ -122,7 +122,7 @@ run_main (int, ACE_TCHAR *[])
   // Set the thread hook!
   ACE_Thread_Hook::thread_hook (new My_Thread_Hook);
 
-  int n_threads = ACE_MAX_THREADS;
+  int n_threads = 3; //ACE_MAX_THREADS;
```
jwillemsen, line 125: Something to revert?

TAO/tao/default_resource.cpp
```diff
@@ -791,12 +791,14 @@ TAO_Default_Resource_Factory::get_reactor ()
   else
     this->dynamically_allocated_reactor_ = true;
 
+  ACE_Reactor::instance (reactor);
```
jwillemsen, line 794: Doesn't look a good thing to just replace the singleton reactor with the one created by TAO, something to revert, also reclaim

ACE/ace/config-integritySCA.h
```diff
@@ -171,9 +187,12 @@ extern "C"
 #define ACE_LACKS_READLINK
 #define ACE_LACKS_GETPPID
 #define NSIG (SIGRTMAX+1)
+#define ACE_USES_STD_NAMESPACE_FOR_STDCPP_LIB 1
+#define ACE_USE_RCSID 0
```
jwillemsen, line 191: ACE_USE_RCSID is not used anymore

ACE/ace/CDR_Stream.cpp
jwillemsen: ACE/TAO master requires C++ exception support, support for emulated exceptions was removed years ago

ACE/ace/OS_NS_dirent.inl
sonndinh: There is a check for `ACE_LACKS_CLOSEDIR` prior to this. I agree with that in general; in this case, however, there isn't such clear separation.

ACE/ace/CDR_Stream.cpp
sonndinh: One of the design principles of ACE is avoiding exception handling: https://github.com/DOCGroup/ACE_TAO/blob/master/ACE/docs/ACE-porting.html#L158-L163. Requiring C++ exceptions in ACE seems to go against it. In some real-time systems which use ACE, it's desirable to turn off exception support to improve determinism.

jwillemsen: That documentation is outdated as far as I know, C++ exception support is I think mandatory for years now, will dig tomorrow briefly in the ACE history to see from when

jwillemsen: Support for non exception builds was removed in 2010, looks the docu need an update https://github.com/DOCGroup/ACE_TAO/blob/23b4b06cc754916a678f207e234502ffbe35ee2d/ACE/ChangeLogs/ChangeLog-2011a#L5371

jwillemsen:
I have updated the porting documentation, see https://github.com/DOCGroup/ACE_TAO/pull/2344

ACE/ace/CDR_Stream.cpp
mitza-oci: When will this code even be used?  TAO doesn't use `std::string` in the IDL mapping.
Assuming it's not needed, we can just exclude it from the embedded system builds.

jwillemsen: It is used by the alternate STL mapping support in TAO_IDL as far as I know and it is used by TAOX11/AXCIOMA in various places

TAO/tao/default_resource.cpp
jwillemsen: Maybe TAO_DEFAULT_REACTOR_TYPE is a helpful define to change the default reactor type for TAO?

ACE/ace/OS_NS_Thread.inl
```diff
@@ -1434,7 +1442,12 @@ ACE_OS::sema_destroy (ACE_sema_t *s)
   ACE_OSCALL (::semDelete (s->sema_), int, result);
   s->sema_ = 0;
   return result;
-# endif /* ACE_HAS_PTHREADS */
+# elif defined (INTEGRITY)
+  // INTEGRITY has CloseSemaphore kernel call that closes the semaphore object.
+  // However, we are imitating INTEGRITY-178 behavior which does not support this operation.
```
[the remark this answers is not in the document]
sonndinh: The original support for INTEGRITY API was for debugging our support of INTEGRITY-178 API, so I made it as close as possible to the changes made for INTEGRITY-178. INTEGRITY API, in general, provides more functionalities and the current support can be enhanced. For example, Task on INTEGRITY can be created with an argument, so it maybe able to avoid the workaround for wrapping the Task's entry point argument.

I can add a preprocessor for this specific call.

ACE/ace/CDR_Stream.cpp
mitza-oci: We will exclude it on this platform/configuration, which doesn't support that std library function.

mitza-oci: Actually I got that wrong -- the std library function is supported but error handling is different (std::allocator behaves differently).  @sonndinh will update the code.

ACE/ace/config-integritySCA.h
sonndinh: I checked and no further code needs to be cleaned up.

ACE/ace/CDR_Stream.cpp
```diff
@@ -1780,14 +1787,18 @@ ACE_InputCDR::read_wstring (std::wstring& x)
         }
       else
         {
-          try
+#if !defined (ACE_STD_ALLOCATOR_NOTHROW)
+        try
```
mitza-oci, line 1791:
```suggestion
          try
```

ACE/ace/OS_NS_Thread.h
```diff
@@ -40,6 +40,12 @@
 # endif /* VxWorks and ! SMP */
 #endif
 
+# if defined (INTEGRITY)
+#   include "ace/Log_Msg.h"
+
+#   include <map>
```
jwillemsen, line 46: map include necessary?

ACE/ace/OS_NS_Thread.h
```diff
@@ -224,7 +230,149 @@ ACE_END_VERSIONED_NAMESPACE_DECL
...
+    private:
+      struct TaskInfo
+      {
+        TaskInfo ()
+          : id {}
+          , tss_base {}
+          , valid {}
+        {}
+
+        ACE_hthread_t id;
```
jwillemsen, line 306: Just use uniform init

``` suggestion
        ACE_hthread_t id {};
```

ACE/ace/config-integrity-common.h
```diff
@@ -0,0 +1,185 @@
+#ifndef ACE_CONFIG_INTEGRITY_COMMON_H
+#define ACE_CONFIG_INTEGRITY_COMMON_H
+
+#define ghs
```
jwillemsen, line 4: Why use ghs, not a ACE style define

ACE/ace/OS_NS_Thread.cpp
```diff
@@ -1151,11 +1151,11 @@ ACE_OS::cond_broadcast (ACE_cond_t *cv)
         result = -1;
       // Wait for all the awakened threads to acquire their part of
       // the counting semaphore.
-#   if defined (ACE_VXWORKS) || defined (ACE_MQX)
+#   if defined (ACE_VXWORKS) || defined (ACE_MQX) || defined (INTEGRITY)
```
jwillemsen, line 1154: What about adding ACE_INTEGRITY instead of INTEGRITY so that the define style is consistent

ACE/ace/OS_NS_Thread.cpp
```diff
@@ -3259,6 +3297,152 @@ ACE_OS::set_scheduling_params (const ACE_Sched_Params &/*sched_params*/, ACE_id_
...
+  typedef Int178_Stack Int178_Stack_Pool[NUM_DYNAMIC_THREADS];
+
+  struct StackInfo
+  {
+    StackInfo ()
+      : id {}
+      , occupied {}
+    {}
+
+    ACE_hthread_t id;
```
jwillemsen, line 3347: Uniform init

ACE/ace/OS_NS_Thread.inl
```diff
@@ -2496,6 +2553,13 @@ ACE_OS::thr_getprio (ACE_hthread_t ht_id, int &priority, int &policy)
   return 0;
 # elif defined (ACE_HAS_VXTHREADS)
   return ::taskPriorityGet (ht_id, &priority);
+# elif defined (INTEGRITY)
+  Value active_prio;
+  Error err = ::GetActivePriority (ht_id, &active_prio);
```
jwillemsen, line 2558: const?

ACE/ace/OS_NS_arpa_inet.inl
```diff
@@ -9,7 +9,40 @@ ACE_INLINE unsigned long
 ACE_OS::inet_addr (const char *name)
 {
   ACE_OS_TRACE ("ACE_OS::inet_addr");
-#if defined (ACE_LACKS_INET_ADDR)
+#if defined (INTEGRITY) && defined (ACE_LACKS_INET_ADDR)
+
+  u_long ret = 0;
+  u_int segment;
+  u_int valid = 1;
```
jwillemsen, line 16: valid as bool?

ACE/ace/Service_Config.cpp
```diff
@@ -299,13 +299,13 @@ ACE_Service_Config::open_i (const ACE_TCHAR program_name[],
   // user forgets to define an object of type ACE_Service_Config).
   ACE_Service_Repository::instance (ACE_Service_Gestalt::MAX_SERVICES);
 
+#if !defined (ACE_LACKS_UNIX_SIGNALS)
```
jwillemsen, line 302: Retrieving reactor is not dependent on unix signals, concerned that this break things

ACE/ace/config-integrity-common.h
```diff
@@ -0,0 +1,185 @@
...
+#define ACE_LACKS_SETSID
+#define ACE_LACKS_SETEGID
+#define ACE_LACKS_SETUID
+#define ACE_LACKS_SETEUID
+#define ACE_LACKS_GETEUID
+#define ACE_LACKS_GETUID
+#define ACE_LACKS_GETEGID
+#define ACE_LACKS_GETGID
+
+#define ACE_HAS_STANDARD_CPP_LIBRARY 1
```
jwillemsen, line 19: Looks you are setting some old defines which we zapped some time ago

ACE/ace/config-integrity-common.h
```diff
@@ -0,0 +1,185 @@
...
+#define ACE_HAS_TYPENAME_KEYWORD
+#define ACE_NEEDS_FUNC_DEFINITIONS
+#define _REENTRANT
+
+#define ACE_MT_SAFE 1
+#define ACE_HAS_THREADS
+#define ACE_MAX_NUM_THREADS 64
+
+// Compiler/platform has correctly prototyped header files.
+#define ACE_HAS_CPLUSPLUS_HEADERS
```
jwillemsen, line 34: old

ACE/ace/config-integrity-common.h
```diff
@@ -0,0 +1,185 @@
...
+#define ACE_LACKS_SEEKDIR
+#define ACE_LACKS_GETHOSTENT
+
+#if !_HAS_EXCEPTIONS
+# define ACE_STD_ALLOCATOR_NOTHROW
+#endif
+
+// When config.h defines ACE_NO_GHS_LIBIO, the functions provided by the
+// INTEGRITY-178 Low-Level I/O Library (libio_178b_be.a) are not called by ACE.
+#ifdef ACE_NO_GHS_LIBIO
```
jwillemsen, line 161: Why not used ACE_NO_INTEGRITY_LIBIO, sometimes you use ghs, any reason to not use integrity?

ACE/ace/config-integrity.h
```diff
@@ -0,0 +1,74 @@
...
+#include "config-integrity-common.h"
+
+#define ACE_HAS_TSS_EMULATION
+// Even though INTEGRITY has support for thread-specific storage (see TaskKey API),
+// we are not using it since we are imitating the behavior of INTEGRITY-178 which
+// does not have thread-specific storage.
+//#define ACE_HAS_THREAD_SPECIFIC_STORAGE
+
+#if defined ACE_LACKS_POSIX
+# define ACE_LACKS_SIGEMPTYSET // new
```
jwillemsen, line 26: Remove new comment

jwillemsen (changes requested):

ACE/ace/config-integrity-common.h
sonndinh: I think this macro is used to exclude a few I/O functions from both INTEGRITY and INTEGRITY-178, thus it's named with _GHS_ in it.

ACE/ace/config-integrity-common.h
sonndinh: This was inheriting from the existing code -- the current code has mixing uses of `ghs` and `INTEGRITY` which seem to have the same meaning. I think it makes sense now to consolidate them to a single macro, may be `ACE_INTEGRITY` as another comment suggests, to follow ACE style.

TAO/tao/default_resource.cpp
mitza-oci: TAO allows, or seems to allow, a different reactor to be configured for the ORB, separate from ACE's singleton reactor or any other reactor.

ACE/ace/OS_NS_arpa_inet.cpp
```diff
@@ -99,4 +99,22 @@ ACE_OS::inet_aton (const char *host_name, struct in_addr *addr)
...
 }
 
+#if defined (ACE_INTEGRITY) && defined (ACE_LACKS_INET_NTOA)
+char* ACE_OS::inet_ntoa (const struct in_addr addr)
+{
+  ACE_OS_TRACE ("ACE_OS::inet_ntoa");
+
+  static char addrstr[INET_ADDRSTRLEN + 1] = { 0 };
+  ACE_UINT32 ipaddr = ntohl (addr.s_addr);
+  // printf("Socket address %X, IP address %X.\n",addr.s_addr,ipaddr);
```
iguessthislldo, line 109:
```suggestion
```

ACE/ace/OS_NS_sys_utsname.cpp
```diff
@@ -213,17 +213,31 @@ ACE_OS::uname (ACE_utsname *name)
...
+  const int status = ::gethostname (name->nodename, __SYS_NMLN);
+  strcpy (name->release, "minor");
+  strcpy (name->version, "5.0.0");
+  strcpy (name->machine, "a standard name");
+#    else
+  strcpy (name->sysname,"INTEGRITY");
+  const int status = ::gethostname (name->nodename, __SYS_NMLN);
+  strcpy (name->release,"minor");
+  strcpy (name->version,"11.4.6");
+  strcpy (name->machine,"a standard name");
```
iguessthislldo, line 237:
```suggestion
  strcpy (name->release, "minor");
  strcpy (name->version, "11.4.6");
  strcpy (name->machine, "a standard name");
```

iguessthislldo (changes requested):

ACE/ace/OS_NS_time.h
```diff
@@ -131,7 +131,7 @@ inline double ace_difftime(time_t t1, time_t t0)
   return difftime (t1, t0);
 # undef difftime
 # else
-  return ACE_STD_NAMESPACE::difftime (t1, t0);
```
sonndinh: Awesome. Thank you for finding this out!

ACE/ace/config-integrity-common.h
```diff
@@ -0,0 +1,181 @@
...
+
+#ifdef ppc
+# define ACE_HAS_POWERPC_TIMER
+#endif
+
+#define ACE_HAS_NONCONST_SELECT_TIMEVAL
+#include <sys/uio.h> // needed to define iovec
+#define ACE_LACKS_READLINK
+#define ACE_LACKS_GETPPID
+#define ACE_USES_STD_NAMESPACE_FOR_STDCPP_LIB 1
```
jwillemsen, line 124: Can be removed as far as I know

ACE/ace/config-integrity-common.h
```diff
@@ -0,0 +1,181 @@
...
+#define PRI_FIFO_MIN 1
+#define PRI_FIFO_MAX 127
+#define ACE_THR_PRI_FIFO_DEF 127
+#define PRI_RR_MIN 1
+#define PRI_RR_MAX 127
+#define ACE_THR_PRI_RR_DEF 127
+#define PRI_OTHER_MIN 1
+#define PRI_OTHER_MAX 127
+#define ACE_THR_PRI_OTHER_DEF 127
+#define ACE_PTHREAD_RETURN_ON_EXIT
```
jwillemsen, line 137: Is this define ever used in ACE?

ACE/ace/config-integritySCA.h
```diff
@@ -1,30 +1,47 @@
...
+//#define ACE_LACKS_GETEUID
+//#define ACE_LACKS_GETUID
+//#define ACE_LACKS_GETEGID
+//#define ACE_LACKS_GETGID
+
+#ifndef ACE_HAS_EXCEPTIONS
+  #define ACE_HAS_EXCEPTIONS
+#endif
+#define ACE_NEW_THROWS_EXCEPTIONS
+#define ACE_HAS_TEMPLATE_SPECIALIZATION
```
jwillemsen, line 39: ACE_HAS_TEMPLATE_SPECIALIZATION is not used in ACE, check the others here

ACE/ace/config-integritySCA.h
```diff
@@ -84,65 +112,54 @@
...
+#define ACE_LACKS_MALLOC_H /* netbsd's just includes stdlib.h */
+#define ACE_LACKS_MEMORY_H /* netbsd's just includes string.h */
 #define ACE_LACKS_READV
-#define ACE_LACKS_SYSCONF
-#define ACE_LACKS_GETOPT
 /* below refers to fcntl style locking */
 #define ACE_LACKS_FILELOCKS
 
 #define ACE_LACKS_REALPATH
-#define ACE_HAS_CONST_CHAR_SWAB
```
jwillemsen, line 128: Looks this is the last file which defined this, code in ACE can be cleaned up

ACE/ace/config-integritySCA.h
```diff
@@ -171,9 +183,11 @@ extern "C"
 #define ACE_LACKS_READLINK
 #define ACE_LACKS_GETPPID
 #define NSIG (SIGRTMAX+1)
+#define ACE_USES_STD_NAMESPACE_FOR_STDCPP_LIB 1
```
jwillemsen, line 186: No need to set this here

TAO/tao/Strategies/advanced_resource.cpp
```diff
@@ -473,6 +476,10 @@ TAO_Advanced_Resource_Factory::allocate_reactor_impl () const
       break;
 #endif  /* ACE_HAS_EVENT_POLL || ACE_HAS_DEV_POLL */
 
+    case TAO_REACTOR_SINGLE_INPUT:
+      ACE_NEW_RETURN (impl, ACE_Single_Input_Reactor, 0);
```
jwillemsen, line 480: return nullptr instead of 0

iguessthislldo (dismissed):

TAO/tao/Strategies/advanced_resource.cpp
sonndinh: There are other `case`s in the existing code that use 0. I think this can be done in a separate PR to address all of them.

ACE/ace/POSIX_Asynch_IO.h
```diff
@@ -32,7 +32,7 @@
 #include "ace/Unbounded_Queue.h"
 #include "ace/Map_Manager.h"
 #include "ace/Event_Handler.h"
-#if defined(INTEGRITY)
```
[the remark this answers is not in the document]
sonndinh: I'm not sure. This is just changing the `INTEGRITY` macro to `ACE_INTEGRITY`.

TAO/tao/Strategies/advanced_resource.cpp
jwillemsen: Yes, there are other cases, just use nullptr in new code

ACE/ace/POSIX_Asynch_IO.h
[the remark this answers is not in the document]
jwillemsen: As you have integrity to compile it seems a quick test to see whether this include is really necessary here

ACE/ace/OS_NS_Thread.h
```diff
@@ -224,7 +228,147 @@ ACE_END_VERSIONED_NAMESPACE_DECL
...
+                // Earlier call has already insert this entry.
+                // Just return now.
+                return ti.tss_base;
+              }
+          }
+        // If we go here, the calling thread doesn't have an entry yet.
+        // Use the first available slot.
+
+        // Should NOT create more than ACE_MAX_NUM_THREADS Tasks!
+        ACE_ASSERT (first_slot != ACE_MAX_NUM_THREADS);
```
jwillemsen, line 337: first asset, than exit, the assert will abort the process already

ACE/ace/OS_NS_Thread.inl
```diff
@@ -2496,6 +2553,13 @@ ACE_OS::thr_getprio (ACE_hthread_t ht_id, int &priority, int &policy)
   return 0;
 # elif defined (ACE_HAS_VXTHREADS)
   return ::taskPriorityGet (ht_id, &priority);
+# elif defined (ACE_INTEGRITY)
+  Value active_prio;
+  const Error err = ::GetActivePriority (ht_id, &active_prio);
+  if (err != Success)
+    return -1;
```
jwillemsen, line 2560: ACE_FAIL_RETURN?

ACE/ace/OS_NS_unistd.inl
```diff
@@ -373,7 +384,11 @@ ACE_INLINE char *
 ACE_OS::getcwd (char *buf, size_t size)
 {
   ACE_OS_TRACE ("ACE_OS::getcwd");
-#if defined (ACE_GETCWD_EQUIVALENT)
+#if defined (ACE_LACKS_GETCWD)
+  ACE_UNUSED_ARG (buf);
+  ACE_UNUSED_ARG (size);
+  ACE_NOTSUP_RETURN ((char*)-1);
```
jwillemsen, line 390: -1 looks strange, why not return nullptr

TAO/tao/default_resource.cpp
```diff
@@ -791,12 +791,14 @@ TAO_Default_Resource_Factory::get_reactor ()
...
   return reactor;
 }
 
 void
 TAO_Default_Resource_Factory::reclaim_reactor (ACE_Reactor *reactor)
 {
+  ACE_Reactor::instance (0);
```
jwillemsen, line 801: Still needs to be removed

TAO/tao/default_resource.cpp
jwillemsen: Yes, but the code here still resets it, so something to remove

ACE/ace/config-integritySCA.h
```diff
@@ -173,7 +165,7 @@ extern "C"
 #define NSIG (SIGRTMAX+1)
 #define ACE_LACKS_CUSERID
 #define ACE_NEEDS_HUGE_THREAD_STACKSIZE 0x5000
-#define fileno(file) ((file)->io_channel) //Hack to get Svc_Conf_l.cpp compiled
+//#define fileno(file) ((file)->io_channel) //Hack to get Svc_Conf_l.cpp compiled
```
jwillemsen, line 168: remove commented out line

ACE/ace/config-integritySCA.h
```diff
@@ -1,30 +1,32 @@
...
-#define ACE_LACKS_GETGID
-
-#define TAO_USE_SEQUENCE_TEMPLATES
+#define ACE_LACKS_PWD_FUNCTIONS
+#define ACE_LACKS_ALPHASORT
+
+#ifndef ACE_HAS_EXCEPTIONS
+# define ACE_HAS_EXCEPTIONS
+#endif
+#define ACE_NEW_THROWS_EXCEPTIONS
```
jwillemsen, line 29: Already set in config-all.h

ACE/ace/config-integrity178.h
```diff
@@ -0,0 +1,123 @@
...
+// However, socket functions from GHNet-178's libsocket.a must be available.
+#if defined ACE_NO_GHNET178_LIBNET
+/***** Network utility functions *****/
+# define ACE_LACKS_GETADDRINFO
+# define ACE_LACKS_GAI_STRERROR
+# define ACE_LACKS_GETNAMEINFO
+# define ACE_LACKS_GETSERVBYNAME
+# define ACE_LACKS_IF_NAMEINDEX
+# define ACE_LACKS_IF_NAMETOINDEX
+# define ACE_LACKS_GETHOSTBYADDR // new
```
jwillemsen, line 84: Remove new comment, doesn't add value

ACE/ace/config-integrity.h
```diff
@@ -0,0 +1,74 @@
...
+
+#define ACE_LACKS_SENDMSG
+#define ACE_LACKS_RWLOCK_T
+#define ACE_LACKS_SEMBUF_T
+
+#define ACE_HAS_WCHAR
+
+#define NSIG (SIGRTMAX+1)
+#define ACE_NEEDS_HUGE_THREAD_STACKSIZE 0x5000
+#define fileno(file) ((file)->io_channel) //Hack to get Svc_Conf_l.cpp compiled
```
jwillemsen, line 62: commented out in another file, still required here?

jwillemsen:
@coderabbitai full review

ACE/ace/config-integrity-common.h
```diff
@@ -0,0 +1,166 @@
...
+#define ACE_LACKS_ALPHASORT
+#define ACE_LACKS_ITOW
+
+#define ACE_HAS_SIGWAIT
+#define ACE_HAS_PTHREAD_SIGMASK_PROTO
+
+#define ACE_LACKS_SUSECONDS_T
+#define ACE_LACKS_USECONDS_T
+#define ACE_HAS_POSIX_TIME
+#define ACE_HAS_STRDUP_EMULATION
```
jwillemsen, line 57: Is ACE_HAS_STRDUP_EMULATION really required for integrity, seems the last platform still using this, would be nice when we could remove this from ACE

TAO/tao/Strategies/advanced_resource.cpp
mitza-oci: I have other changes to this coming soon, so I can change it.

ACE/ace/POSIX_Asynch_IO.h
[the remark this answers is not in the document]
mitza-oci: There are multiple Integrity configurations.  `ACE/ace/config-integritySCA.h` does have `ACE_HAS_AIO_CALLS`, but we're not using Integrity-SCA.  So I think the safest thing to do is just to leave it in place.

TAO/tao/default_resource.cpp
mitza-oci: We're working on it -- see https://github.com/DOCGroup/ACE_TAO/pull/2340#discussion_r1983330540

ACE/ace/config-integrity-common.h
mitza-oci: I don't think Integrity has `strdup` as far as I can tell.  It's also not in std C until C23 so there's bound to be some non-POSIX platform that needs it.

ACE/ace/OS_NS_Thread.inl
sonndinh: From its definition and places where it's used, looks like ACE_FAIL_RETURN is intended for Windows.

ACE/ace/config-integrity.h
sonndinh: `ace_fileno_helper()` calls this macro if one is defined. But we can remove it since this INTEGRITY version already provides `fileno` function.

mitza-oci (dismissed):

jwillemsen:
Will review on monday, found some pending comments

ACE/ace/OS_NS_Thread.cpp
```diff
@@ -21,6 +21,10 @@
 #include <memory>
 #include <algorithm>
 
+#if defined (ACE_INTEGRITY)
```
jwillemsen, line 24: `if defined (ACE_INTEGRITY) && !defined (ACE_HAS_PTHREADS)`?

ACE/ace/OS_NS_Thread.cpp
```diff
@@ -3259,6 +3301,147 @@ ACE_OS::set_scheduling_params (const ACE_Sched_Params &/*sched_params*/, ACE_id_
...
+        if (!si.occupied)
+          {
+            si.occupied = true;
+            slot = i;
+            return stack_pool_[i];
+          }
+      }
+
+    // We have created the maximum number of concurrent threads!
+    return 0;
```
jwillemsen, line 3371: nullptr

ACE/ace/OS_NS_Thread.cpp
```diff
@@ -3807,6 +3990,72 @@ ACE_OS::thr_create (ACE_THR_FUNC func,
...
+
+#   if defined (ACE_INTEGRITY178B)
+    // Use the pre-allocated stack if no stack provided
+    unsigned slot = NUM_DYNAMIC_THREADS;
+    if (!stack)
+      {
+        stack = int178_stack_manager.acquire (slot);
+        if (!stack)
+          return -1;
+        stacksize = ACE_INT178_DEFAULT_STACK_SIZE;
```
jwillemsen, line 4018: `ACE_INTEGRITY178B_DEFAULT_STACK_SIZE`?

ACE/ace/OS_NS_Thread.cpp
```diff
@@ -3951,6 +4219,72 @@ ACE_OS::thr_join (ACE_thread_t waiter_id,
 }
 #endif /* ACE_HAS_VXTHREADS */
 
+#if defined (ACE_INTEGRITY) && !defined (ACE_HAS_PTHREADS)
+int
+ACE_OS::thr_join (ACE_hthread_t thr_handle,
+                  ACE_THR_FUNC_RETURN* status)
+{
+  ACE_UNUSED_ARG (status);
```
jwillemsen, line 4227: You can comment out status as argument name, makes this unused arg not necessary

ACE/ace/OS_NS_Thread.cpp
```diff
@@ -3951,6 +4219,72 @@ ACE_OS::thr_join (ACE_thread_t waiter_id,
...
+    }
+  return retval;
+}
+
+int
+ACE_OS::thr_join (ACE_thread_t waiter_id,
+                  ACE_thread_t* thr_id,
+                  ACE_THR_FUNC_RETURN* status)
+{
+  ACE_UNUSED_ARG (thr_id);
```
jwillemsen, line 4283: commend out thr_id as arugment

ACE/ace/OS_NS_Thread.h
```diff
@@ -40,6 +40,10 @@
 # endif /* VxWorks and ! SMP */
 #endif
 
+# if defined (ACE_INTEGRITY)
+#   include "ace/Log_Msg.h"
```
jwillemsen, line 44: Do we need this include here, don't see it used

ACE/ace/OS_NS_arpa_inet.cpp
```diff
@@ -99,4 +102,24 @@ ACE_OS::inet_aton (const char *host_name, struct in_addr *addr)
 #endif  /* ACE_LACKS_INET_ATON */
 }
 
+#if defined (ACE_INTEGRITY) && defined (ACE_LACKS_INET_NTOA)
+char* ACE_OS::inet_ntoa (const struct in_addr addr)
+{
+  ACE_OS_TRACE ("ACE_OS::inet_ntoa");
+
+  static ACE_thread_mutex_t mutex;
+  static char addrstr[INET_ADDRSTRLEN + 1] = { 0 };
```
jwillemsen, line 111: The lock makes this thread safe within this function, but you are returning a static buffer so threads could overwrite this, shouldn't the buffer be a in thread local storage?

ACE/ace/OS_NS_fcntl.cpp
```diff
@@ -106,14 +106,25 @@ ACE_OS::open (const char *filename,
...
     return h;
-#elif defined (INTEGRITY)
+#elif defined (ACE_LACKS_OPEN)
+  ACE_UNUSED_ARG (filename);
+  ACE_UNUSED_ARG (mode);
+  ACE_UNUSED_ARG (perms);
+  ACE_UNUSED_ARG (sa);
+  ACE_NOTSUP_RETURN (ACE_INVALID_HANDLE);
+#elif defined (ACE_INTEGRITY)
+# if defined (ACE_INTEGRITY178B)
```
jwillemsen, line 116: Can't the elif just check for `ACE_INTEGRITY178B`?

jwillemsen: ACE_LACKS_OPEN is added to the config file for integrity, just check this part again

ACE/ace/OS_NS_signal.inl
```diff
@@ -85,7 +86,11 @@ ACE_INLINE int
 sigaddset (sigset_t *s, int signum)
 {
   ACE_OS_TRACE ("ACE_OS::sigaddset");
-#if defined (ACE_LACKS_SIGSET)
+#if defined (ACE_LACKS_SIGADDSET)
```
jwillemsen, line 89: config-mqx.h already sets this `ACE_LACKS_SIGADDSET` which isn't used in the code until now, something to check

ACE/ace/OS_NS_sys_socket.inl
```diff
@@ -135,6 +135,9 @@ ACE_OS::closesocket (ACE_HANDLE handle)
   //       process.
 
   ACE_SOCKCALL_RETURN (::closesocket ((SOCKET) handle), int, -1);
+#elif defined ACE_LACKS_CLOSE
+  ACE_UNUSED_ARG (handle);
```
jwillemsen, line 139: This looks to be a wrong check, ACE_LACKS_CLOSESOCKET?

ACE/ace/OS_NS_sys_utsname.cpp
```diff
@@ -213,17 +213,35 @@ ACE_OS::uname (ACE_utsname *name)
   ACE_OS::strcpy (name->machine, sysModel ());
 
   return ACE_OS::hostname (name->nodename, maxnamelen);
-#elif defined (INTEGRITY)
-  if(!name) {
-    errno = EFAULT;
+#elif defined (ACE_INTEGRITY)
+#  if defined (ACE_LACKS_GETHOSTNAME)
```
jwillemsen, line 217: ACE_LACKS_GETHOSTNAME is checked already in ACE_OS::hostname, so why not call this ACE_OS::hostname and when it returns -1, return -1 in this function

ACE/ace/OS_NS_unistd.inl
```diff
@@ -929,17 +954,27 @@ ACE_OS::sleep (const ACE_Time_Value &tv)
   //FUZZ: enable check_for_lack_ACE_OS
 #else
 # if defined (ACE_HAS_NONCONST_SELECT_TIMEVAL)
+#   if !defined (ACE_LACKS_SELECT)
```
jwillemsen, line 957: Maybe move ACE_LACKS_SELECT before the check for ACE_HAS_NONCONST_SELECT_TIMEVAL, would only be one time necessary at that moment instead of 2 times here

ACE/ace/config-integrity-common.h
```diff
@@ -0,0 +1,166 @@
...
+#define ACE_LACKS_SETSID
+#define ACE_LACKS_GETEGID
+#define ACE_LACKS_SETEGID
+#define ACE_LACKS_GETUID
+#define ACE_LACKS_SETUID
+#define ACE_LACKS_GETEUID
+#define ACE_LACKS_SETEUID
+#define ACE_LACKS_GETGID
+
+#define _REENTRANT
```
jwillemsen, line 19: Shouldn't this come from the OS headers?

ACE/ace/config-integrity-common.h
```diff
@@ -0,0 +1,166 @@
...
+#include <time.h>
+
+typedef void (*__sighandler_t) (int);
+
+#ifdef ppc
+# define ACE_HAS_POWERPC_TIMER
+#endif
+
+#define ACE_HAS_NONCONST_SELECT_TIMEVAL
+#include <sys/uio.h> // needed to define iovec
```
jwillemsen, line 108: Do you need that include here?

ACE/ace/config-integritySCA.h
```diff
@@ -84,83 +96,58 @@
...
-
-/***** STUFF INTEGRITY 4.0.8 APPEARS TO SUPPORT ****/
-/* note, possibly untested with ace */
-
-/***** TAO STUFF ****/
-#define TAO_USE_DOTTED_DECIMAL_ADDRESSES 1
 
 #include <INTEGRITY.h>
 
 #include <time.h>
```
jwillemsen, line 147: Do you need this include here?

TAO/tao/Strategies/advanced_resource.cpp
```diff
@@ -186,95 +176,80 @@ TAO_Advanced_Resource_Factory::init (int argc, ACE_TCHAR** argv)
...
-                ACE_TEXT("TkReactor not supported by Advanced_Resources_Factory. Please use TAO_TkResource_Loader instead."),
-                         current_arg);
+          else if (ACE_OS::strcasecmp (current_arg, ACE_TEXT ("fl")) == 0)
+            this->report_option_value_error (ACE_TEXT ("FlReactor not supported by Advanced_Resources_Factory. Please use TAO_FlResource_Loader instead."),
+                                             current_arg);
+          else if (ACE_OS::strcasecmp (current_arg, ACE_TEXT ("tk")) == 0)
+            this->report_option_value_error (ACE_TEXT ("TkReactor not supported by Advanced_Resources_Factory. Please use TAO_TkResource_Loader instead."),
+                                             current_arg);
+          else if (ACE_OS::strcasecmp (current_arg, ACE_TEXT ("single_input")) == 0)
+            this->reactor_type_ = TAO_REACTOR_SINGLE_INPUT;
```
jwillemsen, line 186: Mention in docs

jwillemsen:
Add news items

ACE/ace/OS_NS_fcntl.cpp
sonndinh: Yes, `ACE_LACKS_OPEN` is added because there is INTEGRITY-178 config where `open` is not available.

ACE/ace/OS_NS_signal.inl
sonndinh: Looks like it is added to config-mqx.h in PR #917 but isn't used anywhere else. I don't have an MQX system to test, so I don't know whether we should keep it in config-mqx.h.

ACE/ace/OS_NS_arpa_inet.cpp
sonndinh: Or `inet_ntoa` can be non-thread-safe (https://pubs.opengroup.org/onlinepubs/9699919799.orig/) and the caller will need to be aware of that?

ACE/ace/OS_NS_sys_socket.inl
sonndinh: I think it's correct -- `close()` is called if `ACE_LACKS_CLOSE` is not defined.

TAO/tao/Strategies/advanced_resource.cpp
sonndinh: I updated TAO\docs\Options.html to add this new `-ORBReactorType ` type. Is there other place that needs to be updated?

ACE/ace/OS_NS_fcntl.cpp
jwillemsen: Ok, I see that on line 109, but on line 115/116 you again check for integrity, where the code on line 117/118 is the same as on 133/134

TAO/tao/Strategies/advanced_resource.cpp
jwillemsen: No, that is the place

ACE/ace/OS_NS_sys_socket.inl
jwillemsen: Why not call ACE_OS::close here?

ACE/ace/OS_NS_signal.inl
jwillemsen: Safest seems to remove ACE_LACKS_SIGADDSET from config-mqx.h, at the moment it is required a compile error is given, if you keep it there and it shouldn't be there it could give only a strange runtime issue

ACE/ace/OS_NS_sys_socket.inl
sonndinh: That sounds good.

ACE/ace/OS_NS_Thread.inl
```diff
@@ -2038,6 +2065,23 @@ ACE_OS::sema_wait (ACE_sema_t *s)
...
+  if (::GetSemaphoreValue (*s, &curr_val) != Success)
+    return -1;
+
+  if (curr_val == LONG_MIN)
+    // If we have this many Tasks waiting for a semaphore, something seems incorrect.
+    return -1;
+#     endif
+
+  return ::WaitForSemaphore (*s) == Success ? 0 : -1;
 #   endif /* ACE_HAS_PTHREADS */
```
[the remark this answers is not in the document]
sonndinh: While the check for the semaphore's value is not perfect, the potential race is harmless -- it's just to return early and prevent a call to `WaitForSemphore` in case the semaphore's value is already minimum. However, a call to `WaitForSemaphore` in that case might still be handled by this function itself.

The case when the semaphore's value can underflow is very rare as it requires so many threads alive simultaneously and all uses this semaphore. So this workaround check can be removed.

ACE/ace/Flag_Manip.cpp
```diff
@@ -21,43 +21,90 @@ ACE_BEGIN_VERSIONED_NAMESPACE_DECL
...
-
-  if (ACE_OS::fcntl (handle, F_SETFL, val) == -1)
-    return -1;
-  else
-    return 0;
-#endif /* ACE_LACKS_FCNTL */
+# else
+  return set_flags_helper (handle, flags);
+# endif /* ACE_LACKS_FCNTL */
+#endif /* ACE_USES_GHS_ISIMPPC */
```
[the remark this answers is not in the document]
sonndinh: The code is already self-documented and I don't see any changes needed.

ACE/NEWS
```diff
@@ -18,6 +18,8 @@ USER VISIBLE CHANGES BETWEEN ACE-8.0.2 and ACE-8.0.3
 
 . Dropped support for Embarcadero C++ Builder bcc32c and bcc64 compilers
 
+. Updated support for Green Hills INTEGRITY and INTEGRITY-178 tuMP RTOS
```
jwillemsen, line 21: Maybe mention here which versions?

ACE/ace/OS_NS_Thread.cpp
```diff
@@ -3807,6 +3990,72 @@ ACE_OS::thr_create (ACE_THR_FUNC func,
...
+
+    *thr_id = *thr_handle;
+
+    // Set the argument for the new Task's entrypoint, then run it.
+    mutex_lock (&integrity_task_args_lock);
+    integrity_task_args[*thr_handle] = thread_args;
+
+    Semaphore sema;
+    if (::CreateSemaphore (0, &sema) != Success)
+      return -1;
```
jwillemsen, line 4053: You are not releasing integrity_task_args_lock when this fails, maybe use a guard for all calls to mutex_lock/mutex_unlock?

ACE/ace/OS_NS_Thread.cpp
```diff
@@ -3951,6 +4219,70 @@ ACE_OS::thr_join (ACE_thread_t waiter_id,
...
+      retval = EDEADLK;
+    }
+  else
+    {
+      Semaphore join_sema = 0;
+      ACE_OS::mutex_lock (&integrity_task_args_lock);
+      auto it = integrity_join_semas.find (thr_handle);
+      if (it == integrity_join_semas.end ())
+        {
+          errno = EINVAL;
```
jwillemsen, line 4247: Not releasing the lock in case of a failure, use guard?

ACE/ace/OS_NS_sys_socket.inl
```diff
@@ -892,6 +891,7 @@ ACE_OS::sendv (ACE_HANDLE handle,
     }
   return ACE_OS::writev (handle, local_iov, n);
 
+
```
jwillemsen, line 894: Remove this additional empty line

ACE/ace/OS_NS_Thread.inl
```diff
@@ -2349,6 +2384,10 @@ ACE_OS::sigwait (sigset_t *sset, int *sig)
     // means forever.
     *sig = ::sigtimedwait (sset, 0, 0);
     return *sig;
+# elif defined (ACE_INTEGRITY)
```
jwillemsen, line 2387: Add a ACE_LACKS_SIGWAIT?

ACE/ace/OS_NS_Thread.cpp
sonndinh: Good catch. Thanks!

ACE/ace/OS_NS_Thread.cpp
```diff
@@ -3410,10 +3410,12 @@ extern "C" void integrity_task_adapter ()
 {
   const ACE_hthread_t curr_task = CurrentTask ();
 
-  // The thread argument should be ready when this function starts
-  ACE_OS::mutex_lock (&integrity_task_args_lock);
-  ACE_Base_Thread_Adapter *const thr_arg = integrity_task_args[curr_task];
-  ACE_OS::mutex_unlock (&integrity_task_args_lock);
+  ACE_Base_Thread_Adapter *thr_arg = 0;
```
jwillemsen, line 3413: nullptr

jwillemsen:
I will review another time later today

ACE/ace/Flag_Manip.cpp
```diff
@@ -21,43 +21,90 @@ ACE_BEGIN_VERSIONED_NAMESPACE_DECL
...
 int
 ACE::set_flags (ACE_HANDLE handle, int flags)
 {
   ACE_TRACE ("ACE::set_flags");
-#if defined (ACE_LACKS_FCNTL)
+#if defined (ACE_USES_GHS_ISIMPPC)
+  switch (flags)
+    {
+    case ACE_NONBLOCK:
+      ACE_UNUSED_ARG (handle);
```
jwillemsen, line 70: Shouldn't this be a not sup return -1, the higher layer has explicitly requested nonblock which is not supported, currently this requested flag is just silently ignored

ACE/ace/OS_NS_Thread.inl
```diff
@@ -2328,7 +2361,7 @@ ACE_OS::sigwait (sigset_t *sset, int *sig)
   if (sig == 0)
     sig = &local_sig;
 #if defined (ACE_HAS_THREADS)
-# if (defined (__FreeBSD__) && (__FreeBSD__ < 3))
+# if (defined(__FreeBSD__) && (__FreeBSD__ < 3)) || defined (ACE_HAS_WTHREADS) || defined (ACE_LACKS_SIGWAIT)
```
jwillemsen, line 2364: __FreeBSD__ 3 is from 1996, can you maybe remove the __FreeBSD__ check here?

ACE/tests/Dynamic_Priority_Test.cpp
```diff
@@ -47,9 +47,9 @@
 #include "ace/High_Res_Timer.h"
 #include "ace/Sched_Params.h"
 
-#if defined (VXWORKS)
+#if defined (VXWORKS) || defined (INTEGRITY)
 # include "ace/OS_NS_string.h"
```
jwillemsen, line 51: Why add INTEGRITY here?

ACE/ace/Flag_Manip.cpp
sonndinh: This is specific to the INTEGRITY simulator which doesn't support non-blocking mode. Returning 0 allows the applications/tests to continue on the simulator and running with blocking mode is acceptable in those programs. That is, it isn't an error indication in this case.

ACE/tests/Dynamic_Priority_Test.cpp
sonndinh: It's needed for the `ACE_OS::strlen()` function.

sonndinh:
> No other comments as far as I can tell, I assume you have done some local builds on linux/windows with running all tests to make sure it doesn't break anything (github CI only compiles the core and ACE unit tests but doesn't run any tests)

I have a build on Windows. Is there a scoreboard that I can compare the status of the tests against?

jwillemsen:
At this moment there is no public build

sonndinh:
> At this moment there is no public build

I ran the ACE tests locally on a Windows build. There are 4 failed tests:
- Bug_2610_Regression_Test and TSS_Leak_Test have similar error messages as one of the scoreboard build (https://storage.googleapis.com/oci-scoreboard/ACE_TAO_Ubuntu_22_04_Clang19_x64_Debug/2025_03_21_15_01_Brief.html#section_6).
- The other two failed tests are Bug_2434_Regression_Test and Config_Test. However, from the output messages, they don't seem to come from the changes in this PR.

jwillemsen:
Nothing to add at this moment, but I haven't reviewed all last changes in detail due to lack of time

ACE/ace/OS_NS_arpa_inet.cpp
mitza-oci: There's a lot of added complexity in going to a thread-local; and the lack of safety in going with non-thread-safe (even though POSIX allows it) doesn't look good either.

How about just ACE_LACKS_INET_NTOA for this platform?  It doesn't seem to be used in core ACE+TAO libs and inet_ntop is a better alternative.

ACE/ace/OS_NS_sys_socket.inl
```diff
@@ -136,9 +137,7 @@ ACE_OS::closesocket (ACE_HANDLE handle)
 
   ACE_SOCKCALL_RETURN (::closesocket ((SOCKET) handle), int, -1);
 #else
-  //FUZZ: disable check_for_lack_ACE_OS
-  return ::close (handle);
-  //FUZZ: enable check_for_lack_ACE_OS
+  return ACE_OS::close (handle);
```
mitza-oci, line 140: Do we know why the old code went directly to `::close` instead of `ACE_OS::close`?

ACE/ace/config-integrity-common.h
mitza-oci: I'm not entirely sure how Integrity uses it, but older versions of glibc left this up to the user to define.

ACE/ace/OS_NS_arpa_inet.cpp
sonndinh: If I remember correctly, there were places in TAO when the ORB is initialized that this function is called. But that was the older TAO version. I'm not sure if current TAO still uses it.

ACE/ace/OS_NS_sys_socket.inl
sonndinh: I'm not sure. Maybe by mistake and it hasn't caught a platform with `ACE_LACKS_CLOSE`?

ACE/ace/OS_NS_arpa_inet.cpp
sonndinh: I tested locally with this emulation function removed and it doesn't look like TAO is using it. So I removed it and `inet_ntoa` now just returns unsupported with ACE_LACKS_INET_NTOA on this platform. @jwillemsen

ACE/ace/OS_NS_sys_utsname.cpp
```diff
@@ -214,13 +214,17 @@ ACE_OS::uname (ACE_utsname *name)
 
   return ACE_OS::hostname (name->nodename, maxnamelen);
 #elif defined (ACE_INTEGRITY)
+# if defined (ACE_LACKS_GETHOSTNAME)
+  ACE_UNUSED_ARG (name);
```
jwillemsen, line 218: Why make this exception here, doesn't ACE_OS::hostname handle this?

sonndinh: Calling ACE_OS::hostname causes an infinite calling loop between ACE_OS::hostname and ACE_OS::uname and gethostname is never called.

mitza-oci (approved):

PR #2373, opened 2025-04-30
Added missing Debug Level testing for "Send Reply Current is 0"

jeteve:
Does what it says in the title.

I'm also wondering if it is intentional that this `TAOLIB_DEBUG` statement is not under a debug condition?

NB: Also ignore .devcontainer

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.devcontainer
 *.vcxproj*
 *.tlog
 *.log
--- a/TAO/tao/RTScheduling/Request_Interceptor.cpp
+++ b/TAO/tao/RTScheduling/Request_Interceptor.cpp
@@ -406,8 +406,14 @@ Server_Interceptor::send_reply (PortableInterceptor::ServerRequestInfo_ptr ri)
       // Reset the previous current pointer.
       tss->rtscheduler_previous_current_impl_ = 0;
     }
-  else TAOLIB_DEBUG ((LM_DEBUG,
-                   "Send Reply Current is 0\n"));
+  else
+    {
+      if (TAO_debug_level > 0)
+        {
+          TAOLIB_DEBUG ((LM_DEBUG,
+                        "Send Reply Current is 0\n"));
+        }
+    }
 }
 
 void


mitza-oci:
> wondering if it is intentional that this TAOLIB_DEBUG statement is not under a debug condition?

It may be, see similar statement earlier in this function.  If the developer thought this was a very serious warning it could be logged at all levels (users still have control over the logger objects and if/where they output).  In this case I'm not sure what the correct behavior is.

jeteve:
Thanks @mitza-oci for looking into it.

Maybe @jwillemsen will have an opinion?

If the warning is serious, I would have expected the macro to be used to be `TAOLIB_ERROR` instead.

mitza-oci:
If `Server_Interceptor::receive_request` completes successfully (reaches the end without throwing or returning early), the value should not be null when `Server_Interceptor::send_reply` goes looking for it.

If you have an application that exhibits this behavior it would probably be good to find out why receive_request isn't working, or if there's some other bug.  If you can add a test to TAO that would be ideal.

jeteve:
Ho dear ho dear. I'll look deeper. I saw this patch in our application code, and I wasn't 100% sure, but it smells like the wrong fix for our real problem.

jeteve:
Something is weird though.

If you stay in the context of `Request_Interceptor.cpp`, it looks like `tss->rtscheduler_current_impl_` can only be asigned if.. it's already assigned:


```cpp
  current = static_cast<TAO_RTScheduler_Current_i *> (tss->rtscheduler_current_impl_);
  // Then under condition:
  if (current != 0)
   {
    .. some logic ..
    // build new_current and..
    tss->rtscheduler_current_impl_ = new_current;
  }
```
So unless `rtscheduler_current_impl_` is bootstrapped somewhere else, it simply cannot be 'not null' just by using the code in this class.

mitza-oci:
See the last statement of `void Server_Interceptor::receive_request (PortableInterceptor::ServerRequestInfo_ptr ri)`:
```C++
  tss->rtscheduler_previous_current_impl_ = this->current_->implementation (new_current);
```
The call to `TAO_RTScheduler_Current_i* TAO_RTScheduler_Current::implementation (TAO_RTScheduler_Current_i* new_current)` stores the value in `tss->rtscheduler_current_impl_`

jeteve:
Thanks @mitza-oci ! Do you know if there's a reason why this wouldn't happen in some application code that integrate the ORB?

PR #2390, opened 2025-06-04
Updates for Visual C++ CI and C++20 compatibility

mitza-oci:


.github/workflows/windows.yml
```diff
@@ -17,59 +17,43 @@ jobs:
...
+            mpctype: vs2022
             BuildPlatform: Win32
             BuildConfiguration: Release
             vcpkgarch: x86-windows
             vcpkglibdir: lib
             vcpkgpackages: '"openssl", "xerces-c", "zlib"'
-          - name: VS2019Debug64Cxx17
-            vmimage: windows-2019
-            mpctype: vs2019
+          - name: VS2022Debug64Cxx17
```
jwillemsen, line 45: Cop17 is the default so the additional mpc args could be removed

mitza-oci: Projects in VS2022 default to C++14.

jwillemsen: Mpc does set cpp17 as default, that is the minimum requirement for ace 800 and newer

mitza-oci: I'd rather have it in the overall project generation steps.  There could be projects that don't inherit `acedefaults.mpb`.

mitza-oci: OK, thanks for linking to that.  It's a different case than what I was thinking of.
I'll just change it here.

I did a test to see which setting takes effect when both `-value_template` and `specific` are used.  Unfortunately it's `-value_template` (there's probably a good reason for that).

PR #2396, opened 2025-06-09
[ACE 6 - TAO 2] Fixed warnings

mitza-oci:
Merge pull request #2395 from mitza-oci/warnings

PR #2399, opened 2025-06-11
[ace6tao2] Add missing include of Truncate.h

jwillemsen:
* ACE/Kokyu/Default_Dispatcher_Impl.cpp:
    * ACE/ace/Message_Queue_NT.cpp:


jwillemsen:
Caused by the removal of the include of Truncate.h as part of https://github.com/DOCGroup/ACE_TAO/pull/2396, some other files didn't include Truncate.h explicitly

PR #2400, opened 2025-06-12
Removed usage of variable length arrays, fixed memory leaks

jwillemsen:
* ACE/ace/Svc_Conf_Lexer.cpp:

## Summary by CodeRabbit

- **Refactor**
  - Improved memory management by replacing macro-based temporary string allocation with modern, safer heap allocation using smart pointers. This enhances code reliability and maintainability without affecting user-facing features.


jwillemsen:
clang20 does always give a warning now when a variable length array is used

mitza-oci:
`std::string` can do the same thing (small string optimization) -- would it be better to use it directly?

PR #2424, opened 2025-07-10
Fixed compile errors with C++23

jwillemsen:


jwillemsen:
This is with another compiler which uses C++23 as default, but can't share the logs and details publicly at this moment.

PR #2446, opened 2025-08-07
[ACE 6] Conditional compilation for Windows thread names

mitza-oci:
Older versions of Windows don't support SetThreadDescription. When building an ACE library that needs to be compatible with these, set ACE_LACKS_SETTHREADDESCRIPTION in config.h.

This can't be determined from the Windows headers which include SetThreadDescription unconditionally.

### DIFF
--- a/ACE/ace/OS_NS_Thread.cpp
+++ b/ACE/ace/OS_NS_Thread.cpp
@@ -4113,10 +4113,12 @@ ACE_OS::thr_create (ACE_THR_FUNC func,
                                                 flags,
                                                 thr_id);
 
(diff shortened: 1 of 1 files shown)


ACE/ace/OS_NS_Thread.cpp
```diff
@@ -4113,10 +4113,12 @@ ACE_OS::thr_create (ACE_THR_FUNC func,
                                                 flags,
                                                 thr_id);
 
+#   ifndef ACE_LACKS_SETTHREADDESCRIPTION
```
jwillemsen, line 4116: Something to document in ace/README?

PR #2447, opened 2025-08-07
Runtime linking for Windows thread names

mitza-oci:
Older versions of Windows don't support SetThreadDescription.
When building an ACE library that needs to be compatible with these,
set ACE_LACKS_SETTHREADDESCRIPTION in config.h.

This can't be determined from the Windows headers which
include SetThreadDescription unconditionally.

## Summary by CodeRabbit

- Bug Fixes
  - Improved Windows compatibility: thread naming is now applied only when the OS provides the capability, avoiding failures on older systems; public APIs unchanged.

- Tests
  - Extended thread attribute test to supply a thread name and additional activation attributes, increasing coverage without altering public interfaces.

### DIFF
--- a/ACE/ace/OS_NS_Thread.cpp
+++ b/ACE/ace/OS_NS_Thread.cpp
@@ -3884,7 +3884,20 @@ ACE_OS::thr_create (ACE_THR_FUNC func,
 
       if (thr_name && *thr_name && *thr_handle)
         {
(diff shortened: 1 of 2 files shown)


jwillemsen:
README should contain a note that this flag can't be set automatically and must be user enabled, else it is candidate for cleanup at some point as no config file defines this. 

Coderabbit has a proposal to only call this method when it is there, maybe try that, that codesnippet can use more documentation for the long term.

iguessthislldo (approved):

jwillemsen (approved):

PR #2449, opened 2025-08-13
[ACE 6] Thread names updates for Windows and Linux

mitza-oci:


jwillemsen:
See https://github.com/DOCGroup/ACE_TAO/pull/2447#issuecomment-3187254391, new runtime failure on linux after the change in the unit test

PR #2468, opened 2025-09-19
Avoid using a raw pointer in the proactor queue to avoid use after free.

thildenbrand:
The timer queues in the ACE_Proactor work on raw pointer of ACE_Handler. In our multicore environment this can lead to situations, where the timer element is already being worked on, while the ACE_Handler is destructed on a different core.

These situations need to be handled correctly by every single user of the timer system, which has proven itself to be error prone.

For example one core could execute ACE_Proactor_Handle_Timeout_Upcall::timeout and access the ACE_Handler::Proxy_Ptr via the ACE_Handler*, while the ACE_Handler was destructed concurrently on another core. This resulted in a double free of the memory managed via the ACE_Handler::Proxy_Ptr.

ACE_Handler::Proxy_Ptr is already a smart pointer and the situation could be improved by storing and ACE_Handler::Proxy_Ptr instead of an ACE_Handler* in the queues of the ACE_Proactor.

Applications could then savely delete the ACE_Handler, while the memory of the ACE_Handler::Proxy would still be valid. Centra application code could could then try to retrieve the ACE_Handler* in a thread-safe way and the code using the timer system could free ACE_Handler at any time, without having to care for potentially outstanding callbacks.

### DIFF
--- a/ACE/ace/Asynch_IO.h
+++ b/ACE/ace/Asynch_IO.h
@@ -1620,6 +1620,9 @@ public:
    */
   virtual void handle_wakeup ();
 
+  /// Call before destruction to ensure no more callbacks can happen.
+  void deregister_callback ();
+
   /// Get the proactor associated with this handler.
   ACE_Proactor *proactor ();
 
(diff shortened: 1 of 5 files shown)


ACE/ace/Asynch_IO.h
```diff
@@ -1649,10 +1652,18 @@ class ACE_Export ACE_Handler
...
   public:
     Proxy (ACE_Handler *handler) : handler_ (handler) {}
-    void reset () { this->handler_ = 0; }
+    void reset ()
+    {
+      acquire ();
+      this->handler_ = 0;
+      release ();
+    }
     ACE_Handler *handler () { return this->handler_; }
```
jwillemsen, line 1661: Is reading itself thread safe?

thildenbrand: It would probably be the bigger change for users, since it requires them to use shared_pointers to manage the life time of their ACE_Handlers, right?

PR #2545, opened 2026-05-27
Update Timer_List_T.cpp

okan-nako:
Since increasing id_counter_ can lead to an overflow, we propose adding an overflow check.

## Bug Fixes

* Improved timer identifier assignment to prevent counter overflow issues during long-running scheduling, resulting in more reliable timer behavior over time.

### DIFF
--- a/ACE/ace/Timer_List_T.cpp
+++ b/ACE/ace/Timer_List_T.cpp
@@ -210,8 +210,15 @@ ACE_Timer_List_T<TYPE, FUNCTOR, ACE_LOCK, TIME_POLICY>::schedule_i (const TYPE &
 
   if (n != 0)
   {
-    long id = this->id_counter_++;
-
+    long id = this->id_counter_;
+    if (this->id_counter_ < LONG_MAX)
+    {
+      this->id_counter_++;
+    }
+    else
+    {
+      this->id_counter_ = 0;
+    }
     if (id != -1) {
       n->set (type, act, future_time, interval, 0, 0, id);
       this->schedule_i (n, future_time);


ACE/ace/Timer_List_T.cpp
```diff
@@ -210,8 +210,15 @@ ACE_Timer_List_T<TYPE, FUNCTOR, ACE_LOCK, TIME_POLICY>::schedule_i (const TYPE &
 
   if (n != 0)
   {
-    long id = this->id_counter_++;
-
+    long id = this->id_counter_;
+    if (this->id_counter_ < LONG_MAX)
```
jwillemsen, line 214: std::numeric_limits<long>::max(), prefix increment, and could id ever be -1?

okan-nako: Yes, current type of "id_counter_" is "long" (Timer_List_T.h:# 211 --> long id_counter_;), we encountered even the overflow (-1).

jwillemsen: But with the new proposed code, I don't see how it can become -1, or do I miss something?

okan-nako: in case "long id = this->id_counter_++;", "id" becomes "-1" when overflow happens.

jwillemsen: That was in the old code, now you check < LONGMAX

jwillemsen (changes requested):

ACE/ace/Timer_List_T.cpp
okan-nako: in the old code, when overflow happens, id was becoming -1. Now we check < LONGMAX and -1 doesnt happen any more.

PR #2555, opened 2026-07-22
Compile fixes for SSL on Windows 64bit

jwillemsen:
## Summary by CodeRabbit

* **Refactor**
  * Clarified platform-aware `select()` width/range handling across SSL accept/connect, proxy connector, SSL stream, and related example event loops without changing behavior.
  * Updated Win32 OpenSSL file-descriptor assignment to use safer integer casting.
  * Streamlined a small SSL context variable initialization in DH parameter loading.
* **Style**
  * Minor cleanup to make readiness/event-loop `select()` calculations more explicit.


jwillemsen:
Fixed https://github.com/DOCGroup/bugzilla/issues/4179

PR #2586, opened 2026-09-21
Value initialize IDL struct members

jwillemsen:
Value initialize IDL struct members to prevent unitialized reads which could potentially trigger MARSHAL exceptions when sending a struct with an unitialized boolean with compiler optimizations enabled

## Summary by CodeRabbit

- **Bug Fixes**
  - Generated IDL structures, exceptions, valuetypes, and valueboxes now initialize fundamental members to defined default values.
  - Improved construction, assignment, copying, and default initialization for fixed-size structures stored in unions.
  - Union-generated code now provides more reliable member storage and initialization.

- **Tests**
  - Added coverage for default values in generated types and boxed values.
  - Added checks confirming union values remain correct during copying and assignment.


jwillemsen:
Impact on generated code https://github.com/DOCGroup/tao_idlgen_test/pull/1

jwillemsen:
Regenerated test code https://github.com/DOCGroup/tao_idlgen_test/pull/3

